PR #67, opened 2018-11-14
Make logging path adjustable

Laurens-makel:


nielsm5:
Fixed in 00bc31056b1747c91e6af09dc22a19235cf82bc7

PR #68, opened 2018-11-16
Add cookie notification

Laurens-makel:


nielsm5:
Nice first draft, expanded it a little and redid the service. See: #349f678

PR #69, opened 2018-11-21
Adjust highlight in menu

Laurens-makel:


nielsm5:
Looks nice, but you forgot the configuration states!

PR #71, opened 2018-11-26
Add tooltips and colors to adapter summary view

Laurens-makel:


nielsm5:
Nice idea, like how you executed things but please use existing classes before creating new ones..
See #3e152a9

PR #87, opened 2019-01-17
Fix ExecuteJdbcQuery issues

dansebmey:
Queries executed through the console's direct query tool will now send the message to the ManageDatabase adapter, allowing the pipeline to be rerun in the Ladybug tool.

Additionally, it is now possible to select a different jmsRealm for each of the queries within a ManageDatabase call. By default, all queries have the default "jdbc" jmsRealm, or the one specified in the direct query form. To change the jmsRealm of a specific query, the `jmsRealm` attribute followed by the realm's name can be added to that query's `<sql>` element.


gvanbrakel:
XmlQuerySender.java regel 290: setJmsRealm hier is niet thread safe  
ExecuteJdbcQuery.java 115: bevat nog System.out.println();

gvanbrakel:
JdbcFacade.java
- getDbmsSupport():
  geen e.printStackTrace();, maar throw new RuntimeException, zoals in getDbmsSupport(String dsName)

XmlQuerySender.java
- regel 291: Strings vergelijken met equals()
- regel 292: hier gaat oorspronkelijke connection verloren. Die wordt nooit meer geclosed.
  In plaats hiervan moet je de nieuwe connectie assignen aan een andere variabele, en die zelf in een finally clause sluiten.
- configure()
  je kan niet aannemen dat jmsRealm="jdbc" 
  je kan niet aannemen dat datasoureName="jdbc/ibis4example"
  of wel?

ExecuteJdbcQueryExecute.java
- regel 171:
  List datasourceNames = JmsRealmFactory.getInstance().getRegisteredRealmNamesAsList();
  datasourceNames wordt gevuld met RealmNames
  
ExecuteJdbcQuery.java
- regel 115: System.out.println verwijderen

example/src/main/resources/Configuration.xml
- verwijder 'secondDatabase', dit is geen test file.

example/src/main/resources/springCustom.xml
- secondDataSource: die zou er niet moeten zijn
- proxiedDataSources: dit lijkt me ook test code

dansebmey:
Hi Gerrit, bedankt voor de feedback! Ik ga er mee aan de slag.

gvanbrakel:
JdbcFacade is set up as a base class connecting to a single database.
It provides methods for descender classes to build upon, like:
- getConnection()
- getDbmsSupport()
- getDatabaseType()
- getPhysicalDestinationName()
To make these methods performant, some basic objects are cached as global fields in JdbcFacade.

The proposed change aims at making a GUI functionality work with a end-user selectable database.
It tries to do so by turning JdbcFacade into a 'multi database' class. I question whether that is the right approach.
As it is right now, it sacrifies performance. For instance in the current situation, a getConnection(), 
which is a method that is executed very often, comes down to:

  this.datasource.getConnection()
  
whereas in the new situation, the datasource is looked up from JNDI each time.

Apart from that, a lot of code in descender classes relies on getDbmsSupport().
This code will break when the multi database feature is used with multiple types of databases.

I don't think it is a good idea to transform JdbcFacade into a multi-database class, just for this US

gvanbrakel:
I would like to suggest to solve the original UC in another way, by having XmlQuerySender hosting a map of DirectQuerySenders. Then the basic interface of JdbcFacade does not need to change

dansebmey:
Hi Gerrit, those changes were added in commit 25819ab ("Add sync block for previously non-thread-safe call"). I can see the changes in the overview of file changes; I think you might have overlooked it.

gvanbrakel:
XmlQuerySender.java
- je hoeft niet helemaal de logica van de JdbcQuerySenders over te bouwen.
  Je zou gewoon dqs.sendMessage moeten kunnen aanroepen.
  Misschien moet voor elke datasource twee DirectQuerySenders hebben: één voor queryType="select" en één voor queryType="other"
  Dan kan XmlQuerySender een stuk eenvoudiger worden.

PR #88, opened 2019-01-17
Add sendGridSender as mailhandler 

alisihab:
using library sendgrid-java:4.3.0

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -526,6 +526,11 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
+			<groupId>com.sendgrid</groupId>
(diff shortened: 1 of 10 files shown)


gvanbrakel:
Ali,
In MailSenderBase you implemented the methods readParameters(), parseXML(), getFrom() and getReplyTo()  as void methods. They return their data in global variables. That makes this class thread-unsafe.
Maybe there could be also some more fields/methods moved to MailSenderBase, for instance:
	protected String charSet = null;
	protected String threadTopic = null;
	private int timeout = 20000;
	private String defaultSubject;
	private String defaultFrom;
	public boolean isSynchronous()

gvanbrakel:
Replaced by PR#91

PR #98, opened 2019-02-22
Javadoc changes #Part1

0851997:
* Removed most of @throws annotations
* Removed some @param
* Rare changes to some of the method documentations


gvanbrakel:
Please change the construct: "This method xxxs"
into: "Xxx" 
This makes the javadoc more compact and readable.
For instance, change: "This method lists the parameters" into: "List the parameters"

0851997:
The last changes for Javadoc.

PR #105, opened 2019-03-11
Add Ftp file system

alisihab:
Introduce ftp filesystem and sender


gvanbrakel:
As discussed, we decided to add the following improvments:
  change 
    public void IFileSystem.augmentFileInfo(XmlBuilder fileInfo, FTPFile f) 
  into
    public Map<String,Object> IFileSystem.getAdditionalFileProperties(FTPFile f) 
This will also have some impact on other FileSystems and FileSystemSenders

FtpFileSystem can extend FtpSession.
FtpSession will be refactored / cleaned up

PR #106, opened 2019-03-12
Add Cmis sender delete action

alisihab:
Add delete action and fix configure method related with get action


cmis/src/main/java/nl/nn/adapterframework/extensions/cmis/CmisSender.java
```diff
@@ -678,6 +697,32 @@ private void processProperties(Element propertiesElement, Map<String, Object> pr
...
+				log.info(getLogPrefix() + "document with id [" + message + "] not found", e);
+				return getResultOnNotFound();
+			} else {
+				throw new SenderException(e);
+			}
+		}
+		if (object.hasAllowableAction(Action.CAN_DELETE_OBJECT)) { //// You can delete 
+			Document suppDoc = (Document) object;
+			suppDoc.delete(true);
+			return "Document Deleted";
```
nielsm5: Could you return a XML output instead?

PR #133, opened 2019-05-17
Avoid memory bomb exploding when test-xmls are validated

gvanbrakel:


nielsm5:
It might be good to add that this is because Eclipse validates all resources, and crashes while validating scenario09/in.xml

PR #151, opened 2019-05-29
xml-external-entities Added a property which turns off XML external e…

anydoby:
…ntity resolution.

Our pentesters discovered that you can use XXE in `TestPipeLineExecute` and `TestServiceExecute`. 

For a better implementation I guess it would be nicer if XmlValidator and other classes that are doing XML processing were spring beans (for easier configuration).


nielsm5:
This has been fixed in PR #1459.

PR #159, opened 2019-06-12
Improve testing for Xpaths with escaped characters

lenardvandermaas:
Create, move and change Larva, Pipe and Unit tests for Xpaths with escaped characters


gvanbrakel:
Decided to close, after discussion with Niels and Jaco

PR #160, opened 2019-06-17
IbisDocApp (v1)

celmoussaoui:
The first version of the new IbisDoc application. 

- Contains Attributes of the class
- Contains Attributes of the inherited class(es)
- Contains an 'Examples' link
- Contains multiple search bars


example/pom.xml
```diff
@@ -89,6 +89,9 @@
               <value>LOC</value>
             </systemProperty>
           </systemProperties>
+          <webAppConfig>
```
nielsm5: Can you remove this, or is it absolutely necessarily?

core/src/main/resources/xml/xsl/web/adapterSite.xsl
```diff
@@ -467,6 +467,7 @@
 			<xsl:when test="@type='showsummary'"		>images/ibisstore.gif</xsl:when>
 			<xsl:when test="@type='theme'"				>images/new-gui.png</xsl:when>
 
+
```
nielsm5: Please try to refrain from committing empty lines

core/pom.xml
```diff
@@ -573,6 +573,11 @@
 			<version>2.3.5</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>com.google.code.gson</groupId>
+			<artifactId>gson</artifactId>
+			<version>2.8.5</version>
+		</dependency>
```
nielsm5: Why do we need this dependency? It's never used?

webapp/src/main/webapp/jsp/menuBar.jsp
```diff
@@ -121,4 +121,4 @@
 				</imagelink>
 			<% } %>
 	</imagelinkMenu>
-</menuBar>
\ No newline at end of file
+</menuBar>
```
nielsm5: Different line endings?

core/pom.xml
celmoussaoui: I used it for conversion to JSON, but it's not needed since I now use the already org.json dependency.

core/src/main/resources/xml/xsl/web/adapterSite.xsl
celmoussaoui: I'll hold this in mind when committing, my bad.

example/pom.xml
celmoussaoui: It is not necessary, so it will be deleted.

webapp/src/main/webapp/jsp/menuBar.jsp
celmoussaoui: Another tracing enter, deleted.

PR #162, opened 2019-06-18
Ibis description

celmoussaoui:
- Create a new annotation: IbisDescription
- Convert Javadoc into IbisDescription annotations
- Add the descriptions of classes to the IbisDoc


gvanbrakel:
Will be replaced by new IbisDoc / FrankDoc generation

PR #165, opened 2019-06-28
Dynamic iaftest

0851997:
Liquibase schema added for automatic creation of tables,
Edited expected results in order to resolve failed scenarios


0851997:
I resolved the issue. It wasn't necessary to change the expected results inside the test scenarios or ignore the precision or scale field attribute. The problem was in definition of types inside the liquibase schema itself.

0851997:
I changed the DirectoryListener to be fit for the purpose of creating multiple directories. I also created and added a configuration file which creates directories required by iaf-test for scenario testing.

gvanbrakel:
Zoals besproken:
Het is geen goed idee om de DirectoryListener aan te passen voor het creëren van een lijst subdirectories.
Beter is om de bestaande Ant scripts op te nemen in de pom file, zodat ze geautomatiseerd uitgevoerd kunnen worden.

0851997:
Ik heb gekeken naar de bestaande DirectoryListeners die misschien de directories konden aanmaken met behulp van createInputDirectory attribuut op true te zetten. Die heb ik niet tussen gezien. Directories worden blijbaar bij configuraties gemaakt dmv. een FilePipe. 

Waar het kon heb ik aanpassingen gedaan binnen de pipe en voor de andere gevallen heb ik binnen de configuratie een DirectoryListener aangemaakt die een directory listener maakt.

core/src/main/java/nl/nn/adapterframework/pipes/UnzipPipe.java
```diff
@@ -327,4 +330,14 @@ public void setCreateSubdirectories(boolean b) {
 	public boolean isCreateSubdirectories() {
 		return createSubdirectories;
 	}
+	
+	@IbisDoc({"if set <code>true</code>, validation on directory is ignored", "false"})
+	public boolean isTestDirectory()
+	{
+		return testDirectory;
+	}
+	public void setTestDirectory(boolean testDirectory)
```
nielsm5: Test directory sounds like it has something to do with testing, which can be a little bit confusing. Try and think more in the direction of checkDirectory or createDirectory(-IfNotExist)

core/src/main/java/nl/nn/adapterframework/util/Dir2Xml.java
```diff
@@ -63,13 +62,24 @@ public String getDirList(boolean includeDirectories, int maxItems) {
...
 		for (int i = 0; i < count; i++) {
 			File file = files[i];
 			if (file.isDirectory() && !includeDirectories) {
+				numberOfDirectories++;
 				continue;
 			}
 			dirXml.addSubElement(getFileAsXmlBuilder(file,file.getName()));
 		}
+		
+		// TODO: implement includeDirectories on WildCardFilter
```
nielsm5: I remember discussing this particular problem and like the solution, you could however document it a little better next time :).

test/src/test/testtool/FilePipe/common.properties
```diff
@@ -7,26 +7,26 @@ java.FilePipe.className=nl.nn.adapterframework.senders.IbisJavaSender
 java.FilePipe.serviceName=ibis4test-FilePipe
 
 fs.insert.className=nl.nn.adapterframework.testtool.FileSender
-fs.insert.filename=${testdata.dir}/fp2/dummy.txt
+fs.insert.filename=${testdata.dir}/dummy.txt
```
nielsm5: Why did you remove the subfolders? Without it will be more difficult to maintain the code and see what file/folder belongs where...

0851997: Because there is nothing that creates fp2/ folder therefor starting iaf-test will result in one adapter/configuration that will not start properly. Also, I had previously created a DirectoryListener for that adapter/configuration but Gerrit didn't approve of that solution. So we decided on changing the folder where files are created.

core/src/main/java/nl/nn/adapterframework/pipes/UnzipPipe.java
0851997: I think I get where you are coming from. As for the class itself the attribute name is independent of the use case it was created for. Which means e.g. checkDirectory name is a better for the Class itself. 

I changed the attribute name to checkDirectory.

PR #174, opened 2019-07-26
Added a Javascript Sender

JHuibers:
Added a sender which can execute a javascript function from a given javascript file using J2V8.
- Added a JUnit test for this sender.
- Added a Larva scenario for this sender.


core/src/main/java/nl/nn/adapterframework/extensions/javascript/JavascriptEngine.java
```diff
@@ -0,0 +1,21 @@
+package nl.nn.adapterframework.extensions.javascript;
+
+import nl.nn.adapterframework.core.ISender;
+import nl.nn.adapterframework.core.ISenderWithParameters;
+import nl.nn.adapterframework.parameters.ParameterResolutionContext;
+
+public interface JavascriptEngine<E> {
```
nielsm5: Try and add more javadoc info to the interface. All implementations inherit the javadoc, which will make it much more readable and future proof!

core/src/main/java/nl/nn/adapterframework/extensions/javascript/Rhino.java
```diff
@@ -0,0 +1,41 @@
...
+	}
+	
+	public Context get() {
+		return cx;
+	}
+
+	@Override
+	public void registerCallback(ISender sender, ParameterResolutionContext prc) {
+		// TODO Auto-generated method stub
+
```
nielsm5: Since this is not implemented, maybe throw an error or log a warning?

core/src/main/java/nl/nn/adapterframework/senders/JavascriptSender.java
```diff
@@ -0,0 +1,161 @@
...
+	public void setjsFunctionName(String jsFunctionName) {
+		this.jsFunctionName = jsFunctionName;
+	}
+
+	public String getjsFunctionName() {
+		return jsFunctionName;
+	}
+	
+	@IbisDoc({"the name of the javascript engine to be used", "J2V8"})
+	public void setengineName(String engineName) {
```
nielsm5: Please use Camel-Case method names, ie; setEngineName

PR #189, opened 2019-09-10
Automatically wire Spring beans into the IbisApplicationContext

nielsm5:
This enables autowireable beans to be loaded in the IbisApplicationContext, as part of the Spring Lifecycle.
Enables the use of Spring Aware interfaces such as: ServletContextAware, InitializingBean, DisposableBean, to populate certain setters.

The Service, Controller and Repository annotations are not enabled, instead use IbisInitializer annotation.

Also added the DynamicRegistration interface, which is currently only used for registering Servlets. 
The CMIS module has implemented the IbisInitializer annotation and the DynamicRegistration.Servlet interface.
The following properties can be set to change the DynamicRegistration.Servlet registration behaviour:
servlet.<name>.transportGuarantee (forces https when set to CONFIDENTIAL)
servlet.<name>.securityroles (use the default IBIS roles or create your own)
servlet.<name>.urlMapping (path the servlet listens to)
servlet.<name>.loadOnStartup (automatically load or use lazyloading)


cmis/src/main/java/nl/nn/adapterframework/extensions/cmis/servlets/CmisLifecycleBean.java
```diff
@@ -0,0 +1,93 @@
...
+	 */
+	private CmisServiceFactory createServiceFactory() {
+		String className = "nl.nn.adapterframework.extensions.cmis.server.RepositoryConnectorFactory";
+
+		// create a factory instance
+		Object object = null;
+		try {
+			object = ClassLoaderUtil.loadClass(className).newInstance();
+		} catch (Exception e) {
+			log.warn("Could not create a services factory instance: "+ e.toString(), e);
```
gvanbrakel: log.warn("Could not create a services factory instance: "+ e.toString(), e);

gvanbrakel: : "+ e.toString() is dubbelop

core/src/main/java/nl/nn/adapterframework/configuration/IbisContext.java
```diff
@@ -34,6 +34,7 @@
 import nl.nn.adapterframework.extensions.graphviz.GraphvizEngine;
 import nl.nn.adapterframework.http.RestServiceDispatcher;
 import nl.nn.adapterframework.jdbc.migration.Migrator;
+import nl.nn.adapterframework.lifecycle.IbisApplicationContext;
```
gvanbrakel: Als dit de enige change in deze file is, is hij vast niet nodig

nielsm5: Jawel want hij is verplaatst naar een andere package

PR #196, opened 2019-09-13
Output streaming

gvanbrakel:


core/src/main/java/nl/nn/adapterframework/stream/MessageOutputStream.java
```diff
@@ -0,0 +1,160 @@
...
+
+import org.apache.commons.io.output.WriterOutputStream;
+import org.apache.log4j.Logger;
+import org.xml.sax.ContentHandler;
+
+import nl.nn.adapterframework.util.LogUtil;
+import nl.nn.adapterframework.util.StreamUtil;
+
+public class MessageOutputStream {
+	protected Logger log = LogUtil.getLogger(this);
```
nielsm5: Add more (trace) logging?

PR #245, opened 2019-11-04
added aop around for StreamingSenderBase (err. target argument)

robertkarajev:
Added an aop around of the sender base. But there is a problem with the target argument. You can read that problem in the ibis log. Here is the link that might be relevent: https://stackoverflow.com/questions/23780951/xlintinvalidabsolutetypename


gvanbrakel:
This PR has been replaced by subsequent ones for the same issue

PR #268, opened 2019-11-20
Larva refactor

mkmeral:
This pull request includes a new package nl.nn.adapterframework.larva based on nl.nn.adapterframework.testtool as well as changes to test module and webapp. The detailed explanation for changes/new files are below.

- nl.nn.adapterframework.larva.api.LarvaApi
  - This is a REST API endpoint for running Larva. Detailed explanations regarding endpoints can be found in Javadocs.
-  nl.nn.adapterframework.larva.controller.*
   - This package includes sender/listener controlling functions (initialization, read and write functions, etc.). It is based on old testtool.
- nl.nn.adapterframework.larva.test
  - IbisTester class will later be used by command line test execution for starting up IBIS and adapters.
- nl.nn.adapterframework.larva.Message
  - Basic class to create a message structure for MessageListener.
- nl.nn.adapterframework.larva.MessageListener
  - It is a basic logging class for TestTool, all the outputs from tests are sent here and reporting is done through this class.
- nl.nn.adapterframework.larva.ResultComparer
  - This class is based on old testtool. It also includes a new option "treatSlashesAsSame", If set to true it will ignore differences between forward and back slashes. This is introduced for avoiding problems with paths.
- nl.nn.adapterframework.larva.ScenarioTester
  - This is a class that extends Thread. It is used to execute a list of scenarios, and report back to the given MessageListener.
- nl.nn.adapterframework.larva.TestPreparer
  - This class is based on old testtool. It contains classes that are required to read and prepare test execution. It also introduces a new feature, variables in xml for testing. It will replace ${variable.name} with variable.name property that is set in .properties files.
- nl.nn.adapterframework.larva.TestTool
  - This is the central class for test execution, it contains runScenarios function for execution.
- Rest of classes in nl.nn.adapterframework.larva
  - They are copied directly from old testtool. The reason to copy and not use them is to seperate them as much as possible, giving more freedom for new features that might be added in the future. 
- webapp/iaf/views/Larva.html
  - It is the new html view for larva test tool that is based on Angular. It uses the same structure as all the other views.
- webapp/iaf/js/services.js
  - Adapted poller service to be able to use URL that are based on generators, so that it can poll a new URL each time based on the generator function.
- webapp/iaf/js/ 
  - Added a new controller for Larva view. 
- test/src/testtool/*
  - Adapted the tests to use new variable feature. Replaced all c:/temp with ${testdata.dir} property.


nielsm5:
Please don't forget to add the license information at the top of each java class
Could you improve the style of the new larva so it looks more like the old larva?
In the LarvaApi class, you're running the tests in a new thread, why?
Perhaps it's an idea to create a new log4j appender (or overwrite all others with a new log4j4ibis.xml file) and make the appender write to the sysout, instead of using system.out.println in your code? 

I've reverted the c:/temp vs ${testtool.dir} property to test the old larva and new larva (perhaps we can configure your ibis in such a way both scenario's also work on your machine?) I'd like to create a new pr for these changes.

mkmeral:
> In the LarvaApi class, you're running the tests in a new thread, why?

It's blocking method. I need to return a response to the html request, otherwise it would timeout. I start the testing process, and return 200 if thread has started without problems. Then, start polling.

> Perhaps it's an idea to create a new log4j appender (or overwrite all others with a new log4j4ibis.xml file) and make the appender write to the sysout, instead of using system.out.println in your code?

I forgot to clean the system outs probably. I don't use them.

> I've reverted the c:/temp vs ${testtool.dir} property to test the old larva and new larva (perhaps we can configure your ibis in such a way both scenario's also work on your machine?) I'd like to create a new pr for these changes.

I don't see any way to get the old larva to adapt. Although, @robertkarajev  was working on implementing the variables to the old larva as well, I don't know if it's finished or not. Maybe we could try using relative directories?

mkmeral:
> Please don't forget to add the license information at the top of each java class

Yep, will do.

> Could you improve the style of the new larva so it looks more like the old larva?

What do you mean? frontend style? or coding style?

nielsm5:
> > Please don't forget to add the license information at the top of each java class
> 
> Yep, will do.
> 
> > Could you improve the style of the new larva so it looks more like the old larva?
> 
> What do you mean? frontend style? or coding style?

Frontend style. When you open a report you get a lot of blue headers which you can open. Maybe it can be more like the old larva where you, by default, get some information (if all is good) and more info about any errors if any.

nielsm5:
Closing this PR;
Most backend changes have been merged over the years.
The frontend will need to be re-written in Angular (15+).

PR #283, opened 2019-11-25
Disable streaming XSLT

gvanbrakel:
This appeared to cause NullPointerExceptions in Xalans
TransformerHandlerImpl, therefor we disabled this setting.
Also because it is using more memory than expected, e.g.
ForEachChildElementPipe with elementXPathExpression still goes out of
memory on large files

### DIFF
--- a/ladybug/src/main/java/nl/nn/ibistesttool/IbisDebuggerAdvice.java
+++ b/ladybug/src/main/java/nl/nn/ibistesttool/IbisDebuggerAdvice.java
@@ -24,6 +24,7 @@ import org.aspectj.lang.ProceedingJoinPoint;
 
 import nl.nn.adapterframework.core.ICorrelatedPullingListener;
 import nl.nn.adapterframework.core.IExtendedPipe;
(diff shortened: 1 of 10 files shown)


ladybug/src/main/java/nl/nn/ibistesttool/IbisDebuggerAdvice.java
```diff
@@ -145,6 +150,7 @@ public Object debugSenderWithParametersInputOutputAbort(ProceedingJoinPoint proc
 	}
 
 	public Object debugStreamingSenderInputOutputAbort(ProceedingJoinPoint proceedingJoinPoint, String correlationId, String message, ParameterResolutionContext parameterResolutionContext, MessageOutputStream target) throws Throwable {
+		if (debug) System.out.println("debugStreamingSenderInputOutputAbort thread id ["+Thread.currentThread().getId()+"] thread name ["+Thread.currentThread().getName()+"] correlationId ["+correlationId+"]");
```
nielsm5: I dislike the use of printing to sysout, can a logger be used? Use trace/debug

PR #333, opened 2019-12-24
Xerces 1.1

mkmeral:
Adapted and implemented the new version of Xerces2-J (which supports XSD 1.1). The forked version of Xerces should be added to the repository. Also I'd recommend adding org.eclipse.wst.xml.xpath2.processor to the repository as well. 

Created new tests for testing some basic functionality such as override, redefine, assert, etc.

Current Status: Requires extra testing on Jenkins.


core/src/main/java/nl/nn/adapterframework/validation/XercesXmlValidator.java
```diff
@@ -257,23 +265,31 @@ public XercesValidationContext createValidationContext(IPipeLineSession session,
...
 		} catch (SAXException e) {
 			throw new ConfigurationException(logPrefix + "Cannot create schema", e);
 		}
 		try {
-			//validatorHandler.setFeature(NAMESPACES_FEATURE_ID, true);
 			validatorHandler.setFeature(VALIDATION_FEATURE_ID, true);
 			validatorHandler.setFeature(SCHEMA_VALIDATION_FEATURE_ID, true);
 			validatorHandler.setFeature(SCHEMA_FULL_CHECKING_FEATURE_ID, isFullSchemaChecking());
+
+			validatorHandler.setContentHandler(context.getContentHandler());
```
gvanbrakel: Why is this necessary now, and wasn't it before? Was it a bug?

core/pom.xml
```diff
@@ -193,7 +192,12 @@
 		<dependency>
 			<groupId>org.ibissource</groupId>
 			<artifactId>ibis-xerces</artifactId>
-			<version>2.12.0.1</version>
+			<version>2.12.0-xml-schema-1.1.1</version>
+		</dependency>
+		<dependency>
```
gvanbrakel: Where do we need the xpath2 processor, or where is it used?

core/src/main/java/nl/nn/adapterframework/validation/XercesXmlValidator.java
```diff
@@ -282,18 +298,28 @@ public ValidatorHandler getValidatorHandler(IPipeLineSession session, Validation
...
+		ValidationContext context = createValidationContext(session, rootValidations, invalidRootNamespaces);
+		try {
+			ValidatorHandlerImpl validatorHandler = getValidatorHandler(session, context);
+			SAXSource inputSource = XmlUtils.stringToSAXSource(input.toString(), true, false);
+			validatorHandler.validate(inputSource, null);
+		} catch (SAXException | IOException e) {
+			return finalizeValidation(context, session, e);
+		}
+		return finalizeValidation(context, session, null);
+	}
```
gvanbrakel: Why is it necessary to override validate() here?
If not necessary, please remove it.
If you want to keep it, please avoid assuming 'input' is a string.

core/src/main/java/nl/nn/adapterframework/validation/XercesXmlValidator.java
mkmeral: It was done before as well (see: AbstractXmlValidator.java:213) It lets us check known namespaces. Since I overrode validate function, I needed to re-set it.

core/pom.xml
mkmeral: It is used by Xerces, but it does not use any dependency management tool (like gradle or maven) so I had to add it for now. Since we will eventually add it to our repository, I can set the dependency on its pom file, and remove it from our pom.

core/src/main/java/nl/nn/adapterframework/validation/XercesXmlValidator.java
mkmeral: Because AbstractValidator's validate uses `createValidationContext `to get a XMLReader and uses `reader.parse` method to validate it. However, with new Xerces we are using validator handler, and it does not extend/implement XmlReader, so I could not return xml reader to use AbstractValidator's validate method, I needed to override it.
I will modify it to not assume input is a string.

gvanbrakel: OK, but this means this is only a partial implementation of XSD 1.1 support. The method createValidatingParser should also return a parser that validates agains XSD 1.1.
This is used for instance in the Json2XmlValidator, for conversions from XML to JSON.
To extend your test cases to this, I suggest you migrate the tests you have built in _XmlValidatorExtraTest_ to _XmlValidatorTestBase_. Then it will be used also in the Json2XmlValidatorTest, that uses the validating parser in Json2XmlValidator.

mkmeral: Restructured the validator. Now, we support full implementation of xsd 1.1

core/pom.xml
mkmeral: Removed xpath dependency from core.

core/src/main/java/nl/nn/adapterframework/validation/XercesXmlValidator.java
```diff
@@ -257,23 +274,39 @@ public XercesValidationContext createValidationContext(IPipeLineSession session,
 		result.init(schemasProvider, preparseResult.getSchemasId(), preparseResult.getNamespaceSet(), rootValidations, invalidRootNamespaces, ignoreUnknownNamespaces);
 		return result;
 	}
-	
 
-	public ValidatorHandler getValidatorHandler(IPipeLineSession session, ValidationContext context) throws ConfigurationException, PipeRunException {
+	@Override
+	public ValidatorHandler getValidatorHandler(IPipeLineSession session, ValidationContext context) throws ConfigurationException {
+		// User Impl version from xerces, because it has extra validate() function.
+		ValidatorHandlerImpl validatorHandler;
```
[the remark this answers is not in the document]
mkmeral: Changed it to validator handler.

core/src/main/java/nl/nn/adapterframework/validation/AbstractXmlValidator.java
```diff
@@ -117,17 +118,17 @@ public String getImportedNamespacesToIgnore() {
...
 	 * Configure the XmlValidator
-	 * 
+	 *
 	 * @throws ConfigurationException when:
-	 * <ul>
-	 *   <li>the schema cannot be found</li>
-	 *   <li><{@link #isThrowException()} is false and there is no forward defined for "failure"</li>
-	 *   <li>when the parser does not accept setting the properties for validating</li>
-	 * </ul>
+	 *                                <ul>
```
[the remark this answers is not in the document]
mkmeral: Intellij's (buggy?) beautify function. Fixed it.

gvanbrakel:
Thanks for these works, I included it in another merge. Therefore this Pull Request will be closed.

PR #367, opened 2020-01-17
Pgp pipes

mkmeral:
Created a new PGP pipe for encryption, decryption, signing and verification.
Details about the usage is in javadoc.

One problem is that the library used is not in Maven center, instead it's in JCenter so it needs to be either copied to our own repositories or add additional repository to pom file.


core/src/main/java/nl/nn/adapterframework/pgp/PGPAction.java
```diff
@@ -0,0 +1,91 @@
...
+			// Create configuration
+			KeyringConfigCallback callback = KeyringConfigCallbacks.withUnprotectedKeys();
+			if(secretPassword != null)
+				callback = KeyringConfigCallbacks.withPassword(secretPassword);
+
+			keyringConfig = KeyringConfigs.forGpgExportedKeys(callback);
+
+			// Add public keys
+			if(publicKey != null)
+				for(String s : publicKey)
```
nielsm5: When you nest operations (if/for) and you traverse more then 1 layer deep, it's more readable to use brackets...

core/src/main/java/nl/nn/adapterframework/pgp/PGPAction.java
```diff
@@ -0,0 +1,91 @@
...
+ * This is an abstraction of general pgp actions
+ * such as encryption, verification, etc.
+ * to be used for {@link nl.nn.adapterframework.pipes.PGPPipe}
+ *
+ * @author Murat Kaan Meral
+ */
+public abstract class PGPAction {
+
+	protected InMemoryKeyring keyringConfig;
+	private String[] publicKey;
```
nielsm5: shouldn't the name be the plural publicKeys?

core/src/main/java/nl/nn/adapterframework/pipes/PGPPipe.java
```diff
@@ -0,0 +1,181 @@
...
+	 */
+	private String[] recipients;
+	/**
+	 * Emails of the senders. This will be used to verify that all the senders have signed the given message.
+	 * If not set, and the action is verify; this pipe will validate that at least one person has signed.
+	 *
+	 * For signing action, it needs to be set to the email that was used to generate the private key
+	 * that is being used for this process.
+	 */
+	private String[] senders;
```
nielsm5: I'm afraid the name senders might confuse people... not sure if we should change this to something like verificationAddresses?

core/src/main/java/nl/nn/adapterframework/pipes/PGPPipe.java
```diff
@@ -0,0 +1,181 @@
...
+		Message message = new Message(input);
+		try {
+			OutputStream outputStream = pgpAction.run(message.asInputStream());
+			return new PipeRunResult(findForward("success"), outputStream);
+		} catch (Exception e) {
+			throw new PipeRunException(this, "Exception was thrown during PGPPipe execution.", e);
+		}
+	}
+
+	public void setAction(String action) {
```
nielsm5: Can you add the IbisDoc annotation for all setters, so the ibisdevelopers can see what all these things do?

core/src/test/java/nl/nn/adapterframework/pipes/PgpPipeTest.java
```diff
@@ -0,0 +1,203 @@
...
+
+	/**
+	 * Adds folder's path to every file in the given parameters.
+	 * @param param A list of files separated by semicolon.
+	 * @return A list of files separated by semicolon including the parent folder's path.
+	 */
+	private String addFolderPath(String param) {
+		if (param == null)
+			return null;
+		String[] keys = param.split(";");
```
nielsm5: You can also use a tokenizer, which trims the results, and gives you a iterator to directly iterate over :)

core/src/test/java/nl/nn/adapterframework/pipes/PgpPipeTest.java
```diff
@@ -0,0 +1,203 @@
...
+			} else {
+				throw e;
+			}
+		}
+	}
+
+	/**
+	 * Creates pipes and pipeline session base for testing.
+	 */
+	public void setup() {
```
nielsm5: Wont it be easier to use the Before annotation?

core/pom.xml
```diff
@@ -587,6 +587,12 @@
 			<version>2.3.5</version>
 			<scope>test</scope>
 		</dependency>
+		<!-- https://mvnrepository.com/artifact/name.neuhalfen.projects.crypto.bouncycastle.openpgp/bouncy-gpg -->
```
nielsm5: I'm still unsure what to do with this dependency.. since it's not on maven central nobody can build the framework without explicitly adding another repo :(

Maybe we can upload it to central, but then the groupid has to be org.ibissource (aka we need to fork it)

core/src/main/java/nl/nn/adapterframework/pgp/PGPAction.java
mkmeral: After you get used to python this feels easier :) but I changed it 👍

core/src/main/java/nl/nn/adapterframework/pgp/PGPAction.java
mkmeral: Yeah it should be, also the one in pgp pipe also should be publickeys

core/src/test/java/nl/nn/adapterframework/pipes/PgpPipeTest.java
mkmeral: Yep, and it also doesn't require me to hold all the keys in memory, but when I use it I can't set stringbuilder's capacity, which decreases performance since it needs to scale. So memory vs time :P . But with tokenizer it looks cleaner so I changed it.

core/src/test/java/nl/nn/adapterframework/pipes/PgpPipeTest.java
mkmeral: It does not matter from execution perspective. But yeah I guess @before is cleaner.

core/src/main/java/nl/nn/adapterframework/pipes/PGPPipe.java
mkmeral: Changed to verificationAddresses.

core/pom.xml
mkmeral: I have sent an email to the owner of the project, if he would rather upload to central himself or maybe he does not care, then we can publish it. I am waiting for an anwer, depending on that I will fork and create a pipeline on Jenkins to publish.

core/src/main/java/nl/nn/adapterframework/pgp/PGPAction.java
```diff
@@ -0,0 +1,111 @@
...
+			KeyringConfigCallback callback = KeyringConfigCallbacks.withUnprotectedKeys();
+			if (secretPassword != null)
+				callback = KeyringConfigCallbacks.withPassword(secretPassword);
+
+			keyringConfig = KeyringConfigs.forGpgExportedKeys(callback);
+
+			// Add public keys
+			if (publicKeys != null) {
+				for (String s : publicKeys) {
+					keyringConfig.addPublicKey(IOUtils.toByteArray(new FileInputStream(s)));
```
nielsm5: Can you use ClassUtils.getResourceURL ? it is very unlikely that the resource will be on the local filesystem. this generic method will use the appropiate classloader to retrieve the url

nielsm5: Is it also possible to retrieve these certificates from the JVM keystore? With a property to specify the name of the keystore to retrieve the certificate from

nielsm5:
It's been added to maven central! https://github.com/neuhalje/bouncy-gpg/issues/1

core/src/main/java/nl/nn/adapterframework/pgp/PGPAction.java
mkmeral: For getResourceURL should I use this.getClass().getClassLoader() as classloader or something else? 
Bouncy Castle and Java Keystore API are not compatible. If we need it, I could create an adapter to make them compatible. However, that would require rebuilding the keys each time. See [this url.](https://stackoverflow.com/questions/35928359/how-can-i-store-a-bouncy-castle-pgp-key-in-a-java-keystore)

nielsm5: > For getResourceURL should I use this.getClass().getClassLoader() as classloader or something else?
Have a look at ClassUtils.getResourceURL

> Bouncy Castle and Java Keystore API are not compatible. If we need it, I could create an adapter to make them compatible. However, that would require rebuilding the keys each time. See [this url.](https://stackoverflow.com/questions/35928359/how-can-i-store-a-bouncy-castle-pgp-key-in-a-java-keystore)
Fair enough, was a nice to have in my eyes!

mkmeral:
@nielsm5 I have changed parent pom and yaml file to use java 8, so that travic ci can pass. https://github.com/ibissource/iaf/pull/367/commits/1da221020ee2b462edf5f3ed79780d356536e221

core/src/main/java/nl/nn/adapterframework/pgp/PGPAction.java
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	/**
+	 * Runs the given action (which may be any extensions of this abstract class).
+	 *
+	 * @param inputStream Input for the action.
+	 * @return OutputStream that contains the encrypted/plaintext based on the action.
+	 * @throws Exception Any exception that can be thrown during the action.
+	 */
+	public abstract OutputStream run(InputStream inputStream) throws Exception;
```
gvanbrakel: The result of this action should be an InputStream, not an OutputStream.
A caller can use an OutputStream result to write to, whereas it should read from this.
In all your implementations the OutputStream is in fact a ByteArrayOutputStream, that is just a wrapper around the result which is the byte array.

PR #375, opened 2020-01-21
Fix for handmade inheritance IbisDoc

celmoussaoui:
Implementation of the new @IbisDocRef feature

- This new annotation allows one to refer to another @IbisDoc.
- The first value is order number of the method
- The second value is either the class name with the method or just the class name. In the latter case it will search in the specified class for a matching method name and take that @IbisDoc.
- Also separated the IbisDoc application into it's own class


core/src/main/java/nl/nn/adapterframework/filesystem/FileSystemSender.java
```diff
@@ -65,7 +66,8 @@
```
gvanbrakel: Zou je dezelfde aanpassingen ook willen doen voor de FileSystemPipe

core/src/main/java/nl/nn/adapterframework/doc/IbisDocRef.java
```diff
@@ -0,0 +1,31 @@
+/*
+   Copyright 2018 Integration Partners
```
gvanbrakel: Het is inmiddels 2020...

Issue #492, opened 2020-03-12
HttpSender ContentType getter method type is not the same as the setter


nielsm5:
**Describe the bug**
HttpSender ContentType getter method type is not the same as the setter

**Category**
 - Frank Framework

**To Reproduce**
Start ibis with httpsender and contenttype attribute

**Expected behavior**
No error message

**Environment:**
All

PR #493, opened 2020-03-12
Fix getter method type not being the same as the setter #492

nielsm5:


nielsm5:
I broke something in #420 :(

PR #505, opened 2020-03-16
Fix GUI 3.0 global console warnings

nielsm5:


nielsm5:
Also needs to be merged into 7.5-RC3-snapshot

PR #506, opened 2020-03-16
Fix configurations being reloaded due to not having a (valid) version

nielsm5:


nielsm5:
Must also be merged with 7.5-RC3-SNAPSHOT

PR #513, opened 2020-03-17
Doclet WIP

celmoussaoui:
- Created a new plugin called frankdoc-plugin for the doclet
- Added configuration for description and default value
- Still need to change configurations


gvanbrakel:
Will be replaced by new way of generating FrankDoc

PR #541, opened 2020-03-25
Message Serializable, and fix for MessageStoreSender

gvanbrakel:


core/src/main/java/nl/nn/adapterframework/pipes/MessageSendingPipe.java
```diff
@@ -530,6 +530,7 @@ public PipeRunResult doPipe(Message input, IPipeLineSession session, IOutputStre
 				return wrapResult;
 			} else {
 				input = Message.asMessage(wrapResult.getResult());
+				// TODO must preserve input here, if result can be an InputStream
```
nielsm5: Okey, so lets preserve it?

core/src/main/java/nl/nn/adapterframework/stream/Message.java
```diff
@@ -364,4 +379,17 @@ public static String asString(Object object, String defaultCharset) throws IOExc
 		return Message.asMessage(object).asByteArray(defaultCharset);
 	}
 	
+	private void writeObject(ObjectOutputStream stream) throws IOException {
```
nielsm5: Private method that's never been called?

core/src/main/java/nl/nn/adapterframework/stream/Message.java
```diff
@@ -36,8 +39,11 @@
 import nl.nn.adapterframework.util.StreamUtil;
 import nl.nn.adapterframework.util.XmlUtils;
 
-public class Message {
-	protected Logger log = LogUtil.getLogger(this);
+public class Message implements Serializable {
+
+	private static final long serialVersionUID = 1L;
```
nielsm5: What's the version?

core/src/main/java/nl/nn/adapterframework/stream/Message.java
gvanbrakel: Zo werkt dat met Serializable

core/src/main/java/nl/nn/adapterframework/stream/Message.java
gvanbrakel: ok, ok, ...

PR #546, opened 2020-03-28
Allow the use of custom code in configurations

nielsm5:
Allows the use of custom classes placed on the classpath of the application server, or when enabled, classes that are in a configuration (jar).

NOTE: When a custom class has been loaded in the JVM, you can no longer change it... Reloading the configuration will not reload the loaded java-class!


core/src/main/java/nl/nn/adapterframework/configuration/classloaders/ClassLoaderBase.java
```diff
@@ -217,6 +223,35 @@ public URL getResource(String name, boolean useParent) {
...
+		try {
+			return getParent().loadClass(name); // First try to load the class natively
+		} catch (Throwable t) { // Catch NoClassDefFoundError and ClassNotFoundExceptions
+			throwable = t;
+		}
+
+		String path = name.replace(".", "/")+".class";
+		URL url = null;
+		if(allowCustomClasses) {
+			url = getResource(name);
```
nielsm5: moet dit niet path zijn?

PR #550, opened 2020-03-30
Enable setting additional recipients for errors in the test module

ricardovh:
Extends https://github.com/ibissource/iaf/pull/526 to allow setting additional recipients that will only get the e-mail when tests fail.

### DIFF
--- a/test/src/main/configurations/UTIL/Larva/xsl/CreateMail.xsl
+++ b/test/src/main/configurations/UTIL/Larva/xsl/CreateMail.xsl
@@ -5,6 +5,8 @@
 	<xsl:param name="hostname"/>
 	<xsl:param name="message"/>
 	<xsl:param name="mailaddresses"/>
+	<xsl:param name="mailaddressesError"/>
+	
 	<xsl:variable name="from" select="$hostname"/>
 	<xsl:variable name="to" select="tokenize($mailaddresses,',')[1]"/>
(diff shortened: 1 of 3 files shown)


test/src/main/configurations/UTIL/Larva/xsl/CreateMail.xsl
```diff
@@ -16,11 +18,18 @@
 		</xsl:choose>
 	</xsl:variable>
 
+	<xsl:variable name="ccList">
+		<xsl:choose>
+			<xsl:when test="contains($message, 'All scenarios passed')"><xsl:value-of select="$cc"/></xsl:when>
```
gvanbrakel: Ricardo,
You now determine if there are failed tests by parsing the test results.
I would prefer that you configure a forward named 'fail' on the LarvaPipe. That forward is used when there are failed tests.

test/src/main/resources/DeploymentSpecifics.properties
```diff
@@ -106,4 +106,6 @@ jdbc.querylistener.key=260873
 MemoryLeak.active=false
 
 #needs a comma separated list with mailaddresses for the larva adapter
+#people in the .error-list are only mailed when tests fail
 errorreport.mailaddresses=
+errorreport.mailaddresses.error=
```
gvanbrakel: I would like to suggest to name these mailaddresses:
errorreport.mailaddresses.pass
errorreport.mailaddresses.fail

PR #560, opened 2020-03-31
Throw warning when registering the same realm twice

nielsm5:


nielsm5:
Problems have been fixed in pr's mentioned above. All that remains is notifying the user when a jmsRealm is overriden.

gvanbrakel (approved):

PR #562, opened 2020-03-31
resteasy was removed from core

ccamiletti:
I have commented AuthorizationFilter.java and ServletDispatcher.java classes, there were replaced by CXFAuthorizationFilter and CXFServletDispatcher.
Please, don't test nl.nn.adapterframework.webcontrol.api.Init service, i am still working on this.

I would like to test this services:
1.- nl.nn.adapterframework.webcontrol.api.SendJmsMessage
2.- nl.nn.adapterframework.webcontrol.api.ShowConfiguration (extends nl.nn.adapterframework.webcontrol.api.Base)

3.- nl.nn.adapterframework.webcontrol.api.ShowScheduler
4.- nl.nn.adapterframework.webcontrol.api.TestServiceListener
5.- nl.nn.adapterframework.webcontrol.api.TransactionalStorage


core/src/main/java/nl/nn/adapterframework/webcontrol/api/TransactionalStorage.java
```diff
@@ -445,17 +446,23 @@ public Response browsePipeMessages(
 		return Response.status(Response.Status.OK).entity(getMessages(storage, filter)).build();
 	}
 
-	private String[] getMessages(MultipartFormDataInput input) {
-		try {
-			Map<String, List<InputPart>> inputDataMap = input.getFormDataMap();
-			if(inputDataMap.get("messageIds") != null) {
-				String messageIds = inputDataMap.get("messageIds").get(0).getBodyAsString();
-				return messageIds.split(",");
+	private Optional<String> getAttributeValue(String name, List<Attachment> attachmentList) {
```
nielsm5, line 449: try to not duplicate code, this is already in the base class?

core/src/main/java/nl/nn/adapterframework/webcontrol/api/AuthorizationFilter.java
```diff
@@ -49,62 +34,57 @@
 @Produces(MediaType.APPLICATION_JSON)
 public class AuthorizationFilter implements ContainerRequestFilter {
 
-	private static final ServerResponse ACCESS_FORBIDDEN = new ServerResponse(null, 403, new Headers<Object>());
-	private static final ServerResponse SERVER_ERROR = new ServerResponse(null, 500, new Headers<Object>());
-	protected Logger log = LogUtil.getLogger(this);
-
+	@Override
```
nielsm5, line 37: If this file is no longer required, you can remove it.
Have you tested the authorizations on specific endpoints?

core/src/main/java/nl/nn/adapterframework/webcontrol/api/Base.java
```diff
@@ -186,10 +188,12 @@ protected String resolveStringFromMap(Map<String, List<InputPart>> inputDataMap,
...
 
-	protected <T> T resolveTypeFromMap(Map<String, List<InputPart>> inputDataMap, String key, Class<T> clazz, T defaultValue) throws ApiException {
+	@SuppressWarnings("unchecked")
+	protected <T> T resolveTypeFromMap(List<Attachment> inputDataMap, String key, Class<T> clazz, T defaultValue) throws ApiException {
 		try {
-			if(inputDataMap.get(key) != null) {
-				return inputDataMap.get(key).get(0).getBody(clazz, null);
+			String att = this.getAttributeValue(key, inputDataMap).orElse(null);
+			if (null != att) {
+				return (T) att;
```
nielsm5, line 196: You cannot just cast a String to an object

core/src/main/java/nl/nn/adapterframework/webcontrol/api/Base.java
```diff
@@ -199,4 +203,17 @@ protected String resolveStringFromMap(Map<String, List<InputPart>> inputDataMap,
 		}
 		throw new ApiException("Key ["+key+"] not defined", 400);
 	}
+	
+	private Optional<String> getAttributeValue(String name, List<Attachment> attachmentList) {
+		return attachmentList.stream().filter(att -> name.equalsIgnoreCase(att.getDataHandler().getDataSource().getName())).findAny().map(m -> {
```
nielsm5, line 208: This has become pretty unreadable. Try to avoid using lambda's unless they are short and simple.

core/src/main/java/nl/nn/adapterframework/webcontrol/api/CXFAuthorizationFilter.java
```diff
@@ -0,0 +1,109 @@
+/*
+Copyright 2016-2017 Integration Partners B.V.
```
nielsm5, line 2: Make sure the copyright notice is correct and up to date :)
If you work on a class you need to add the current year to it.

core/src/main/java/nl/nn/adapterframework/webcontrol/api/SendJmsMessage.java
```diff
@@ -212,4 +192,51 @@ private void processMessage(JmsSender qms, String messageId, String message) thr
...
 	}
+	
+	
+	private Optional<String> getAttributeValue(String name, List<Attachment> attachmentList) {
+		return this.getFile(name, attachmentList).map(v -> {
+			return Optional.ofNullable(v.toString());
+		}).orElse(Optional.empty());
+	}
+	
+	private Optional<InputStream> getFile(String name, List<Attachment> attachmentList) {
```
nielsm5, line 203: I would expect an inputstream or null, there's nothing optional about that?

core/src/main/java/nl/nn/adapterframework/webcontrol/api/SendJmsMessage.java
```diff
@@ -212,4 +192,51 @@ private void processMessage(JmsSender qms, String messageId, String message) thr
...
+		return attachmentList.stream().filter(att -> name.equalsIgnoreCase(att.getDataHandler().getDataSource().getName())).findAny().map(m -> {
```
nielsm5, line 204: This written out in normal style java:

for(Attachment attachment: attachmentList) {
  if(name == attachment.getName()) return attachment.getInputStream();
}

reads a lot nicer...

core/src/main/java/nl/nn/adapterframework/webcontrol/api/AuthorizationFilter.java
ccamiletti: no, i don't know how to test it... can you tell me how please.?

core/src/main/java/nl/nn/adapterframework/webcontrol/api/TransactionalStorage.java
ccamiletti: to get an attribute from the new Multipart object is not the same way than "multipart" from "restEasy".

core/src/main/java/nl/nn/adapterframework/webcontrol/api/Base.java
ccamiletti: i already fixed it :)

core/src/main/java/nl/nn/adapterframework/webcontrol/api/Base.java
ccamiletti: I have removed some lambda expression, now it is more readable. let me know what do you think about the new code please

core/src/main/java/nl/nn/adapterframework/webcontrol/api/SendJmsMessage.java
ccamiletti: it is an inputStream wrapped in Optional. it meas null or a inputStream class

core/src/main/java/nl/nn/adapterframework/webcontrol/api/SendJmsMessage.java
ccamiletti: we need to take advantage from stream, in this case, besides to iterate, i need to filter by "name". also, using optional we can avoid nullpionterexception :)

PR #572, opened 2020-04-02
XML Query Output Transformer

mkmeral:
Created an abstract transformer extending XMLFilter, and one implementation (so far).


core/src/main/java/nl/nn/adapterframework/webcontrol/api/ExecuteJdbcQuery.java
```diff
@@ -130,30 +129,23 @@ public Response execute(LinkedHashMap<String, Object> json) throws ApiException
 			qs.setBlobSmartGet(true);
 			qs.configure(true);
 			qs.open();
-			result = qs.sendMessage(new Message(query), null).asString();
+			Message message = qs.sendMessage(new Message(query), null);
+			result = message.asString();
```
nielsm5, line 133: Could you turn the resultType in an if else case and if none matches do asString?

core/src/test/java/nl/nn/adapterframework/jdbc/transformer/toJson.java
```diff
@@ -0,0 +1,55 @@
...
+			xml = new File(FOLDER, i + ".xml");
+			json = new File(FOLDER, i + ".json");
+		}
+
+		return files;
+	}
+
+	@Test
+	public void doTest() throws IOException, SAXException {
+		String expected = new String(Files.readAllBytes(Paths.get(expectedFile.getAbsolutePath())));
```
nielsm5, line 48: We have also made a helper class for this: TestFileUtils.getTestFile(URL)

core/src/main/java/nl/nn/adapterframework/jdbc/transformer/QueryOutputToCSV.java
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	@Override
+	protected void addFieldDefinition(Attributes atts) {
+		output.append("\"").append(atts.getValue("name")).append("\",");
+	}
+
+	@Override
+	protected void addField(String fieldName, String value) {
+		output.append("\"").append(value).append("\",");
```
gvanbrakel, line 34: I suggest you use org.apache.commons.lang.StringEscapeUtils.escapeCSV() here.

PR #600, opened 2020-04-10
Pipe Test Classes

sinasen2000:
Pull Request for PipeTests notes:
- Overall problem: Based on my observations, some classes are highly dependent on each other, this makes it hard to test that single class as you need to change/set the parameters of lots of other classes in order to make the class you are working function normally. This is not ideal as high class coupling reduces ability for modifications and prevents all relevant code to be in one place, thus resulting in difficulty in testing.

For example;
- Message sending pipe: This class is extremely hard to test as there are endless number of branches and blob code. I think it will be better for me to test it after I get a nearly perfect understanding of the whole system. Otherwise, it would be a blind testing for me and I can miss some errors or bugs. Plus, in addition to above, it is dependent on lots of other classes which the user should modify, namely: ParameterResolutionContext, ClassUtils, ISenderWithParameters (which is an interface so the user or extended classes define which one to use. This means it's harder for the tester

- FileLineIteratorTest: This class is also highly dependent on message sending pipe, I tested its methods that are not dependent on message sending pipe. However, I think it is also better to test this one after testing Message Sending Pipe. 

- IsolatedServiceExecutor and IsolatedServiceCaller: These two use JavaListener class and ServiceDispatcher class and it requires modifying the parameters of that class, which is hard to do without testing that class. I tried out some things however couldn't find a way to register a listener in the JavaListener class. There's a registeredListeners map however this map is never used or never updated in a method. 

- Postbox Retriever Pipe: For testing this class, I needed to initialize a IPostBoxListener, which is an interface. The only class that implements this class was PullingJMSListener, so I started using that class as a base of my IPostBoxListener. However, during testing, I realized that in the configuration step of the listener, PullingJMSListener extends JMSListenerBase, which also extends JMSFacade. And in the configuration of JMSFacade a destination name and type must be specified however those fields inaccessible through IPostBoxRetriever because of the class hierarchy. 

- Rhino Pipe: When I test with multiple arguments, the pipe fails. I tried a few different to pass multiple arguments, like "3, 4", "3\n4", "3 4", etc, but all of them failed.

- CRL Pipe: I cannot find any implementations of the generateCRL method used in CRL Pipe. That's why I could only test bad weather cases. I also checked where CRL Pipes are used in the project to understand how it is used, but unfortunately it wasn't used anywhere. So I think it may be deprecated or not fully implemented yet?

- FixedForwardPipe: I think lines 85-86 and 87 are unreachable as the null check for that was done before. So those lines can be redundant

- StreamLineIteratorPipe: in the method getReader, the parameter threadContext is redundant, not used anywhere

Problems to resolve on my side:
- PasswordHashPipeTest: The validation algorithm is private so it cannot be accessed from the test. This raised the following question on my head; how to mimic the hashing in testing?
- Unzip Pipe: Somehow, my the zipinput stream treat text file as an empty file. I need to dive deeper into this problem.


core/src/test/java/nl/nn/adapterframework/pipes/FilePipeTest.java
```diff
@@ -0,0 +1,92 @@
...
+        PipeRunResult res = pipe.doPipe(var, session1);
+        assertEquals(res.getPipeForward().getName(), "success");
+    }
+
+    @Test
+    public void doTestFailAsEncodingNotSupportedBase64() throws Exception {
+        exception.expect(PipeRunException.class);
+        exception.expectMessage("Pipe [FilePipe under test] msgId [null] Error while executing file action(s): (UnsupportedEncodingException) /");
+        Parameter p = new Parameter();
+        p.setSessionKey("key"); p.setName("p1"); p.setValue("15"); p.setType("int"); p.configure();
```
nielsm5, line 80: Please put all the setters underneath each other for better readability.

core/src/main/java/nl/nn/adapterframework/http/AuthSSLProtocolSocketFactoryForJsse10x.java
```diff
@@ -101,7 +101,7 @@ public AuthSSLProtocolSocketFactoryForJsse10x(
         }
         log.debug("Initializing key manager");
         if (StringUtils.isEmpty(algorithm)) {
-        	algorithm=KeyManagerFactory.getDefaultAlgorithm();
+        	algorithm= KeyManagerFactory.getDefaultAlgorithm();
```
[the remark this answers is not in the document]
sinasen2000: This must be unintentional, I se that there are couple of files changed only with a space, I'm going to remove the changes.

core/src/main/java/nl/nn/adapterframework/pipes/FilePipe.java
```diff
@@ -54,7 +54,7 @@ public void configure() throws ConfigurationException {
 	 */
 	public PipeRunResult doPipe(Object input, IPipeLineSession session) throws PipeRunException {
 		try {
-			return new PipeRunResult(getForward(), fileHandler.handle(input, session, getParameterList()));
+ 			return new PipeRunResult(getForward(), fileHandler.handle(input, session, getParameterList()));
```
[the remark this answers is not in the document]
sinasen2000: > Try to not modify any files which don't have any functional changes.

This must be unintentional, I see that there are couple of files changed only with a space, I'm going to remove the changes.

core/src/main/java/nl/nn/adapterframework/pipes/MessageSendingPipe.java
```diff
@@ -258,6 +258,7 @@ public void addParameter(Parameter p){
 		}
 	}
 
+
```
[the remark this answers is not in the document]
sinasen2000: Same holds for this

core/src/test/java/nl/nn/adapterframework/pipes/ExceptionPipeTest.java
```diff
@@ -0,0 +1,54 @@
...
+import java.nio.channels.Pipe;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * ExceptionPipe Tester.
+ *
+ * @author <Sina Sen>
+ * @version 1.0
+ * @since <pre>Mar 5, 2020</pre>
```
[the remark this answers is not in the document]
sinasen2000: > We don't really use the since tag. When we do use these, we use the version number and not the date.
> For all these tests you can remove them, as they don't add any additional value to the file(s).

okay! I used them to index what  I did on which day, I'm going to remove them.

core/src/test/java/nl/nn/adapterframework/pipes/CompareStringPipeTest.java
```diff
@@ -14,24 +15,12 @@ public CompareStringPipe createPipe() {
...
     @Test(expected = ConfigurationException.class)
     public void emptySessionKeys() throws ConfigurationException {
+        PipeForward forw = new PipeForward("lessthan", "/Users/apollo11/Desktop/iaf2/core/src/test/resources/Pipes");
+        pipe.registerForward(forw);
         pipe.setSessionKey1("");
         pipe.setSessionKey2("");
         pipe.configure();
     }
 
-    @Test
```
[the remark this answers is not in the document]
sinasen2000: > Why did you removed these 2 tests?

Thank you for pointing this out, I actually don't have a clue and cannot recall removing them, as this test does not belong to me. I'm going to readd these and I will try to be more careful to avoid errors like this and double check my pull requests

core/src/test/java/nl/nn/adapterframework/pipes/ExceptionPipeTest.java
```diff
@@ -0,0 +1,54 @@
+package nl.nn.adapterframework.pipes;
+
+import nl.nn.adapterframework.core.PipeRunException;
+import org.junit.Test;
+import org.junit.Before; 
+import org.junit.After;
+
+import java.nio.channels.Pipe;
```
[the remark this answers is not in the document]
sinasen2000: > Try not to add any unused imports.
> You can have a look at Codacy (after pushing your changes, it will be available on your PR).

Yes! I looked at it and it indeed gives nice comments on the code quality. I'm going to adjust the code accordingly

core/src/test/java/nl/nn/adapterframework/pipes/FilePipeTest.java
sinasen2000: Thanks for the feedback, will do

core/src/test/java/nl/nn/adapterframework/pipes/IfMultipartTest.java
```diff
@@ -0,0 +1,93 @@
...
+        exception.expectMessage("Pipe [IfMultipart under test] expected HttpServletRequest as input, got [String]");
+        pipe.doPipe("i am a string not a http req", session);
+    }
+
+    @Test
+    public void testRequestUsesElseForward() throws Exception {
+        PipeForward forw = new PipeForward("custom_else", "random/path");
+        pipe.registerForward(forw);
+        pipe.setElseForwardName("custom_else");
+        assertEquals(pipe.doPipe(request, session).getPipeForward().getName().toString(), "custom_else");
```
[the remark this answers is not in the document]
sinasen2000: I couldn't understand what you meant to say

> This is a recipe for getting NullPointerExceptions

I could not understand what you meant to say. Can you explain a bit?

core/src/test/java/nl/nn/adapterframework/pipes/IncreaseIntegerPipeTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Test
+    public void testCannotIncreaseBy2AsNoSessionKey() throws Exception {
+        exception.expect(ConfigurationException.class);
+        exception.expectMessage("sessionKey must be filled");
+        session.put("a", "4");
+        pipe.setIncrement(2);
+        pipe.configure();
+        doPipe(pipe, "doesnt matter", session);
+        assertEquals(session.get("a"), "6");
```
nielsm5, line 49: This wont be triggered as you're expecting doPipe to throw an exception correct?

core/src/test/java/nl/nn/adapterframework/pipes/PasswordGeneratorPipeTest.java
```diff
@@ -0,0 +1,67 @@
...
+        pipe.setUCharacters("ASDF");
+        pipe.setSigns("!@#@");
+        pipe.setNumOfDigits(3);
+        pipe.setNumOfLCharacters(4);
+        pipe.setNumOfUCharacters(2);
+        pipe.setNumOfSigns(2);
+        pipe.setUseSecureRandom(false);
+        pipe.configure();
+        PipeRunResult res = doPipe(pipe, "pipey", session);
+        assertEquals(res.getResult().toString().length(), 19);
```
nielsm5, line 50: Try to use asString instead

core/src/test/java/nl/nn/adapterframework/pipes/PasswordHashPipeTest.java
```diff
@@ -0,0 +1,63 @@
...
+        session.put("key", "3:2342:2342" );
+        pipe.configure();
+        PipeRunResult res = doPipe(pipe, "password", session);
+        assertEquals("success", res.getPipeForward().getName());
+    }
+
+    @Test
+    public void testValidatePipe() throws Exception {
+
+        String sc = PasswordHash.createHash("password");
```
nielsm5, line 40: Maybe you can also convert the current PasswordHash tests (in the static main method of the class) to proper junit tests.

core/src/test/java/nl/nn/adapterframework/pipes/PutInSessionTest.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Test
+    public void testNoSessionKey() throws Exception {
+        exception.expectMessage("attribute sessionKey must be specified");
+        exception.expect(ConfigurationException.class);
+        pipe.setValue("val");
+        pipe.configure();
+        PipeRunResult res = pipe.doPipe(null, session);
+        assertFalse(res.getPipeForward().getName().isEmpty());
```
nielsm5, line 66: You're expecting doPipe to throw an exception? then this assertion is not required :)

core/src/test/java/nl/nn/adapterframework/pipes/RemoveFromSessionTest.java
```diff
@@ -0,0 +1,76 @@
...
+    @Test
+    public void testFailAsKeyIsWrong() throws Exception {
+        pipe.setSessionKey("ab");
+        session.put("a", "123");
+        PipeRunResult res = doPipe(pipe, "ab", session);
+        assertEquals(res.getResult().toString(), "String: [null]");
+
+    }
+
+
```
nielsm5, line 72: Try to use asString instead of toString

core/src/test/java/nl/nn/adapterframework/pipes/ReplacerPipeTest.java
```diff
@@ -0,0 +1,97 @@
...
+        pipe.setFind("test");
+        pipe.setReplace("head");
+        pipe.setReplaceNonXmlChar("klkl");
+        pipe.setReplaceNonXmlChars(true);
+        pipe.configure();
+        PipeRunResult res = doPipe(pipe, "<test>lolo</test>/jacjac:)", session);
+        assertEquals("<head>lolo</head>/jacjac:)", res.getResult().toString());
+    }
+
+
```
nielsm5, line 92: Try to use asString instead of toString

core/src/test/resources/Pipes/1.txt
```diff
@@ -0,0 +1,3 @@
+function giveNumber(x) {
+    return x * x;
+}
```
nielsm5, line 3: This is javascript, please name the file accordingly, eg. /javascript/rhino-tests.js

core/src/test/java/nl/nn/adapterframework/pipes/FixedResultTest.java
```diff
@@ -0,0 +1,150 @@
...
+        session1.put("param1","yarr");
+        pipe.addParameter(param);
+        pipe.setLookupAtRuntime(true);
+        pipe.setFileName(sourceFolderPath);
+        pipe.setReplaceFrom("param1");
+        pipe.setReplaceTo("kar");
+        pipe.setReturnString("${param1}andandandparam2");
+        pipe.configure();
+        PipeRunResult res = doPipe(pipe, "whatisthis", session1);
+        assertEquals(res.getResult().toString(), "String: inside the file");
```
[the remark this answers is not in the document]
sinasen2000: thanks for pointing out such a method exists. will replace all toString to asString where it's needed

core/src/test/java/nl/nn/adapterframework/pipes/PutInSessionTest.java
sinasen2000: In some cases, I added assertions or fail()s to decrease the amount of code quality warnings given by codacy.

core/src/test/java/nl/nn/adapterframework/pipes/ReplacerPipeTest.java
```diff
@@ -0,0 +1,97 @@
...
+        pipe.setReplace("head");
+        pipe.setReplaceNonXmlChar("klkl");
+        pipe.setReplaceNonXmlChars(true);
+        pipe.configure();
+        doPipe(pipe, "<test>lolo</test>/jacjac:)", session);
+        fail("this is expected to fail");
+    }
+
+
+
```
nielsm5, line 94: You've only added tests with setReplaceNonXmlChars set to true?

core/src/test/java/nl/nn/adapterframework/pipes/SkipPipeTest.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /**
+     * Method: setSkip(int skip)
+     */
+    @Test
+    public void testDoPipeSkip3Read2WithString() throws Exception {
+        pipe.setSkip(3); pipe.setLength(2);
+        PipeRunResult res = doPipe(pipe, "0123456", session);
+        assertEquals(res.getResult().asString(), "34");
```
nielsm5, line 33: Try to not test too many things in a single junittest
It will be harder to find issues if something might break in the future.
Typically in a junit test you test the smallest possible bit of code, making sure that the entire code as a whole keeps performing well.

core/src/test/java/nl/nn/adapterframework/pipes/Text2XmlPipeTest.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Test
+    public void testSuccessWithoutAdditionalProperties() throws Exception {
+        pipe.setXmlTag("address");
+        pipe.setSplitLines(false);
+        pipe.setUseCdataSection(false);
+        pipe.setIncludeXmlDeclaration(true);
+        pipe.setReplaceNonXmlChars(false);
+        pipe.configure();
```
nielsm5, line 58: For a test that has no additional properties you set a lot of properties

core/src/test/java/nl/nn/adapterframework/pipes/Text2XmlPipeTest.java
```diff
@@ -0,0 +1,91 @@
...
+        PipeRunResult res = doPipe(pipe, "this is an example\nim in cdata", session);
+        assertEquals("<?xml version=\"1.0\" encoding=\"UTF-8\"?><address>this is an example\n" +
+                "im in cdata</address>", res.getResult().asString());
```
nielsm5, line 61: this is not cdata. See [wikipedia](https://en.wikipedia.org/wiki/CDATA) for more info

core/src/test/java/nl/nn/adapterframework/pipes/RemoveFromSessionTest.java
```diff
@@ -0,0 +1,76 @@
...
+            pipe.setSessionKey("a");
+            session.put("a", "123");
+            PipeRunResult res = pipe.doPipe(null, session);
+            assertEquals(res.getResult().asString(), "123");
+
+    }
+    @Test
+    public void testEmptySessionKeyEmptyInput() throws Exception {
+        exception.expect(NullPointerException.class);
+        pipe.setSessionKey(null);
```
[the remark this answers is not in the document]
sinasen2000: Thanks for pointing out, everyday is a learning day! :)

core/src/test/resources/Pipes/1.txt
sinasen2000: done

core/src/test/java/nl/nn/adapterframework/pipes/PasswordHashPipeTest.java
sinasen2000: Okay, I thought you were talking about the PasswordHashPipe rather than PasswordHash class. Checking it out now.

sinasen2000: Converted the main method into a test case.

core/src/test/java/nl/nn/adapterframework/pipes/ReplacerPipeTest.java
sinasen2000: yes indeed, adding a couple more cases with that method set to false. Thanks for pointing it out.

core/src/test/java/nl/nn/adapterframework/pipes/SkipPipeTest.java
sinasen2000: disected the test into smaller tests

core/src/test/java/nl/nn/adapterframework/pipes/Text2XmlPipeTest.java
sinasen2000: yes, changed it accordingly.

core/src/test/java/nl/nn/adapterframework/pipes/PasswordHashPipeTest.java
```diff
@@ -0,0 +1,67 @@
...
+        assertEquals("success", res.getPipeForward().getName());
+    }
+    @Test
+    public void testValidatePipeFailAsNotTheSame() throws Exception {
+
+        session.put("key", "2:22:22");
+        pipe.setHashSessionKey("key");
+        pipe.configure();
+        PipeRunResult res = doPipe(pipe, "password", session);
+        assertEquals(null, res.getPipeForward());
```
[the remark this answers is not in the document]
sinasen2000: I tested this class without knowing that I can use PasswordHash class. Now I am going to refactor this test class as a whole.

core/src/test/java/nl/nn/adapterframework/pipes/PasswordHashPipeTest.java
```diff
@@ -0,0 +1,67 @@
...
+        return new PasswordHashPipe();
+    }
+
+
+    /**
+     * Method: doPipe(Object input, IPipeLineSession session)
+     */
+    @Test
+    public void testHashPipe() throws Exception {
+        session.put("key", "3:2342:2342" );
```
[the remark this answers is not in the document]
sinasen2000: I tested this class without knowing that I can use PasswordHash class. Now I am going to refactor this test class as a whole.

core/src/test/java/nl/nn/adapterframework/pipes/MoveFilePipeTest.java
```diff
@@ -0,0 +1,366 @@
...
+        pipe.configure();
+        pipe.start();
+
+        pipe.doPipe(null, session);
+
+        fail("this is expected to fail");
+    }
+
+    @Test
+    @Order(11)
```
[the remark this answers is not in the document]
sinasen2000: done!

core/src/test/java/nl/nn/adapterframework/pipes/FixedResultTest.java
```diff
@@ -0,0 +1,149 @@
...
+        pipe.addParameter(param);
+        pipe.setLookupAtRuntime(true);
+        pipe.setFileName(sourceFolderPath);
+        pipe.setReplaceFrom("param1");
+        pipe.setReplaceTo("kar");
+        pipe.setReturnString("${param1}andandandparam2");
+        pipe.configure();
+        PipeRunResult res = doPipe(pipe, "whatisthis", session1);
+        assertEquals(res.getResult().asString(), "inside the file");
+    }
```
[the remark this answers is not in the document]
sinasen2000: I copied the identical setters to another seperate method which is called by the other tests. I will  add some comments to the test as well. Do you want me to document the FixedResultPipe as well?

PR #602, opened 2020-04-14
configuration to set proxy to AmazonS3FileSystem class was added

ccamiletti:


core/src/main/java/nl/nn/adapterframework/filesystem/AmazonS3FileSystem.java
```diff
@@ -110,7 +114,8 @@ public void open() {
 		AmazonS3ClientBuilder s3ClientBuilder = AmazonS3ClientBuilder.standard()
 				.withChunkedEncodingDisabled(isChunkedEncodingDisabled())
 				.withForceGlobalBucketAccessEnabled(isForceGlobalBucketAccessEnabled()).withRegion(getClientRegion())
-				.withCredentials(new AWSStaticCredentialsProvider(awsCreds));
+				.withCredentials(new AWSStaticCredentialsProvider(awsCreds))
+				.withClientConfiguration(this.getProxyConfig());
```
gvanbrakel, line 118: I can imagine that you should only add the client configuration if the proxy configuration has really been set,
and that if no proxy is configured, it should not be added.

core/src/main/java/nl/nn/adapterframework/filesystem/AmazonS3FileSystem.java
```diff
@@ -603,5 +608,19 @@ public void setBucketCreationEnabled(boolean bucketCreationEnabled) {
...
+	}
+
+	public void setProxyConfig(String proxyHost, Integer proxyPort) {
+		if (proxyHost != null && proxyPort != null) {
+			proxyConfig = new ClientConfiguration();
+			proxyConfig.setProtocol(Protocol.HTTPS);
+			proxyConfig.setProxyHost(proxyHost);
+			proxyConfig.setProxyPort(proxyPort);
+		}
+	}
```
gvanbrakel, line 623: You'll have to create two separate setters for this: one for proxyHost and one for proxyPort.
Creating these two setters results in the possibility to set values in the configuration.xml, like:
proxyHost="hq.proxy.insim.biz" proxyPort="9400"

test-integration/src/test/java/nl/nn/adapterframework/filesystem/AmazonS3SenderTest.java
```diff
@@ -78,6 +81,7 @@ public AmazonS3Sender createFileSystemSender(){
 		s3.setAccessKey(accessKey);
 		s3.setSecretKey(secretKey);
 		s3.setBucketName(bucketName);
+		s3.setProxyConfig(proxyHost, proxyPort);
 		return s3;
 	}
```
gvanbrakel, line 87: Could you show in a test that it works both with and without a proxy configured?

core/src/main/java/nl/nn/adapterframework/filesystem/AmazonS3FileSystem.java
ccamiletti: if getProxyConfig is null, the connection will be without proxy. we are not going to get a nullPointerException o something like that.

core/src/main/java/nl/nn/adapterframework/filesystem/AmazonS3FileSystem.java
ccamiletti: ok, i will change it

PR #604, opened 2020-04-15
Apply necessary changes for ApiPrincipal Hardening, #595

### LINKED ISSUE #595: ApiPrincipalPipe hardening
**Describe the bug**
- Cookie not Sent Over SSL
- Cookie: HTTPOnly not Set
- Insecure Randomness

If the SSL flag is set, the browser will only send the cookie over HTTPS
The HttpOnly cookie property prevents client-side scripts from accessing the cookie.
new Random() is not a true random. 

**Category**
 - Frank Framework

**Expected behavior**
calling setSecure(true). (only when transportGuarantee == confidential)
calling setHttpOnly(true);
Using java.security.SecureRandom instead of java.util.Random

mkmeral:
Repalced random with Secure Random
Added httpOnly
And refactored servletManager to be able to retrieve transport guarantee.
Added useSecure() if transport guarantee is confidential.

Closes #595

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/lifecycle/ServletManager.java
+++ b/core/src/main/java/nl/nn/adapterframework/lifecycle/ServletManager.java
@@ -93,7 +93,7 @@ public class ServletManager {
 			return;
 
 		ServletRegistration.Dynamic serv = getServletContext().addServlet(servletName, servletClass);
-		ServletSecurity.TransportGuarantee transportGuarantee = getTransportGuarantee(appConstants, propertyPrefix+"transportGuarantee");
+		ServletSecurity.TransportGuarantee transportGuarantee = getTransportGuarantee(propertyPrefix+"transportGuarantee");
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/nl/nn/adapterframework/lifecycle/ServletManager.java
```diff
@@ -155,4 +144,17 @@ public void register(DynamicRegistration.Servlet servlet) {
 
 		registerServlet(servlet.getName(), servlet.getServlet(), servlet.getUrlMapping(), servlet.getRoles(), servlet.loadOnStartUp(), parameters);
 	}
+
+	public static ServletSecurity.TransportGuarantee getTransportGuarantee(AppConstants appConstants, String propertyName) {
```
nielsm5, line 148: Since AppConstants is a static instance you can just refer to it by calling AppConstants.getInstance()
Saves this additional parameter, which is always 'AppConstants.getInstance()'

Issue #629, opened 2020-04-22
When you open a logfile in Google Chrome, you need to refresh the page before you see it


mhdirkse:
**Describe the bug**
When you open a logfile in Google Chrome, you see an animation that the logfile is loading. The logfile does not appear then, unless you refresh your browser page. Only after refreshing, the logfile appears.

**Category**
 - Frank Console

**To Reproduce**
1. Start the Frank!Runner again. To produce this error, the Frank!Runner was executed earlier resulting in some log files.
2. Open Google Chrome and browse to http://localhost.
3. In the main menu, click "Logging".
4. Click some logfile.
5. An animation is shown that the logfile is being loaded. That animation does not go away and the logfile does not appear yet.
6. Refresh your browser. Now the logfile appears.

**Expected behavior**
Refreshing your browser should not be necessary for the logfile to appear.

**Screenshots**

**Environment:**
* Frank!Runner
* Google Chrome
* Frank!Framework version 7.6-20200422.121508

**Additional context**
Add any other context about the problem here. (f.e. ladybug report / test adapter with larva test)

nielsm5:
I was finally able to reproduce this by using FireFox. The problem is due to the cpu cycles being faster then a 50ms timeout that was previously set, causing a property to be set too late and therefore not being called. Added it to #643 as well!

mhdirkse:
Verified.

Issue #632, opened 2020-04-23
When in the Frank!Console you click the parent directory of the logfile, you get locked in an error


mhdirkse:
**Describe the bug**
When you click "Logging" in the main menu of the Frank!Framework, you see the directory listing of the directory containing the logfiles. This directory listing allows you to click the parent directory named `..`, like it allows you to click a file. However, when you click `..`, you get the screen shown below:

![image](https://user-images.githubusercontent.com/6784732/80089725-a632f980-855e-11ea-8056-c241b3bcb739.png)

There are no files or directories to click and an "unknown error" has occurred. You cannot go back to the list of logfiles easily. In the end I could do this by going to the Adapter Status page, refreshing my browser window and then returning to Logging.

**Category**
 - Frank Framework

**To Reproduce**
1. Start the Frank!Runner.
2. Use Google Chrome to browse to http://localhost.
3. In the main menu click "Logging".
4. In the directory listing that comes up, click `..`.

**Expected behavior**
The `..` entry should not be clickable.

**Screenshots**
Already shown.

**Environment:**
 - Frank!Runner
 - Google Chrome
 - IAF Version 7.6-20200422.121508

**Additional context**
I discovered this bug while writing the Frank!Manual.

mhdirkse:
Verified.

Issue #631, opened 2020-04-23
When you open a logfile within the Frank!Console, you do not see the filename


mhdirkse:
**Is your feature request related to a problem? Please describe.**
When you open a logfile within the Frank!Console, you do not see the filename. You only see the directory of the file as shown below:

![image](https://user-images.githubusercontent.com/6784732/80083203-a084e600-8555-11ea-9250-6281d1c8f8f0.png)

**Describe the solution you'd like**
After the word "Logging", you already see the directory name. I recommend adding the filename to this.

**Describe alternatives you've considered**

**Additional context**
I created this issue when working on the Frank!Manual. In many detailed screens, you get feedback on where you have navigated to. An example that I like is Ladybug. When you select some node in a test report to the bottom-left panel, the bottom-right shows detailed information. One of these details is the path you selected in the test report. In the detailed view to the bottom-right, you have feedback on where you are: the selected node in the bottom-left panel. This feedback is missing in the Logging pages, because you do not see the name of the file you just opened.

mhdirkse:
Now I can see the filename. Thanks.

Issue #628, opened 2020-04-22
When you open the Frank!Console in Microsoft Edge, you cannot see logfiles


mhdirkse:
**Describe the bug**
When you open the Frank!Console in Microsoft Edge, you cannot see logfiles. In Google Chrome you can.

**Category**
 - Frank Console

**To Reproduce**
1. Start the Frank!Runner again. To produce this error, the Frank!Runner was executed earlier resulting in some log files.
2. Open Microsoft Edge and browse to http://localhost.
3. In the main menu, click "Logging".
4. Click some logfile.
5. An animation is shown that the logfile is being loaded. That animation does not go away, also not when you refresh your browser. The logfile does not appear.

**Expected behavior**
The chosen logfile should appear.

**Screenshots**
![image](https://user-images.githubusercontent.com/6784732/80001637-e0e55500-84be-11ea-9b61-0e11a08c82c1.png)

**Environment:**
 - Frank!Runner
 - Microsoft Edge
 - Frank!Framework version 7.6-20200422.121508

**Additional context**
Add any other context about the problem here. (f.e. ladybug report / test adapter with larva test)

nielsm5:
I think this is a duplicate of #629 where certain callbacks are not being 'called'.

mhdirkse:
I verified that the issue has been fixed. Thanks.

PR #644, opened 2020-04-27
GUI 3.0 Logging bugfixes

nielsm5:


nielsm5:
Fixes #629, #632, #628, #631 as described in PR #643

PR #688, opened 2020-05-08
fix #645 avoid CXF publish check failure on WebSpere

### LINKED ISSUE #645: CXF security permissions
**Describe the bug**
If you want to use CXF on WebSphere, with Java2Security enabled, `com.ibm.websphere.java2secman.norethrow = false`, it does not allow you to register CXF endpoints, causing the ibis to fail to initialize.

The fix is to set the property: `org.apache.cxf.jaxws.checkPublishEndpointPermissionWithSecurityManager=false`
We might need to automatically set this!??

**Category**
 - Frank Framework

**To Reproduce**
Start an Ibis on WebSphere

**Environment:**
WebSphere

**Additional context**
https://issues.apache.org/jira/browse/CXF-4085

gvanbrakel:


core/src/main/resources/AppConstants.properties
```diff
@@ -102,6 +102,8 @@ browse.messages.max = 100
 browse.messages.order = DESC
 browse.errors.order = ASC
 
+http.publish.checkPermission = false
```
nielsm5, line 105: http.cxf.checkPublishPermission

PR #716, opened 2020-05-14
Merge 4 Util test classes

sinasen2000:
Possible changes for Misc class:
- :  Changing the naming of listToString method to stringListToString, or even better, add support for other types such as integers.
- : In cleanseMessage, the parameter "hideRegex" should be renamed. It can be understood as a boolean to hide the regex portions in the string or not. A possible renaming can be "regexForCleaning" or "regexForHiding".


core/src/test/java/nl/nn/adapterframework/util/CounterStatisticTest.java
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     *
+     * Method: performAction(int action)
+     *
+     */
+    @Test
+    public void testPerformAction34() throws Exception {
+        CounterStatistic cs = new CounterStatistic(10);
+        cs.performAction(3);
+        assertEquals(cs.getValue(), 10);    }
```
nielsm5, line 48: code style dictates that the closing bracket, when present should be on a new line.

Applies in multiple tests

core/src/test/java/nl/nn/adapterframework/util/SizeLimitedVectorTest.java
```diff
@@ -0,0 +1,41 @@
...
+     */
+    @Test
+    public void testSetMaxSize() throws Exception {
+        SizeLimitedVector slv = new SizeLimitedVector(10);
+        slv.setMaxSize(14);
+        assertEquals(slv.getMaxSize(), 14);
+    }
+
+
+}
```
nielsm5, line 41: Try to also test the function of this class. set a limit, add limit +1 messages to the vector.
Test the size
Test that the first item has been removed

core/src/test/java/nl/nn/adapterframework/util/CounterStatisticTest.java
sinasen2000: changed it

core/src/test/java/nl/nn/adapterframework/util/SizeLimitedVectorTest.java
sinasen2000: If I understood what you meant correctly, now it is tested

core/src/test/java/nl/nn/adapterframework/util/SizeLimitedVectorTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Test
+    public void testMaxSizePassed() throws ArrayIndexOutOfBoundsException {
+        try{SizeLimitedVector slv = new SizeLimitedVector(1);
+        slv.add(13);
+        slv.add(14);
+        }
+        catch (ArrayIndexOutOfBoundsException e) {
+            fail("array capacity is passed, capacity was 1, but 2 elements got added");
+        }
```
nielsm5, line 39: What exactly are you testing here? this catch clause should never happen?
Maybe add some test assertions :)

core/src/test/java/nl/nn/adapterframework/util/FilenameUtilsTest.java
```diff
@@ -0,0 +1,272 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+
+/**
+ * FilenameUtils Tester.
+ *
+ * @author <Sina Sen>
+ */
+public class FilenameUtilsTest {
```
nielsm5, line 13: This class didn't require any tests. It's a utilities class copied from apache commons io.
The only reason it's here is in order to avoid websphere compilations and runtime issues.

core/src/test/java/nl/nn/adapterframework/util/FilenameUtilsTest.java
```diff
@@ -0,0 +1,272 @@
...
+
+
+
+    /**
+     * Method: isSystemWindows()
+     */
+    @Test
+    public void testIsSystemWindows() throws Exception {
+        assertEquals(FilenameUtils.isSystemWindows(), true);
+    }
```
nielsm5, line 23: You might run on windows, but many of our servers do not :) this will fail on Travis and Jenkins

core/src/test/java/nl/nn/adapterframework/util/SizeLimitedVectorTest.java
sinasen2000: why shouldn't the catch clause happen? The second add operation passes the size of the vector thus raises and AIOoBounds exception. I'm testing the case that if one adds more items to the vector than the size obliges, what happens next. In the original method .add there was not such a case that raises an expception so I used my own one.

core/src/test/java/nl/nn/adapterframework/util/FilenameUtilsTest.java
sinasen2000: yes indeed, I removed this test case.

core/src/test/java/nl/nn/adapterframework/util/SizeLimitedVectorTest.java
nielsm5: The whole point of the SizeLimitedVector is that you can append as many things to it as you like but it only stores the last X amount.

core/src/test/java/nl/nn/adapterframework/util/SizeLimitedVectorTest.java
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     *
+     * Method: setMaxSize(int maxSize)
+     *
+     */
+    @Test
+    public void testSetMaxSize() throws Exception {
+        SizeLimitedVector slv = new SizeLimitedVector(10);
+        slv.setMaxSize(14);
+        assertEquals(slv.getMaxSize(), 14);
```
nielsm5, line 51: Please also add 14+1 elements and verify more then just the setter being changed.
e.g. calling getSize or capacity to verify the contents of the vector.

core/src/test/java/nl/nn/adapterframework/util/SizeLimitedVectorTest.java
```diff
@@ -0,0 +1,55 @@
...
+public class SizeLimitedVectorTest {
+
+
+    /**
+     *
+     * Method: add(Object o)
+     *
+     */
+    @Test
+    public void testAdd() throws ArrayIndexOutOfBoundsException {
```
nielsm5, line 22: You dont have to declare this RuntimeException here do you? Non of the functions are throwing any exceptions

core/src/test/java/nl/nn/adapterframework/util/CounterStatisticTest.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+    
+
+    /**
+     *
+     * Method: performAction(int action)
+     *
+     */
+    @Test
+    public void testPerformAction34() throws Exception {
```
nielsm5, line 46: Try to use more descriptive names.  testAction1, testAction2, etc. is a bit vague when you're going through the log files :)

core/src/test/java/nl/nn/adapterframework/util/SizeLimitedVectorTest.java
sinasen2000: Yes. I am going to remove it, thanks for pointing out.

core/src/test/java/nl/nn/adapterframework/util/CounterStatisticTest.java
sinasen2000: Indeed, the naming I used looks terrible, I used numbers to reflect the action with number given in the CounterStatistic class, however using the action names rather than the number will be more descriptive. I am changing that.

core/src/test/java/nl/nn/adapterframework/util/FilenameUtilsTest.java
sinasen2000: okay, more communication will be assured on testing new classes.

core/src/test/java/nl/nn/adapterframework/util/SizeLimitedVectorTest.java
sinasen2000: yes, I changed the test accordingly.

core/src/test/java/nl/nn/adapterframework/util/SizeLimitedVectorTest.java
```diff
@@ -0,0 +1,55 @@
...
+     */
+    @Test
+    public void testSetMaxSize() throws Exception {
+        SizeLimitedVector slv = new SizeLimitedVector(5);
+        slv.setMaxSize(6);
+        slv.add(1); slv.add(2); slv.add(3); slv.add(4); slv.add(5); slv.add(6); slv.add(7);
+        assertEquals(slv.getMaxSize(), 6);
+        assertEquals(slv.capacity(), 10);
+        assertEquals(slv.size(), 6);
+        assertEquals(slv.get(0), 2);
```
nielsm5, line 51: 👍

core/src/test/java/nl/nn/adapterframework/util/CounterStatisticTest.java
```diff
@@ -0,0 +1,67 @@
...
+     *
+     * Method: performAction(int action)
+     *
+     */
+    @Test
+    public void testPerformActionMarkMainOrFull() throws Exception {
+        CounterStatistic cs = new CounterStatistic(10);
+        cs.performAction(3);
+        assertEquals(cs.getValue(), 10);
+    }
```
nielsm5, line 50: You are testing the actions here, instead of using the integer values, 123, could you use the more descriptive `HasStatistics` integer fields? So action 2 would become `HasStatistics.STATISTICS_ACTION_RESET`.

Maybe you could also write 2 more tests for action 3 and 4 (mark main and mark full)

core/src/test/java/nl/nn/adapterframework/util/MiscTest.java
```diff
@@ -1,43 +1,610 @@
...
+     */
+    @Test
+    public void testFileToStringFileName() throws Exception {
+        //Misc.resourceToString()
+        assertEquals(Misc.fileToString(file.getName()), "inside the lebron file");
+    }
+    @Test
+    public void testFileToStringFileNameEndLine() throws Exception {
+        //Misc.resourceToString()
+        assertEquals(Misc.fileToString(file.getName(), " the end"), "inside the lebron file");
```
nielsm5, line 259: You're not testing the `endOfLineString` string here?

nielsm5: I see you're testing it in another test, maybe to not confuse things leave it blank or `null` instead :)

core/src/test/java/nl/nn/adapterframework/util/MiscTest.java
```diff
@@ -1,43 +1,610 @@
...
+    /**
+     * Method: compress(String input)
+     */
+    @Test
+    public void testCompressDecompressInput() throws Exception {
+        String s = "test";
+        byte[] compressed = Misc.compress(s);
+        String decompressed = Misc.decompressToString(compressed);
+        assertEquals(decompressed, "test");
+    }
```
nielsm5, line 378: Maybe also test the compressed variant to validate if it has actually been compressed

core/src/test/java/nl/nn/adapterframework/util/MiscTest.java
```diff
@@ -1,43 +1,610 @@
...
+                "\n" +
+                "<body>\n" +
+                "<div>\n" +
+                "    <h1>Example Domain</h1>\n" +
+                "    <p>This domain is for use in illustrative examples in documents. You may use this\n" +
+                "    domain in literature without prior coordination or asking for permission.</p>\n" +
+                "    <p><a href=\"https://www.iana.org/domains/example\">More information...</a></p>\n" +
+                "</div>\n" +
+                "</body>\n" +
+                "</html>\n");
```
nielsm5, line 591: In stead of using a big concatinated string, it might be easier and more readable to use `TestFileUtils.getResource` and read the contents from a file instead

core/src/test/java/nl/nn/adapterframework/util/MiscTest.java
```diff
@@ -1,43 +1,610 @@
...
+    }
```
nielsm5, line 602: In stead of using a big concatinated string, it might be easier and more readable to use `TestFileUtils.getResource` and read the contents from a file instead

core/src/test/java/nl/nn/adapterframework/util/MiscTest.java
```diff
@@ -1,43 +1,610 @@
...
```
nielsm5, line 609: In stead of using a big concatinated string, it might be easier and more readable to use `TestFileUtils.getResource` and read the contents from a file instead

nielsm5:
Looking good, a few more bits and bobs and it's ready to go!

core/src/test/java/nl/nn/adapterframework/util/CounterStatisticTest.java
sinasen2000: will add.

sinasen2000: added

core/src/test/java/nl/nn/adapterframework/util/MiscTest.java
sinasen2000: I removed this test. It was not testing anything unique.

core/src/test/java/nl/nn/adapterframework/util/MiscTest.java
sinasen2000: I assumed you meant `TestFileUtils.getTestFile` which uses `getResource`. So I'm using this method.

core/src/test/java/nl/nn/adapterframework/util/MiscTest.java
sinasen2000: done

core/src/test/java/nl/nn/adapterframework/util/MiscTest.java
sinasen2000: done

core/src/test/java/nl/nn/adapterframework/util/MiscTest.java
sinasen2000: I didn't see this yesterday, will take a look at it now.

sinasen2000: I added a case that tests if the string is compressed. However, I did this by testing the first two indexes of the compressed byte array, which was the same for all compressed Strings, so I assumed this is a sign that it is compressed. Am I correct? Is there another way of testing if it is actually compressed? @nielsm5

PR #734, opened 2020-05-19
set error before call flushBuffer method

ccamiletti:


core/src/main/java/nl/nn/adapterframework/http/rest/ApiListenerServlet.java
```diff
@@ -500,8 +500,8 @@ else if(segment.startsWith("{") && segment.endsWith("}")) {
 		catch (Exception e) {
 			log.warn("ApiListenerServlet caught exception, will rethrow as ServletException", e);
 			try {
-				response.flushBuffer();
```
nielsm5, line 503: I think respose.reset(); should be called instead of flushBuffer, as you want to purge everything and throw the exeption (see log warning on line 501).

PR #771, opened 2020-05-29
Fix #591 Avoid use of plain XMLReaderFactory and TransformerFactory

### LINKED ISSUE #591: XML External Entity Injection
**Describe the bug**
It is possible to include external entities in an XML file.
The following XML document shows an example of an XXE attack.
```
<?xml version="1.0" encoding="ISO-8859-1"?>
<!DOCTYPE foo [
<!ELEMENT foo ANY >
<!ENTITY xxe SYSTEM "file:///dev/random" >]><foo>&xxe;</foo>
```
The property classloader.allowed.protocols exists to control which external entities may be accessed.
By using 
ClassLoaderEntityResolver, ClassLoaderURIResolver and ClassLoaderXmlEntityResolver as implementations of the EntityResolver we can control how resources should be retrieved. See ClassUtils.getResourceURL(ClassLoader, String, String);

Another method is by setting the following features on a TransformerFactory.
factory.setFeature("http://xml.org/sax/features/external-general-entities", false);
factory.setFeature("http://xml.org/sax/features/external-parameter-entities", false);

**Category**
 - Frank Framework

**To Reproduce**
Quick example shows placing 
`<!ENTITY xxe SYSTEM "file:///dev/random" >]><foo>&xxe;</foo>`
in Configuration.xml allows the use of external resources, why is this possible as this should use the ClassLoaderEntityResolver?

**Expected behavior**
Not load file:///dev/random into the JVM's memory and blowing up the server.

**Additional context**
- DomTreeAligner:167 -- TransformerFactory tf = TransformerFactory.newInstance(); uses a custom factory instead of the XmlUtils.identityTransform(Resource); This is used in more places.

- I'd like every XMLReader to use the ClassLoaderEntityResolver.
- Is the NonResolvingExternalEntityResolver still used? Has this become obsolete by using the ClassLoaderEntityResolver?

gvanbrakel:


core/src/test/java/nl/nn/adapterframework/xml/XmlWriterTest.java
```diff
@@ -62,8 +62,8 @@ public void testNoLexicalHandling() throws Exception {
 		XmlWriter xmlWriter = new XmlWriter();
 		
 		InputSource inputSource = new InputSource(new StringReader(input));
-		XMLReader xmlReader = XmlUtils.getXMLReader(true, true);
-		xmlReader.setContentHandler(xmlWriter);
+		XMLReader xmlReader = XmlUtils.getXMLReader(xmlWriter);
+		xmlReader.setProperty("http://xml.org/sax/properties/lexical-handler", null);
```
nielsm5, line 66: Maybe you could add a small description (comment) why you're explicitly turning off the lexical-handler

core/src/test/java/nl/nn/adapterframework/xml/XmlPrettyPrintFilterTest.java
```diff
@@ -47,8 +47,8 @@ public void testNoLexicalHandling() throws Exception {
 		filter.setContentHandler(xmlWriter);
 
 		InputSource inputSource = new InputSource(new StringReader(input));
-		XMLReader xmlReader = XmlUtils.getXMLReader(true, true);
-		xmlReader.setContentHandler(filter);
+		XMLReader xmlReader = XmlUtils.getXMLReader(filter);
+		xmlReader.setProperty("http://xml.org/sax/properties/lexical-handler", null);
```
nielsm5, line 51: Maybe you could add a small description (comment) why you're explicitly turning off the lexical-handler

core/src/test/java/nl/nn/adapterframework/util/XmlParserTest.java
```diff
@@ -0,0 +1,84 @@
...
+	@Ignore("requires proper setup on the classpath filesystem")
+	public void testParseXmlResourceWithExternalEntityInjection() throws IOException, SAXException, ParserConfigurationException {
+		Resource input  = Resource.getResource("/XmlUtils/EntityResolution/in-file-entity-c-temp.xml");
+		String expected = TestFileUtils.getTestFile("/XmlUtils/EntityResolution/out-resolved.xml");
+		
+		XmlWriter writer = new XmlWriter();
+		XmlUtils.parseXml(input, writer);
+		
+		MatchUtils.assertXmlEquals(expected,writer.toString());		
+	}
```
nielsm5, line 80: What do you mean by proper setup? Is it possible to get the absolute path (feed the URL to a file and call `toPath()`). 
I believe one of the 'resource validation checks' is if the resource starts with the file:// protocol.

core/src/main/java/nl/nn/adapterframework/util/XmlUtils.java
```diff
@@ -978,28 +985,19 @@ public static Source stringToSource(String xmlString) throws DomBuilderException
...
 	}	
 	
-	public static SAXSource inputSourceToSAXSource(InputSource is, boolean namespaceAware, boolean resolveExternalEntities) throws SAXException {
+	public static SAXSource inputSourceToSAXSource(InputSource is, Resource classloaderProvider) throws SAXException {
+		return inputSourceToSAXSource(is, true, classloaderProvider);
+	}
+	public static SAXSource inputSourceToSAXSource(InputSource is, boolean namespaceAware, Resource classloaderProvider) throws SAXException {
 		try {
-			return new SAXSource(getXMLReader(namespaceAware, resolveExternalEntities), is);
+			return new SAXSource(getXMLReader(namespaceAware, classloaderProvider), is);
```
nielsm5, line 1000: Ik vind dit raar. De inputsource moet uit de Resource komen!

PR #870, opened 2020-06-23
Add javadoc to Misc class

sinasen2000:
- Added javadoc to all the methods I have tested in the Misc class.
- Some system methods still lack javadocs.
- Also in the hide method, changed the variable name of hideRegex into regexForHiding as the former can be understood as a boolean to whether hide the regexes or not.

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/util/Misc.java
+++ b/core/src/main/java/nl/nn/adapterframework/util/Misc.java
@@ -490,7 +490,8 @@ public class Misc {
 	}
 
 	/**
-	 * Concatenates two strings, if specified, uses the separator in between two strings.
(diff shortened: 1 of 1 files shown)


core/src/main/java/nl/nn/adapterframework/util/Misc.java
```diff
@@ -324,8 +487,25 @@ public static String replace (String source, String from, String to) {
 		}
 		buffer.append (sourceArray, srcPos, sourceArray.length-srcPos);
 		return buffer.toString();
-	 }
+	}
 
+	/**
+	 * Concatenates two strings, if specified, uses the separator in between two strings.
```
gvanbrakel, line 493: You should explain that the separator is only used when part1 is non empty.

PR #871, opened 2020-06-23
Simplify IbisDebuggerAdvice handling of preserveInput and Senders

gvanbrakel:


test/src/main/configurations/MainConfig/Configuration.xml
```diff
@@ -64,6 +65,7 @@
 	&ExceptionPipe;
 	&SkipEmptyTags;
 	&ReplacerPipe;
+	&RelaceInputAndPreserve;
```
nielsm5, line 68: Re`p`lace

PR #886, opened 2020-06-30
Sort Schedulers by name and add new fields while configuring the job

alisihab:
- Sort schedulers by name
- Allow configuring the job group name
- Allow configuring description for a job

### DIFF
--- a/core/src/main/resources/IAF_WebControl/IAF_DatabaseChangelog.xml
+++ b/core/src/main/resources/IAF_WebControl/IAF_DatabaseChangelog.xml
@@ -128,5 +128,12 @@ TagDatabase
 		<!-- in Quartz the JobKey is defined by name and group, hence this being the primairy key constraint in the database as well -->
 		<addPrimaryKey tableName="IBISSCHEDULES" columnNames="JOBNAME, JOBGROUP" constraintName="PK_IBISSCHEDULES"/>
 	</changeSet>
+	
+	<changeSet id="IAF_Util:7" author="Ali Sihab Akcan">
+		<comment>Add description column to IBISSCHEDULES Table</comment>
+		<addColumn tableName="IBISSCHEDULES" >
+			<column afterColumn="MESSAGE" name="DESCRIPTION"  type="java.sql.Types.VARCHAR(255)"/> 
+		</addColumn>
+	</changeSet>
 
 </databaseChangeLog>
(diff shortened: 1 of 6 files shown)


core/src/main/resources/IAF_WebControl/IAF_DatabaseChangelog.xml
```diff
@@ -128,5 +128,12 @@ TagDatabase
 		<!-- in Quartz the JobKey is defined by name and group, hence this being the primairy key constraint in the database as well -->
 		<addPrimaryKey tableName="IBISSCHEDULES" columnNames="JOBNAME, JOBGROUP" constraintName="PK_IBISSCHEDULES"/>
 	</changeSet>
+	
+	<changeSet id="IAF_Util:7" author="Ali Sihab Akcan">
+		<comment>Add description column to IBISSCHEDULES Table</comment>
+		<addColumn tableName="IBISSCHEDULES" >
+			<column afterColumn="MESSAGE" name="DESCRIPTION"  type="java.sql.Types.VARCHAR(255)"/> 
+		</addColumn>
+	</changeSet>
```
nielsm5, line 137: I think it's wise to add a default (for example `null`) to the description field. If you want to add a column without default value to an already populated table you will get sql errors

example/src/test/resources/webcontrol/pipes/ShowConfigurationStatus_all.xml
```diff
@@ -8,7 +8,7 @@
   </configurations>
   <registeredAdapters all="true" log="false">
       <configurationMessages>
-         <configurationMessage date="0000-00-00 00:00:00.000" level="INFO">Configuration [IAF_WebControl] LiquiBase applied [6] change(s) and added tag [IAF_Util:6:Niels Meijer]</configurationMessage>
+         <configurationMessage date="0000-00-00 00:00:00.000" level="INFO">Configuration [IAF_WebControl] LiquiBase applied [7] change(s) and added tag [IAF_Util:7:Ali Sihab Akcan]</configurationMessage>
```
nielsm5, line 11: I'm assuming this is the only change in this (massive) file. If so, could you revert the other changes? The test performs an XML diff so it doesn't look at the order of the elements.

core/src/main/resources/IAF_WebControl/IAF_DatabaseChangelog.xml
alisihab: I think having defaultValue is required when we have a nullable=false constraint on the column for the existing rows to have a real value. With the above change, the column(nullable) will be added and the existing rows will have NULL value by default

PR #913, opened 2020-07-07
Allow AppConstants properties in Ant scripts

mhdirkse:


larva/src/main/java/nl/nn/adapterframework/testtool/FileSender.java
```diff
@@ -97,6 +99,16 @@ private void runAntScript() {
...
+		
+		// iterate over appConstants and add them as properties
+		AppConstants appConstants = AppConstants.getInstance();
+		@SuppressWarnings("unchecked")
+		Enumeration<String> enums = (Enumeration<String>) appConstants.propertyNames();
+		while (enums.hasMoreElements()) {
+			String key = enums.nextElement();
+			ant.setProperty(key, appConstants.getResolvedProperty(key));
+		}
+
```
gvanbrakel, line 111: Je voegt nu deze AppConstants-properties toe aan de FileSender. Ligt het niet meer voor de hand om ze op een hoger niveau toe te voegen, zodat je de AppConstants ook op andere plekken in Larva kunt gebruiken?

mhdirkse: Without this pull request, Larva allows you already to use AppConstants properties in scenario files, include files and input files. Niels has decided that properties should not be supported in output files; I have forgotten the reason. Only the option to use AppConstants properties in Ant scripts is missing. Adding this is the purpose of this pull request.

Laurens-makel: See this PR which enabled resolving properties in input files https://github.com/ibissource/iaf/pull/730

The reason Niels decided not to resolve properties in output files is that it will break the save result feature, since it will save the resolved property instead of the unresolved one.

Another reason that in output files it shouldn't be necessary, you can accomplish the same result with applying ignores to your scenario.

PR #919, opened 2020-07-08
Fix GUI page names

nielsm5:


nielsm5:
Add configuration flow diagram
image

PR #936, opened 2020-07-15
Add MariaDB connectivity

ricardovh:


ricardovh:
IAF-Test fails on multiple tests because of sequences not being created by LiquiBase. MariaDB itself does support sequences, so nextval.target.MariaDB is set, but as the sequence is not present, this makes the test fail.

core/src/main/java/nl/nn/adapterframework/jdbc/dbms/MariaDbDbmsSupport.java
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.commons.lang.StringUtils;
+
+import nl.nn.adapterframework.jdbc.JdbcException;
+import nl.nn.adapterframework.util.JdbcUtil;
+
+/**
+* Support for MySql/MariaDB.
+* 
+*/
+public class MariaDbDbmsSupport extends GenericDbmsSupport {
```
gvanbrakel, line 31: Would it be possible to let MariaDbDbmsSupport to extend MySqlDbmsSupport instead of GenericDbmsSupport?

PR #938, opened 2020-07-15
Use a shorter stubbed result for UUIDGeneratorPipe when type is numeric

ricardovh:


ricardovh:
Reason for this PR is that I have an old adapter that processes the result of the UUIDGeneratorPipe with type='numeric' based on that length (max database column length). The normal stubbed result matches the type='alphanumeric', which is longer than that, resulting in different behavior when stubbing. In older versions the UUIDGeneratorPipe was not stubbed, but now during upgrading it is stubbed.

PR #948, opened 2020-07-17
Add possibility to start Tomcat server from pom.xml

leeuw73:


jacodg:
Doesn't this have the same purpose as the Frank!Runner?

PR #961, opened 2020-07-21
FileSepAware: Make Base64Pipe tests file separator aware

mhdirkse:


mhdirkse:
This pull request is about the "test" subproject, sometimes called "Ibis4Test", which holds Larva tests that test the Frank!Framework. This pull fixes most test scenarios under Linux, but not all. Only issues with the path separator are fixed.

All scenarios still succeed under Windows, but under Linux the following still fail:
- FxF3 scenario 01
- FxF3 scenario 02
- MessageStoreSenderAndListener scenario 01
- MessageStoreSenderAndListener scenario 02

mhdirkse:
Hello Niels and Gerrit,

I followed your comment that these SEP words should be replaced by file paths. I tested my work again under Windows and under Linux. Under Windows, all tests succeed. Under Linux, only four tests fail like before. Now that I write this, I realize I did not check that it are exactly the same four tests as before. I will redo the test to check this.

Regards,

Martijn Dirkse

mhdirkse:
Yes, the four scenarios that fail under Linux are the same.

mhdirkse:
The Larva tests were already fixed under Linux. No longer relevant.

PR #981, opened 2020-08-06
Filename comparator test

sinasen2000:


core/src/test/java/nl/nn/adapterframework/util/FileNameComparatorTest.java
```diff
@@ -0,0 +1,94 @@
...
+		File f3 = new File("first.txt");
+		File f4 = new File("FiRst.txt");
+		int i = FileNameComparator.compareFilenames(f1, f2);
+		int j = FileNameComparator.compareFilenames(f1, f3);
+		int k = FileNameComparator.compareFilenames(f1, f4);
+		int l = FileNameComparator.compareFilenames(f2, f1);
+		assertEquals(-13, i);
+		assertEquals(0, j);
+		assertEquals(32, k);
+		assertEquals(13, l);
```
nielsm5, line 91: Instead of giving exact values, perhaps you can use
` ... == 0`, `... < 0` and `... > 0` instead.

core/src/test/java/nl/nn/adapterframework/util/FileNameComparatorTest.java
```diff
@@ -0,0 +1,94 @@
...
+		String s4 = "File2.txt";
+		int a = FileNameComparator.compareStringsNaturalOrder(s1, s2, true);
+		int b = FileNameComparator.compareStringsNaturalOrder(s1, s3, true);
+		int c = FileNameComparator.compareStringsNaturalOrder(s2, s3, true);
+		int d = FileNameComparator.compareStringsNaturalOrder(s2, s4, true);
+		assertEquals(32, a);
+		assertEquals(0, b);
+		assertEquals(-32, c);
+		assertEquals(0, d);
+	}
```
nielsm5, line 73: This test looks a lot like the test underneath, perhaps here you can test the 3rd argument of the `compareStringsNaturalOrder` method instead (case sensitivity)

core/src/test/java/nl/nn/adapterframework/util/FileNameComparatorTest.java
```diff
@@ -0,0 +1,94 @@
...
+	/**
+	 * Method: getNextIndex(String s, int start, boolean numericPart)
+	 */
+	@Test
+	public void testGetNextIndex() throws Exception {
+		int i = FileNameComparator.getNextIndex("test.txt", 0, true);
+		int j = FileNameComparator.getNextIndex("test.txt", 0, false);
+		assertEquals(0, i);
+		assertEquals(8, j);
+	}
```
nielsm5, line 44: This is an edge case. In a filesystem it's not possible to have 2 files with the same name. Use test.txt twice theoretically cannot happen. Also don't test for an exact integer, but rather something equal, higher or lower then `0`.

nielsm5 (changes requested):

core/src/test/java/nl/nn/adapterframework/util/FileNameComparatorTest.java
```diff
@@ -0,0 +1,77 @@
...
+		File f3 = new File("first.txt");
+		File f4 = new File("FiRst.txt");
+		int i = FileNameComparator.compareFilenames(f1, f2);
+		int j = FileNameComparator.compareFilenames(f1, f3);
+		int k = FileNameComparator.compareFilenames(f1, f4);
+		int l = FileNameComparator.compareFilenames(f2, f1);
+		assertTrue(i < 0);
+		assertTrue(j == 0);
+		assertTrue(k > 0);
+		assertTrue(l > 0);
```
gvanbrakel, line 74: Please put each of these test on a single line, like:
```
assertTrue(FileNameComparator.compareFilenames(f1, f2) < 0);
```

core/src/test/java/nl/nn/adapterframework/util/FileNameComparatorTest.java
```diff
@@ -0,0 +1,77 @@
...
+	}
+
```
nielsm5, line 76: Can you also add test for file names that end with a number? `ibis_xml.log.1`  `ibis_xml.log.2` etc
those should also be sorted properly.

Also the `ibis-stats_2020W31.log` files?

gvanbrakel: And also compare filename4.txt to filename13.txt

core/src/test/java/nl/nn/adapterframework/util/FileNameComparatorTest.java
sinasen2000: thank you, changing right now!

core/src/test/java/nl/nn/adapterframework/util/FileNameComparatorTest.java
sinasen2000: Added both, adding some more interesting cases as well.

gvanbrakel (approved):

nielsm5 (approved):

PR #982, opened 2020-08-07
Add text to the Javadoc that gives an overview of how the FF! works

mhdirkse:


src/main/javadoc/overview.html
```diff
@@ -0,0 +1,131 @@
...
+source of user documentation is the Frank!Doc, which you can download from the
+Frank!Framework's web console.</a>.
+<p>
+The Frank!Framework is a Java application that provides a web interface.
+It runs on an application server. WeAreFrank! supports running the
+Frank!Framework on Apache Tomcat, JBoss Application Server (recently renamed to WildFly)
+or WebSphere Application Server. The web interface is typically
+used by operators who are responsible for the proper operation
+of the integration solution, or as WeAreFrank! says: the Frank.
+The Frank!Console is described in the <a href="http://frank-manual.readthedocs.io">Frank!Manual</a>.
```
gvanbrakel, line 45: I would suggest to deep link here to the Frank!Console chapter of the manual directly

mhdirkse:
Thanks, Gerrit. I agree and I will change this. The text being created here is not complete yet. I need more information from you or Niels on how the Frank!Framework works. I will ask one of you to be able to continue.

src/main/javadoc/overview.html
mhdirkse: I left out the Frank!Console because I did not know how to introduce that term. I already tell that the Frank!Framework is a web application, so I thought it confusing to name the web interface "Frank!Console". I have links to the Frank!Manual and to ibis4example to show the reader what the Frank!Framework does.

jjansenvr:
@mhdirkse today we decided we do not want this text to be part of the javadoc documentation, because this will mean we will have to maintain documentation in yet another place. 

We would like to keep the first paragraph and make a reference to the [github](https://github.com/ibissource/iaf) 

For google ranking it is a good idea to add the readthedocs and other dcumentation url's

mhdirkse:
I wanted to add a link to GitHub here, but that does not work. I did a Maven build to create the JavaDoc and opened the produced index.html in Firefox. I could see the text I typed in the file `overview.html`. There is a link with text "GitHub" as I intended, but clicking it did not work. The JavaDoc webpage tried to open GitHub within the JavaDoc site, but GitHub refused the connection. See picture

image

mhdirkse:
I propose that we do not merge this PR, also not the "first paragraph" that is left now. We cannot add the links we want. Furthermore, I think it is not likely that someone who does not know our F!F will get the JavaDocs first. Potential customers will find our GitHub repository or the Frank!Manual first. For technical people, we can publish the JavaDocs on GitHub. See https://stackoverflow.com/questions/15347808/display-javadocs-on-github/15364790. I have seen that you can create releases on GitHub, like this:

image

If we make releases like shown, we can also add JavaDocs there.

jacodg:
Maybe we can just add the links as text? So somebody could copy/paste it into the address bar of the browser?

jjansenvr:
https://stackoverflow.com/questions/1082050/linking-to-an-external-url-in-javadoc

```
/**
 * @see <a href="http://google.com">http://google.com</a>
 */
```
will show as a link in the javadoc

nielsm5:
> I propose that we do not merge this PR, also not the "first paragraph" that is left now. We cannot add the links we want. Furthermore, I think it is not likely that someone who does not know our F!F will get the JavaDocs first. Potential customers will find our GitHub repository or the Frank!Manual first. For technical people, we can publish the JavaDocs on GitHub. See https://stackoverflow.com/questions/15347808/display-javadocs-on-github/15364790. I have seen that you can create releases on GitHub, like this:
> 
> image
> 
> If we make releases like shown, we can also add JavaDocs there.

We already publish releases as shown. I don't understand what the benefits are of adding the javadoc artifact over publishing it to https://javadoc.frankframework.org

mhdirkse:
@jjansenvr, your suggestion does not work for GitHub. I updated the link such that the href attribute and the label of the link are the same, but I still get the same unwanted result. The ReadTheDocs link works fine, even though the href attribute and the label are different. I guess GitHub has some security policy that prevents it from being shown within another webpage.

I also tried to use http instead of https (not committed and not pushed), but that does not work neither.

jjansenvr:
lets close this PR and move on

mhdirkse:
I close it now.

nielsm5:
Use the attribute `target="_blank"` within the `a` element to open the page in a new window!

gvanbrakel (approved):

nielsm5 (approved):

mhdirkse:
@nielsm5, I have tested your change of this PR. It works now, thanks.

PR #996, opened 2020-08-19
File utils test

sinasen2000:


core/src/test/java/nl/nn/adapterframework/util/FileUtilsTest.java
```diff
@@ -0,0 +1,283 @@
...
+	@ClassRule
+	public static TemporaryFolder testFolderDest = new TemporaryFolder();
+
+	public static String sourceFolderPath;
+
+	public static String destFolderPath;
+
+	private static File f1;
+
+	public String sep = File.separator;
```
nielsm5, line 35: Since this is a field that doesn't change, I recommend you turning it in to a `public static final String` and with final fields it's a java standard to use all caps as field name, so for example `public static final String FILE_SEPARATOR = File.separator` Not sure if it will be any shorter then `File.separator` which I believe was the idea here but just my 2c.

core/src/test/java/nl/nn/adapterframework/util/FileUtilsTest.java
```diff
@@ -0,0 +1,283 @@
...
+	public void testGetFreeFile() throws Exception {
+		File f = new File(".." + sep + "core" + sep + "src" + sep + "test" + sep + "resources" + sep + "Pipes");
+		File res = FileUtils.getFreeFile(f);
+		assertEquals("Pipes", res.getName());
+	}
+
+
+	@Test
+	public void testMoveFile() throws Exception {
+		File toBeMoved = new File(".." + sep + "core" + sep + "src" + sep + "test" + sep + "resources" + sep + "Pipes" + sep + "movingFile.txt");
```
nielsm5, line 55: Please use an URL and turn that into a file here. Renaming the path or whatever would otherwise needlessly break the test.
`this.getClass().getResource("/Pipes/movingFile.txt");` will return an URL (to which you could apply a null check (`assertNotNull(xxxx)`) which you can then feed to the new File constructor :)

core/src/test/java/nl/nn/adapterframework/util/FileUtilsTest.java
```diff
@@ -0,0 +1,283 @@
...
+		String[] l = f.list();
+		assertEquals(l[0], res.getName());
+	}
+	/**
+	 * Method: encodeFileName(String fileName)
+	 */
+	@Test
+	public void testEncodeFileName() throws Exception {
+		String s = " abc.txt";
+		String encoded = FileUtils.encodeFileName(s);
```
nielsm5, line 264: Try to add some funky characters that require to be encoded, such as `/`. `:`, `?` to spice things up a bit.

core/src/test/java/nl/nn/adapterframework/util/FileUtilsTest.java
nielsm5: I just realized you do this more often, maybe you could create a method that does all this and returns a `File`?

nielsm5 (changes requested):

core/src/test/java/nl/nn/adapterframework/util/FileUtilsTest.java
sinasen2000: yes, I will create a method for this and replace all.

core/src/test/resources/Util/FileUtils/copyFile.txt
```diff
@@ -0,0 +1 @@
+will be copied to copyFIle.txt and appended to destinationFile.txtwill be copied to copyFIle.txt and appended to destinationFile.txtwill be copied to copyFIle.txt and appended to destinationFile.txtwill be copied to copyFIle.txt and appended to destinationFile.txtwill be copied to copyFIle.txt and appended to destinationFile.txtwill be copied to copyFIle.txt and appended to destinationFile.txtwill be copied to copyFIle.txt and appended to destinationFile.txtwill be copied to copyFIle.txt and appended to destinationFile.txtwill be copied to copyFIle.txt and appended to destinationFile.txtwill be copied to copyFIle.txt and appended to destinationFile.txt
```
gvanbrakel, line 1: Ik vind het een beetje stom om zo'n bij elkaar gebaggerde file te hebben. De inhoud kan ook enkelvoudig zijn en begrijpelijker

gvanbrakel (changes requested):

gvanbrakel (approved):

nielsm5 (approved):

Issue #1005, opened 2020-08-24
Feature request: Ignore larva test step response


ricardovh:
**Reporter**
Ricardo

**Is your feature request related to a problem? Please describe.**
As part of a larva test, I regularly have to cleanup tables in a database. To ensure I actually work on the state that I created as initialization, I need to do a .read to wait for the response. This requires me to have a file to compare that result with. As it is an initialization step, the response might change between runs, which would require several ignores. However, for later steps I am interested in the values returned, so ignores cannot be used. Currently the only solution is to use a DelayPipe and wait long enough for the transaction to finish.

**Describe the solution you'd like**
An .ignoreRead, which makes the adapter read the response, but do not do any compares with it.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

gvanbrakel:
I don't understand why ignores cannot be used (though I understand that it ), and I also do not understand how the values could be used in later steps.
Could you explain?

ricardovh:
I have a test in which I empty and fill various tables in the database as part of initialization. I need to be sure that that step has finished, so I need to do a read before doing a write on the actual adapter I want to test. As the content in the database might actually change between tests, I would need to ignore the <rowsupdated>X</rowsupdated>. The actual adapter ends with a FixedQuerySender, giving me the same rowsupdated in the response, where I do want to know how many rows have been updated. Then I empty the tables again, and there I also would want to know how much rows have been deleted. Of course I do verify the contents of the tables as part of the test, so with some set of ignores I could ignore the rowsupdated and still verify that my adapter worked as intended, but there might be other usecases where an .ignoreRead might be useful.

gvanbrakel:
OK. I think I understand now.
You want an ignore, but only for this specific step. If you would construct an ignore of content between keys <result> and </result>, they would be ignored everywhere.

ricardovh:
Yes, you understand correctly.

PR #1010, opened 2020-08-25
Add option to ignore .read actions in Larva

Laurens-makel:
Implements feature request #1005


larva/src/main/java/nl/nn/adapterframework/testtool/TestTool.java
```diff
@@ -3072,12 +3072,16 @@ private static String getEncoding(String fileName, String content) {
 	}
 
 	public static int compareResult(String step, String stepDisplayName, String fileName, String expectedResult, String actualResult, Properties properties, Map<String, Object> writers, String queueName) {
+		if(properties.getProperty(queueName + ".ignoreRead") != null && properties.getProperty(queueName + ".ignoreRead").equalsIgnoreCase("true") ){
```
gvanbrakel, line 3075: As I understand, the idea of this solution is that configuration will be something like:
```
step1.manage.db.write = scenario01/md-delete.xml
step2.manage.db.read = scenario01/md-delete-result.xml
step2.manage.db.ignoreRead = true
step3.java....
```
Although this provides the requested functionality, it is a bit clumsy: it requires a 'expected file' that is never actually compared to the 'actual file'. I think this is confusing.

I would like to propose to 
- rename the key 'ignoreRead' into 'differences', that can have the folloing values:
  - fail: the default, as it is now: show the differences, and fail the test if they exist
  - ignore: show the differences, allow to update the expected, but do not fail the test
  - ignore-silent: do not even show the differences
- add a key 'readIgnore' that can be set to true, to will act as a combination of 'read' and 'differences=ignore-silent'

gvanbrakel (changes requested):

gvanbrakel:
superseded by #1059

PR #1085, opened 2020-09-23
Subproject test should produce a WAR that accesses the database on arbitrary IP address

mhdirkse:
I tested this by checking that I can still run the tests in Eclipse. I also tested a command-line Maven build in which I adjust the database IP address.

I did not test this end-to-end, but I checked that context.xml was updated as I wanted. I also compared with a .war file that was created without this fix. I extracted the two .war files and did a WinDiff.

One thing struck me about the WinDiff output. There were a few .jar files that were different. These had nothing to do with this change. I think this is not relevant, but I hope that someone can confirm this. I can show the WinDiff result in a chat if necessary.


test/pom.xml
```diff
@@ -14,6 +14,7 @@
 	<properties>
 		<m2eclipse.wtp.contextRoot>iaf-test</m2eclipse.wtp.contextRoot>
 		<failOnMissingWebXml>false</failOnMissingWebXml>
+		<databaseIpAddress>localhost</databaseIpAddress>
```
gvanbrakel, line 17: Ik zou dit niet ``databaesIpAddress`` noemen, maar ``databaseHost``

gvanbrakel (changes requested):

test/pom.xml
mhdirkse: Gedaan.

gvanbrakel (approved):

PR #1091, opened 2020-09-24
Use folder relative to tempdir and not the root of the project

nielsm5:
Removes `hoooray.zip` file from project root folder


gvanbrakel (changes requested):
Zou je nog pipe.configure() en pipe.open() kunnen toevoegen aan elke test?

gvanbrakel (approved):

PR #1129, opened 2020-10-07
Fix #1026 Fix exception overriding in webservice sender

### LINKED ISSUE #1026: WebServiceSender-multipart response without file
**Describe the bug**
In some cases a response from the WebServiceListener to a WebServiceSender is required to be in multipart format, even when there are no files/parts to be returned. It appears that in either the WebServiceListener or the WebServiceSender this work not properly for the case that there is no file.

**Reporter**
Ali

**Category** 
 - Frank Console
 - Frank Framework
* - Ladybug (debugging tool) --> also is a separate project with own issues*
 - Larva (unit test tool)
 - Performance

**To Reproduce**

**Expected behavior**

**Screenshots**

**Environment:**
 - Application server (f.e. Tomcat/WebSphere/JBOSS/Jetty)
 - IAF Version [e.g. 7.5 RC3, 7.6]

**Additional context**

alisihab:
- multipartResponse attribute will be deprecated. Responses with a 'multipart/*' content-type will be detected automatically.
Fixes #1026

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/http/HttpSender.java
+++ b/core/src/main/java/nl/nn/adapterframework/http/HttpSender.java
@@ -569,6 +569,10 @@ public class HttpSender extends HttpSenderBase {
 		}
 	}
 
+	/**
+	 * When an exception occurs and the response cannot be parsed, we do not want to throw a 'missing response' exception. 
+	 * Since this method is used when handling other exceptions thrown by {@link WebServiceSender} and {@link HttpSender} , silently return null, to avoid npe's, ioexceptions etc.
+	 */
 	public String getResponseBodyAsString(HttpResponseHandler responseHandler, boolean throwIOExceptionWhenParsingResponse) throws IOException {
 		String charset = responseHandler.getCharset();
 		log.debug(getLogPrefix()+"response body uses charset ["+charset+"]");
--- a/core/src/main/java/nl/nn/adapterframework/http/HttpSender.java
+++ b/core/src/main/java/nl/nn/adapterframework/http/HttpSender.java
@@ -36,6 +36,7 @@ import javax.mail.internet.MimeMultipart;
 import javax.servlet.http.HttpServletResponse;
 
 import org.apache.commons.codec.binary.Base64InputStream;
+import org.apache.commons.lang.BooleanUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.http.Header;
 import org.apache.http.HttpEntity;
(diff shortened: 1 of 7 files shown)


core/src/test/java/nl/nn/adapterframework/http/WebServiceSenderResultTest.java
```diff
@@ -0,0 +1,202 @@
...
+		this.sender = sender;
+		return sender;
+	}
+
+	public WebServiceSender createWebServiceSenderFromFile(String testFile, int statusCode) throws IOException {
+		InputStream file = getFile(testFile);
+		byte[] fileArray = Misc.streamToBytes(file);
+		String contentType = null;
+
+		BufferedReader reader = new BufferedReader(new InputStreamReader(new ByteArrayInputStream(fileArray)));
```
nielsm5, line 94: What is happening here? an inputstream to bytearray to inputstream to reader?

core/src/test/java/nl/nn/adapterframework/http/WebServiceSenderResultTest.java
```diff
@@ -0,0 +1,202 @@
...
+import nl.nn.adapterframework.util.Misc;
+
+public class WebServiceSenderResultTest extends Mockito {
+
+	@Rule
+	public ExpectedException thrown = ExpectedException.none();
+
+	private WebServiceSender sender = null;
+
+	public WebServiceSender createWebSeriveSender(InputStream responseStream, String contentType, int statusCode) throws IOException {
```
nielsm5, line 56: createWebSeriveSender should be createWebServiceSender

nielsm5 (changes requested):

core/src/main/java/nl/nn/adapterframework/http/HttpSender.java
```diff
@@ -577,7 +583,16 @@ public String getResponseBodyAsString(HttpResponseHandler responseHandler) throw
...
 		}
-		String responseBody = responseHandler.getResponseAsString(true);
+		String responseBody = null;
+		try {
+			responseBody = responseHandler.getResponseAsString(true);
+		} catch(IOException e) {
+			if(throwIOExceptionWhenParsingResponse) {
+				throw e;
+			}
+			return null;
```
gvanbrakel, line 593: And what happens to the exception then? It's just gone?
At least log it!

gvanbrakel (changes requested):

core/src/main/java/nl/nn/adapterframework/http/HttpSender.java
alisihab: This method will be called in case of an exception. The exception thrown by this method is not the actual cause. Therefore, it can be ignored. The caller methods will throw the actual exception

nielsm5: I think it would be wise to add a comment describing the problem. Either near the try/catch or as javadoc above the method.

When an exception occurs and the response cannot be parsed, we do not want to throw a 'missing response' exception. Since this method is used when handling other exceptions (link to wss exception here) (link to http exception here), silently return null, to avoid npe's, ioexceptions etc.

nielsm5 (approved):

gvanbrakel (approved):

PR #1131, opened 2020-10-08
Fix type mismatch between return datatype of getter and argumet datatype of setter

### LINKED ISSUE #1056: ApiListener authentication attributes give warnings in console
**Describe the bug**
According to the IbisDoc, I should be able to set authenticationMethod and authenticationRoles on an ApiListener. Setting these attributes results in warnings in the console. I did not test if the values are actually honored or if they are also not working, as locally I do not enable security constraints.

**Reporter**
Ricardo

**Category** 
 - Frank Console
 - Frank Framework

**To Reproduce**
1. Start an adapter with the below ApiListener
2. Open console
3. See warnings

**Expected behavior**
No warnings

**Screenshots**
N/A

**Environment:**
 - Application server (f.e. Tomcat/WebSphere/JBOSS/Jetty): Tomcat
 - Browser [e.g. chrome, safari]: Chrome
 - IAF Version [e.g. 7.5 RC3, 7.6]: 7.5-RC4, 7.6-20200909.154756

**Additional context**
`<listener name="Test" 
				className="nl.nn.adapterframework.http.rest.ApiListener"
				consumes="JSON"
				produces="JSON"
				method="POST"
				uriPattern="test"
				authenticationMethod="AUTHROLE"
				authenticationRoles="IbisObserver" 
				messageIdHeader="x-messageid"/>
`

alisihab:
Fix datatype for authentication attributes to let property utils to find appropriate write method
Fixes #1056

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/http/rest/ApiListener.java
+++ b/core/src/main/java/nl/nn/adapterframework/http/rest/ApiListener.java
@@ -230,7 +230,7 @@ public class ApiListener extends PushingListenerAdapter<String> implements HasPh
 		}
 	}
 
-	public AuthenticationMethods getAuthenticationMethod() {
+	public AuthenticationMethods getAuthenticationMethodEnum() {
 		if(authenticationMethod == null) {
 			authenticationMethod = AuthenticationMethods.NONE;
 		}
@@ -252,7 +252,7 @@ public class ApiListener extends PushingListenerAdapter<String> implements HasPh
 
 		this.authenticationRoles = roles;
 	}
-	public List<String> getAuthenticationRoles() {
+	public List<String> getAuthenticationRoleList() {
 		return authenticationRoles;
 	}
 
--- a/core/src/main/java/nl/nn/adapterframework/http/rest/ApiListener.java
+++ b/core/src/main/java/nl/nn/adapterframework/http/rest/ApiListener.java
@@ -45,8 +45,8 @@ public class ApiListener extends PushingListenerAdapter<String> implements HasPh
 	private String method;
 	private List<String> methods = Arrays.asList("GET", "PUT", "POST", "DELETE");
 
-	private String authenticationMethod = null;
-	private String authenticationRoles = null;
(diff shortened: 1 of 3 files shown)


core/src/main/java/nl/nn/adapterframework/http/rest/ApiListener.java
```diff
@@ -221,39 +229,20 @@ public boolean getUpdateEtag() {
...
+	public void setAuthenticationMethod(String authenticationMethod) {
+		this.authenticationMethod = authenticationMethod;
 	}
 
-	public AuthenticationMethods getAuthenticationMethod() {
+	public AuthenticationMethods getAuthenticationMethodType() {
 		if(authenticationMethod == null) {
-			authenticationMethod = AuthenticationMethods.NONE;
+			return AuthenticationMethods.NONE;
 		}
```
nielsm5, line 239: All you had to do was rename the method, the logic could have stayed where it was.

The idea of enums is that once (in the near future?) we can add them to our xsd so people can use autocomplete to fill the attribute values.

There isn't really a standard yet, but I think we should use `getMethodNameEnum()` so in this case `getAuthenticationMethodEnum`. Since there is no need to have a getter, and we don't want the 'BeansParser' to error when the dataTypes are different, just renaming the method should be enough. 

If a getter is desirable, the getter-method should call the `getMethodNameEnum().name()` method, which in turn returns the correct dataType.

core/src/test/java/nl/nn/adapterframework/http/rest/ApiListenerTest.java
```diff
@@ -191,19 +196,17 @@ public void acceptsJson() throws ConfigurationException {
 		assertTrue("accepts JSON", listener.accepts(acceptHeader));
 	}
 
-	@Test(expected = ConfigurationException.class)
+	@Test
```
nielsm5, line 199: I'm confused, shouldn't this test return an exception?

nielsm5 (changes requested):

nielsm5 (approved):

PR #1134, opened 2020-10-08
Upgrade commons digester

nielsm5:
I bumped into many issues while trying to parse the old XML digester rules file. It turns out that since version 2 most classes have become either package private or final so it's near impossible to extend. The DTD supplied by the dependency is also out of date creating inconsistent xml parsing/validating problems. Many bugs have been reported but since 2013 there has been no development activity. See: [DigesterRuleParser](https://issues.apache.org/jira/browse/DIGESTER-168), [properties-rule](https://user.commons.apache.narkive.com/QeKWeYQo/digester-xml-rules-set-properties-rule-and-ignoring-missing-properties),  [and dtd inconsistencies](https://stackoverflow.com/questions/11169761/when-using-commons-digester-xml-rules-how-do-i-exclude-some-attributes-of-the-x).

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/configuration/digester/FrankDigesterRules.java
+++ b/core/src/main/java/nl/nn/adapterframework/configuration/digester/FrankDigesterRules.java
@@ -15,6 +15,16 @@
 */
 package nl.nn.adapterframework.configuration.digester;
 
+import java.io.IOException;
+
+import javax.xml.XMLConstants;
+import javax.xml.bind.JAXBContext;
+import javax.xml.bind.JAXBException;
+import javax.xml.bind.Unmarshaller;
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.parsers.ParserConfigurationException;
+
 import org.apache.commons.digester3.Digester;
 import org.apache.commons.digester3.ObjectCreationFactory;
 import org.apache.commons.digester3.Rule;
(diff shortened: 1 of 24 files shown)


core/src/main/resources/springCommon.xml
```diff
@@ -278,7 +278,7 @@
 	Beans created from Configuration.xml, but prototyped here for
 	extra dependencies which cannot be created otherwise.
 	Bean names are prefixed with 'proto-' so that autowiring by
-	name doesn't create unnessecary instances.
+	name doesn't create unnecessarily instances.
```
gvanbrakel, line 281: doesn't create instances unnecessarily

core/src/main/java/nl/nn/adapterframework/configuration/digester/FrankDigesterRules.java
```diff
@@ -0,0 +1,139 @@
...
+	private GenericFactory factory = new GenericFactory();
+	private Rule attributeChecker = new AttributeCheckingRule();
+	private Digester digester;
+
+	public FrankDigesterRules(Digester digester) {
+		this.digester = digester;
+	}
+
+	@Override
+	public void configure(RulesBinder rulesBinder) {
```
gvanbrakel, line 46: Ik zie hier nog niet de structuur van digester-rules.xml in terug, dus per pattern een 'create', 'set-properties' en 'register'

gvanbrakel (approved):

PR #1190, opened 2020-10-22
Introduce responseRoot attribute in exits for openapi schema generation

alisihab:


core/src/main/java/nl/nn/adapterframework/core/PipeLineExit.java
```diff
@@ -72,30 +73,35 @@ public String getName() {
...
 
 	@IbisDoc({"when using restlistener and set to <code>true</code>, this removes the output and shows a blank page, the output is still logged in the ladybug testtool", "false"})
 	public void setEmpty(String b) {
 		emptyResult = Boolean.parseBoolean(b);
 	}
+	public boolean getEmptyResult() {
+		return emptyResult;
+	}
+
+	@IbisDoc({"Name of the element that the exit expects. In case the responseRoot attribute is set in the validator, for the exit code 200 it will use the first element in the responseRoot and for the other codes last element of the responseRoot will be used. (Remark: Only for OpenApi schema generation. The output will not be validated by looking at this attribute.)", ""})
```
gvanbrakel, line 100: I think you should rephrase this a little. Instead of 
``In case the responseRoot attribute is set in the validator`` you can write
``In case elementName is not set, but the responseRoot attribute is set in the validator``

core/src/main/java/nl/nn/adapterframework/http/rest/ApiServiceDispatcher.java
```diff
@@ -179,7 +179,7 @@ protected JsonObject generateOpenApiJsonSchema(Collection<ApiDispatchConfig> cli
 					if (StringUtils.isNotEmpty(adapter.getDescription())) {
 						methodBuilder.add("summary", adapter.getDescription());
 					}
-					methodBuilder.add("operationId", adapter.getName());
+					methodBuilder.add("operationId", listener.getName());
```
gvanbrakel, line 182: I do agree that the operationId should be based on receiver or listener, and not on adapter, but the name attribute as it is right now is not a nice solution, because it tends to get a name like ``listener of [Json2XmlValidator-API-POST]``.
Will this be solved when you extend this exit.elementName solution to the input? Maybe introduce a 'operationId' attribute in ApiListener then.

gvanbrakel (changes requested):

gvanbrakel:
Please run iaf-test and make sure all scenarios pass. Currently three scenarios in /Validators fail on schema generation

core/src/test/java/nl/nn/adapterframework/http/rest/OpenApiTest.java
```diff
@@ -21,8 +22,10 @@
 	public void simpleEndpointTest() throws Exception {
 		ApiServiceDispatcher dispatcher = ApiServiceDispatcher.getInstance();
 		assertEquals("there are still registered patterns! Threading issue?", 0, dispatcher.getPatternClients().size());
-
-		new AdapterBuilder("myAdapterName", "description4simple-get").setListener("users", "get").setValidator("simple.xsd", null, "user").build(true);
+		Parameter param = new Parameter();
+		param.setName("parameter");
+		param.setValue("parameter");
+		new AdapterBuilder("myAdapterName", "description4simple-get").setListener("users", "get", "operationId4simple-get").setValidator("simple.xsd", null, "user", param).build(true);
```
nielsm5, line 29: Please create a new test for every new thing you add.. Right now this single method is testing over 10 different things..

nielsm5 (changes requested):

core/src/main/java/nl/nn/adapterframework/http/rest/ApiListener.java
```diff
@@ -135,7 +141,10 @@ public String getCleanPattern() {
 		if(pattern.endsWith("/"))
 			pattern = pattern.substring(0, pattern.length()-1);
 
-		return pattern.replaceAll("\\{.*?}", "*");
+		if(replacePathParams)
+			pattern = pattern.replaceAll("\\{.*?}", "*");
+
```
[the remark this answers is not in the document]
alisihab: It will stay as it is regex corresponds to {*}

core/src/main/java/nl/nn/adapterframework/http/rest/ApiServiceDispatcher.java
```diff
@@ -171,27 +175,33 @@ protected JsonObject generateOpenApiJsonSchema(Collection<ApiDispatchConfig> cli
...
 					//ContentType may have more parameters such as charset and formdata-boundry
 					MediaTypes produces = listener.getProducesEnum();
 					methodBuilder.add("responses", mapResponses(adapter, produces, schemas));
 				}
-
 				methods.add(method.toLowerCase(), methodBuilder);
 			}
-			paths.add("/"+config.getUriPattern(), methods);
+			if(listener != null) {
+				paths.add("/"+listener.getCleanPattern(false), methods);
```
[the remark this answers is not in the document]
alisihab: The idea was "not to mask" specified path parameter. For the configured uripatterns with *, i think it is a good idea to leave them as they are

core/src/main/java/nl/nn/adapterframework/core/PipeLineExit.java
```diff
@@ -50,13 +50,14 @@
...
-	
+
 	public static final String EXIT_STATE_SUCCESS = "success";
-	
+
 	private String path;
 	private String state;
 	private int exitCode = 0;
 	private boolean emptyResult = false;
+	private String elementName;
```
gvanbrakel, line 60: We suggest to call this 'responseRoot'

core/src/main/java/nl/nn/adapterframework/core/PipeLineExit.java
```diff
@@ -72,30 +73,35 @@ public String getName() {
...
 
 	@IbisDoc({"when using restlistener and set to <code>true</code>, this removes the output and shows a blank page, the output is still logged in the ladybug testtool", "false"})
 	public void setEmpty(String b) {
 		emptyResult = Boolean.parseBoolean(b);
 	}
+	public boolean getEmptyResult() {
+		return emptyResult;
+	}
+	// TODO: validate the output by looking at the elementName
+	@IbisDoc({"configures exit specific element reference in open api schema. If it is not set, but the responseRoot attribute is set in the validator, for the exit code 200 it will use the first element in the responseRoot and for the other codes last element of the responseRoot will be used.", ""})
```
gvanbrakel, line 100: The choice of the 'first' or 'last' responseRoot value configured in the validator should be based on exit.state, instead of exit.code.

Please change the IbisDoc to:
Configures the responseRoot in the OpenAPI schema for this exit. If not set, the responseRoot value of the validator will be used. If that contains multiple (comma separated) values, the first will be used for the exits with state 'success', the last for the other exits.

core/src/main/java/nl/nn/adapterframework/http/rest/ApiServiceDispatcher.java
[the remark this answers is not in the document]
gvanbrakel: Please use getUriPattern() where you want to use getCleanPattern(false). Then it is not necessaray to add a parameter to getCleanPattern().

gvanbrakel (changes requested):

core/src/main/java/nl/nn/adapterframework/http/rest/ApiListener.java
```diff
@@ -135,7 +136,7 @@ public String getCleanPattern() {
 		if(pattern.endsWith("/"))
 			pattern = pattern.substring(0, pattern.length()-1);
 
-		return pattern.replaceAll("\\{.*?}", "*");
+		return pattern = pattern.replaceAll("\\{.*?}", "*");
```
gvanbrakel, line 139: No need to assign pattern here

core/src/main/java/nl/nn/adapterframework/http/rest/ApiServiceDispatcher.java
```diff
@@ -210,6 +221,35 @@ public static Json2XmlValidator getJsonValidator(PipeLine pipeline) {
...
+				param.add("schema", Json.createObjectBuilder().add("type", "string"));
+				paramBuilder.add(param);
+			}
+		}
+		Json2XmlValidator validator = getJsonValidator(adapter.getPipeLine());
+		if(validator != null && !validator.getParameterList().isEmpty()) {
+			for (Parameter parameter : validator.getParameterList()) {
+				JsonObjectBuilder param = Json.createObjectBuilder();
+				param.add("name", parameter.getName());
+				param.add("in", "query");
```
gvanbrakel, line 244: You should not map parameters by name, but by sessionKey

core/src/test/java/nl/nn/adapterframework/http/rest/OpenApiTest.java
```diff
@@ -18,17 +19,126 @@
...
+		dispatcher.getPatternClients().clear();
+	}
+
+	@Test
+	@IsolatedThread
+	public void simpleEndpointPostTest() throws Exception {
+		ApiServiceDispatcher dispatcher = ApiServiceDispatcher.getInstance();
+		assertEquals("there are still registered patterns! Threading issue?", 0, dispatcher.getPatternClients().size());
+
+		new AdapterBuilder("myAdapterName", "description4simple-get").setListener("simpleEndpointPostTest", "post", null).setValidator("simple.xsd", null, "user", null).addExit("200").build(true);
```
gvanbrakel, line 42: In this case (i.e. when building a configuration) it is more clear to have a separate line for each method

gvanbrakel (changes requested):

core/src/main/java/nl/nn/adapterframework/http/rest/ApiServiceDispatcher.java
```diff
@@ -171,27 +175,34 @@ protected JsonObject generateOpenApiJsonSchema(Collection<ApiDispatchConfig> cli
...
 					//ContentType may have more parameters such as charset and formdata-boundry
 					MediaTypes produces = listener.getProducesEnum();
 					methodBuilder.add("responses", mapResponses(adapter, produces, schemas));
 				}
-
 				methods.add(method.toLowerCase(), methodBuilder);
 			}
-			paths.add("/"+config.getUriPattern(), methods);
+			if(listener != null) {
+				String uriPattern = listener.getUriPattern().startsWith("/") ? listener.getUriPattern() : "/" + listener.getUriPattern();
```
nielsm5, line 203: I think the `getUriPattern()` should always return a usable pattern with or without prepending slash. 
That way this 'check' would not be necessarily, and we will have a uniform way to displaying the pattern.

core/src/test/resources/OpenApi/envelope.json
```diff
@@ -0,0 +1,105 @@
...
+        "description":"OpenApi document auto-generated by Frank!Framework for test",
+        "version":"0.1.9"
+    },
+    "paths":{
+        "/simpleEndpointWithOperationIdTest":{
+            "get":{
+                "summary":"get envelope adapter description",
+                "operationId":"operationId",
+                "parameters":[
+                ],
```
nielsm5, line 14: Should this block exist if when it's empty?

nielsm5 (changes requested):

gvanbrakel (approved):

core/src/test/java/nl/nn/adapterframework/http/rest/ApiListenerPatternsTest.java
```diff
@@ -68,7 +68,9 @@ public ApiListenerPatternsTest(String pattern, String expected) {
 
 	@Test
 	public void testUriPattern() {
-		assertEquals(uriPattern, listener.getUriPattern());
+		String preFormattedUriPattern = uriPattern.startsWith("/") ? uriPattern : "/" + uriPattern;
+		String expectedUriPattern = preFormattedUriPattern.endsWith("/") ? preFormattedUriPattern.substring(0, preFormattedUriPattern.length()-1) : preFormattedUriPattern;
+		assertEquals(expectedUriPattern, listener.getUriPattern());
 	}
```
nielsm5, line 74: Right now this is a 'hack' so you don't have to change the expected results. On line 39 to 56 you can see the input/output from the `setUriPattern(...)` and `getUriPattern()`
In this case, since the method returns a different value the test should reflect this, **(and not manipulate the test results)**.
I think, the `getUriPattern()` method as well as the `getCleanPattern()` should be tested, now that you moved some functionality over to other methods.

nielsm5 (changes requested):

nielsm5 (approved):

nielsm5:
Travis fails: `[ERROR] Tests run: 3088, Failures: 3, Errors: 0, Skipped: 158`

PR #1228, opened 2020-11-03
Add IbisDoc for some child elements

gvanbrakel:


core/src/main/java/nl/nn/adapterframework/batch/IRecordHandlerManager.java
```diff
@@ -38,6 +39,7 @@
 	/**
 	 * @param flow New flow to be added to the managed flow elements
 	 */
+	@IbisDoc("flow to be added to the managed flow elements")
```
mhdirkse, line 42: This description starts with a lower-caps letter. How about starting with a capital to form a sentence?

mhdirkse: I do not see a number here, so no order is specified here yet for the XSD. Do you want that the XSD requests the same order here as the order of the methods?

core/src/main/java/nl/nn/adapterframework/batch/StreamTransformerPipe.java
```diff
@@ -219,9 +208,7 @@ public void registerChild(IRecordHandlerManager manager) throws Exception {
 		ConfigurationWarnings.add(this, log, "configuration using element 'child' is deprecated. Please use element 'manager'", ConfigurationWarnings.DEPRECATION_SUPPRESS_KEY, getAdapter());
 		registerManager(manager);
 	}
-	/**
-	 * Register a uniquely named manager.
-	 */
+	@IbisDoc({"1", "a uniquely named manager"})
```
mhdirkse, line 211: I do not agree with "1" for the order here. We already put some @IbisDoc annotations for sort orders in an earlier pull request. There we use 10, 20, 30 etc, such that we have the option to insert new methods anywhere in the existing sort order.

core/src/main/java/nl/nn/adapterframework/batch/StreamTransformerPipe.java
```diff
@@ -1,5 +1,5 @@
 /*
-   Copyright 2013, 2018, 2020 Nationale-Nederlanden
+   Copyright 2013, 2018 Nationale-Nederlanden, WeAreFrank!
```
nielsm5, line 2: welk jaar?

nielsm5 (changes requested):

nielsm5 (approved):

core/src/main/java/nl/nn/adapterframework/batch/StreamTransformerPipe.java
gvanbrakel: done

PR #1248, opened 2020-11-09
Use schema name instead of db name when using MSSQL

alisihab:


gvanbrakel (changes requested):
Could you please add a test for this in DbmsSupportTest?

gvanbrakel (approved):

nielsm5 (approved):

PR #1274, opened 2020-11-18
Show a warning for transaction timeout

alisihab:
In case the transaction timeout is greater than the maximum transaction timeout, will add a configuration warning to be shown in the console


core/src/main/java/nl/nn/adapterframework/receivers/Receiver.java
```diff
@@ -703,8 +703,15 @@ public void configure() throws ConfigurationException {
 							ConfigurationWarnings.add(this, log, "has a transaction timeout ["+getTransactionTimeout()+"] which exceeds the system transaction timeout ["+stt+"]");
 						}
 					}
+					String maximumTransactionTimeout = Misc.getMaximumTransactionTimeout();
```
gvanbrakel, line 706: Could you please change Misc.getMaximumTransactionTimeout() and Misc.getSystemTransactionTimeout() in returning 'Integer' instead of 'String'?

gvanbrakel (changes requested):

gvanbrakel (approved):

nielsm5 (approved):

PR #1326, opened 2020-12-08
Fix #1033 Update adapter2dot xsl file

### LINKED ISSUE #1033: Flowcharts and documentation break without hardcoded success forwards
**Describe the bug**
When <**Forward name="success"** > has not been set, while a <**Forward name="exception"** > has been set, the documentation and flowcharts break. Also XmlIf pipes do not work without explicit forwards, and the firstPipe attribute cannot be omitted.

The ability to work without hardcoded forwards or FirstPipes is a welcome one, as pipes tend to be moved around and new pipes will be interposed, and this minimizes errors from incorrect forwards. However, this bug, especially for the flowcharts which are used as concise way of showing the adapter flow to e.g. testers, all but prevents us to use this functionality.

**Reporter**
Bastiaan Udo / Ali Akcan

**Category** 
 - Frank Console

**Expected behavior**
All pipes should be correctly included in the flowcharts and documentation.

**Example**
adapter code:
https://vcs.aws.insim.biz/ibissource/dm/ibis4oma/ibis4oma_oma/-/blob/master/src/main/configuration/oma/Configuration_MarkBulkMessageReady.xml

corresponding Flowchart:
https://srdzzapp0424.insim.biz/ibis4oma/iaf/api/adapters/MarkBulkMessageReady/flow

corresponding documentation
https://srdzzapp0424.insim.biz/ibis4oma/iaf/gui/#!/customView/Documentation >> oma >> MarkBulkMessageReady (full link not possible?)

alisihab:
Closes #1033


core/src/main/resources/xsl/adapter2dot.xsl
```diff
@@ -39,7 +39,14 @@
 		</xsl:choose>
 		<xsl:value-of select="$space" />
 		<xsl:text>&quot;</xsl:text>
-		<xsl:value-of select="@firstPipe" />
+		<xsl:choose>
+			<xsl:when test="string-length(@firstPipe) > 0">
+				<xsl:value-of select="@firstPipe" />
+			</xsl:when>
+			<xsl:otherwise>
+				<xsl:value-of select="//pipe[1]/@name" />
```
gvanbrakel, line 47: The // operator searches the whole document for (in this case) a pipe element. 
I think the following would be better:
``select="pipe[1]/@name"``

core/src/main/resources/xsl/adapter2dot.xsl
```diff
@@ -98,41 +105,41 @@
...
+						<xsl:with-param name="forwardPath" select="$elseForwardName" />
 						<xsl:with-param name="forwardName">elseForwardName</xsl:with-param>
 					</xsl:call-template>
 				</xsl:if>
 			</xsl:when>
 			<xsl:otherwise>
-				<xsl:if test="forward=false()">
+				<xsl:if test="count(forward[@name='success'])=0">
 					<xsl:call-template name="forward">
 						<xsl:with-param name="pipeName" select="@name" />
```
gvanbrakel, line 144: I think you'll need to calculate the 'next' pipe in the pipeline, using ``following-sibling::*[1]/@name``

gvanbrakel (changes requested):
Could you also make a unit test for these cases?

gvanbrakel (approved):

core/src/main/resources/xsl/adapter2dot.xsl
ricardovh: Should it not be `following-sibling::pipe[1]/@name`, to be sure that the XPath goes to a pipe and not something else if someone made a mistake in the configuration?

And what if there is no following pipe? Should it then go to an exit with state `success`? Or does that not matter for this XSLT?

gvanbrakel: @alisihab pointed out to me that the name is already determined

alisihab: There it only checks if there is no success forward configured. Please examine what will happen in the next lines if the condition holds.

nielsm5 (approved):

gvanbrakel (approved):

PR #1344, opened 2020-12-16
Introduce javascript.engines property for flowchart

alisihab:


gvanbrakel (approved):

core/src/main/java/nl/nn/adapterframework/extensions/graphviz/GraphvizEngine.java
```diff
@@ -156,21 +155,22 @@ private String getVisJsWrapper() {
 
 		Engine(String initScript, String graphvisJsLibrary) {
 			// Available JS Engines. Lower index has priority.
-			Class<?>[] engines = new Class[]{J2V8.class, Nashorn.class};
+			String[] engines = AppConstants.getInstance().getString("javascript.engines", "nl.nn.adapterframework.extensions.javascript.J2V8,nl.nn.adapterframework.extensions.javascript.Nashorn").split(",");
```
nielsm5, line 158: I think this should be done once, (moved to the top). Otherwise this will happen every time a new engine is generated..

You can, instead of using split, use `AppConstants.getTokenizedProperty()`

core/src/main/resources/AppConstants.properties
```diff
@@ -50,6 +50,7 @@ upload.dir=${ibis.tmpdir}
 
 flow.adapter.dir=${ibis.tmpdir}/flow/adapter
 flow.config.dir=${ibis.tmpdir}/flow/config
+javascript.engines=nl.nn.adapterframework.extensions.javascript.J2V8,nl.nn.adapterframework.extensions.javascript.Nashorn
```
nielsm5, line 53: Since this property is solely for the flow diagrams maybe change it to `flow.javascript.engine`

core/src/main/java/nl/nn/adapterframework/extensions/graphviz/GraphvizEngine.java
```diff
@@ -156,21 +155,22 @@ private String getVisJsWrapper() {
...
 
 			for (int i = 0; i < engines.length && jsEngine == null; i++) {
 				try {
-					log.debug("Trying Javascript engine [" + engines[i].getName() + "] for Graphviz.");
-					JavascriptEngine<?> engine = ((JavascriptEngine<?>) engines[i].newInstance());
+					Class<?> clazz = Class.forName(engines[i]);
+					log.debug("Trying Javascript engine [" + engines[i] + "] for Graphviz.");
+					JavascriptEngine<?> engine = ((JavascriptEngine<?>) clazz.newInstance());
```
nielsm5, line 164: I think it would be a good idea, to have some form of instanceof check here, to validate if the new class instance is indeed castable to `JavascriptEngine` and if not returns a readable error message. If you move all this logic to the constructor, instead of keeping a `String[]` you can keep using the `Class[]`.

nielsm5 (changes requested):

nielsm5 (approved):

Issue #488, opened 2020-03-11
Browse Tables with psql fails on column rownum


ricardovh:
**Describe the bug**
When using Browse Tables with a PostgreSQL database, browsing fails because column rownum does not exist:
(PSQLException) SQLState [42703]: ERROR: column "rownum" does not exist Position: 45

Using SELECT * FROM <TABLE> on EXECUTE QUERY does work properly.

**Category**
 - Frank Console

**To Reproduce**
1. Go to 'Browse Tables'
2. Fill in a table that exists
3. Send the request

**Expected behavior**
Browsing tables works properly when using PostgreSQL

**Screenshots**
N/A

**Environment:**
 - Application server (f.e. Tomcat/websphere/JBOSS/Jetty): Tomcat
 - Browser [e.g. chrome, safari]: Chrome
 - IAF Version [e.g. 7.4]: 7.5-RC1

**Additional context**
N/A

jjansenvr:
PostgreSQL is not supported yet it is on the list to add

We now support (DB2)/Oracle/MSSQL/MySQL/MariaDB(except XA and multithreaded jdbclisteners)

PR #1377, opened 2021-01-07
Fix #1272, #488 browse jdbc table for different databases

### LINKED ISSUE #1272: Browse JDBC "Number of rows only" does not display number of rows
**Describe the bug**
When using Browse JDBC with the option "Number of rows only" checked, the number of rows is not actually shown. It shows all columns in the table, but no results are displayed because there is no column for the ROWCOUNTER that should contain the COUNT(*)

**Reporter**
Ricardo

**Category** 
 - Frank Console

**To Reproduce**
1. Go to 'Browse tables'
2. Check 'Number of rows only' and fill in the other fields
3. Press send
3. Scroll down to 'result'
4. See no result shown

**Expected behavior**
The number of rows to be displayed.

**Screenshots**
image

**Environment:**
 - Application server (f.e. Tomcat/WebSphere/JBOSS/Jetty): Websphere
 - Browser [e.g. chrome, safari]: Chrome
 - IAF Version [e.g. 7.5 RC3, 7.6]: 7.6-20201113.112750

**Additional context**
N/A

alisihab:


core/src/main/java/nl/nn/adapterframework/webcontrol/api/BrowseJdbcTable.java
```diff
@@ -134,32 +136,58 @@ public Response execute(LinkedHashMap<String, Object> json) throws ApiException
...
+					String field = null;
+					if(!numberOfRowsOnly) {
+						field = "<field name=\""+rnumColumnName+"\" type=\"INTEGER\" />";
+						fielddefinition.append(field);
+						fieldDef.put(rnumColumnName, "INTEGER");
+						while(rs.next()) {
+							field = "<field name=\"" + rs.getString(4) + "\" type=\"" + DB2XMLWriter.getFieldType(rs.getInt(5)) + "\" size=\"" + rs.getInt(7) + "\"/>";
+							fielddefinition.append(field);
+							fieldDef.put(rs.getString(4), DB2XMLWriter.getFieldType(rs.getInt(5)) + "("+rs.getInt(7)+")");
+						}
```
gvanbrakel, line 155: Would it be possibe to use descriptive constants or column names, instead of '4', '5' and '7'?

gvanbrakel (changes requested):

gvanbrakel (approved):

nielsm5 (approved):

PR #1386, opened 2021-01-11
Improve field contents and add labels to GUI 3 Webservices tab

alisihab:


core/src/main/java/nl/nn/adapterframework/receivers/Receiver.java
```diff
@@ -523,7 +527,11 @@ public void configure() throws ConfigurationException {
...
 			if (StringUtils.isEmpty(getName())) {
 				if (getListener()!=null) {
-					setName(ClassUtils.nameOf(getListener()));
+					if(StringUtils.isNotEmpty(getListener().getName())) {
+						setName(getListener().getName() + " " + ClassUtils.nameOf(getListener()));
+					} else {
+						setName(ClassUtils.nameOf(getListener()));
+					}
 				} else {
 					setName(ClassUtils.nameOf(this));
```
gvanbrakel, line 536: I think it would be wise to check for other Receivers (in the same Adapter) that have the same name, and if they exist, disambiguate by adding a sequence number

gvanbrakel (changes requested):

core/src/main/java/nl/nn/adapterframework/core/Adapter.java
```diff
@@ -496,7 +496,7 @@ public long getNumOfMessagesProcessed() {
 	@Override
 	public Receiver getReceiverByName(String receiverName) {
 		for (Receiver receiver: receivers) {
-			if (receiver.getName().equalsIgnoreCase(receiverName)) {
+			if (receiverName.equalsIgnoreCase(receiver.getName())) {
```
gvanbrakel, line 499: receiverName can be null when getReceiverByName is called, e.g. from the GUI. 
We must avoid NullPointerExceptions!

core/src/main/java/nl/nn/adapterframework/core/Adapter.java
```diff
@@ -508,7 +508,7 @@ public Receiver getReceiverByNameAndListener(String receiverName, Class<?> liste
 			return getReceiverByName(receiverName);
 		}
 		for (Receiver receiver: receivers) {
-			if (receiver.getName().equalsIgnoreCase(receiverName) && listenerClass.equals(receiver.getListener().getClass())) {
+			if (receiverName.equalsIgnoreCase(receiver.getName()) && listenerClass.equals(receiver.getListener().getClass())) {
```
gvanbrakel, line 511: same here

core/src/main/java/nl/nn/adapterframework/receivers/Receiver.java
```diff
@@ -726,6 +726,33 @@ public void configure() throws ConfigurationException {
 		}
 	}
 
+	private String generateName() {
+		String name = null;
+		if (getListener()!=null) {
+			if(StringUtils.isNotEmpty(getListener().getName())) {
+				name = getListener().getName() + " " + ClassUtils.nameOf(getListener());
```
gvanbrakel, line 733: I would prefer to extend the Receiver name with the listener-classname only when there are multiple unnamed Receivers in the Adapter that have the same name.

gvanbrakel (changes requested):

core/src/main/java/nl/nn/adapterframework/receivers/Receiver.java
```diff
@@ -726,6 +726,33 @@ public void configure() throws ConfigurationException {
 		}
 	}
 
+	private String generateName() {
```
nielsm5, line 729: I think this logic should come from the Adapter and not the receiver

nielsm5 (changes requested):

gvanbrakel (approved):

nielsm5 (approved):

PR #1394, opened 2021-01-14
Simplify active.xsl

ricardovh:


ricardovh:
Test fails because https://github.com/ibissource/iaf/blob/master/core/src/test/java/nl/nn/adapterframework/util/TransformerPoolTest.java#L76 and https://github.com/ibissource/iaf/blob/master/core/src/test/java/nl/nn/adapterframework/util/TransformerPoolTest.java#L80 force xsltVersion 1, which does not have the string function lower-case.

I'm not sure what should be done with this testcase, as it seems unrelated to the change.

gvanbrakel:
In the failing TransformerPoolTest the xsl can be replaced with another xsl, that does work under XSLT 1 and XSLT 2

nielsm5 (approved):

PR #1395, opened 2021-01-14
Introduce increment parameter in IncreaseIntegerPipe

### LINKED ISSUE #1127: IncreaseIntegerPipe with parameter value
**Reporter**
Ricardo

I want to increase an integer session key with a variable value (some sessionkey/xpathExpression), but currently I can only increment with a fixed value using the IncreaseIntegerPipe.

To have a parameter on the IncreaseIntegerPipe with which I can update the session key.

Current workaround would be to use an XsltPipe with an xpathExpression and store that in a session key.

**Additional context**
N/A

alisihab:
Closes #1127

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/pipes/IncreaseIntegerPipe.java
+++ b/core/src/main/java/nl/nn/adapterframework/pipes/IncreaseIntegerPipe.java
@@ -1,5 +1,5 @@
 /*
-   Copyright 2013, 2020 Nationale-Nederlanden, 2020 WeAreFrank!
+   Copyright 2013, 2020 Nationale-Nederlanden, 2020, 2021 WeAreFrank!
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
@@ -16,19 +16,25 @@
 package nl.nn.adapterframework.pipes;
 
 import nl.nn.adapterframework.configuration.ConfigurationException;
+import nl.nn.adapterframework.configuration.ConfigurationWarning;
 import nl.nn.adapterframework.core.IPipeLineSession;
+import nl.nn.adapterframework.core.ParameterException;
 import nl.nn.adapterframework.core.PipeRunException;
 import nl.nn.adapterframework.core.PipeRunResult;
 
 import nl.nn.adapterframework.doc.IbisDoc;
+import nl.nn.adapterframework.parameters.ParameterValue;
+import nl.nn.adapterframework.parameters.ParameterValueList;
 import nl.nn.adapterframework.stream.Message;
 
 import org.apache.commons.lang.StringUtils;
 
 /**
  * Pipe that increases the integer values of a session variable.
- * Used to in combination with {@link CompareIntegerPipe} to contstruct loops.
- *
+ * Can be used in combination with {@link CompareIntegerPipe} to construct loops.
+ * 
+ * <p>
+ * <table border="1">
  * <tr><td>{@link #setSessionKey(String) sessionKey}</td><td>reference to the session variable whose value is to be increased</td><td></td></tr>
  * <tr><td>{@link #setIncrement(int) increment}</td><td>amount to increment the value</td><td>1</td></tr>
  * </table>
@@ -39,6 +45,7 @@ public class IncreaseIntegerPipe extends FixedForwardPipe {
 
 	private String sessionKey=null;
 	private int increment=1;
+	private final static String INCREMENT = "increment";
 
 	@Override
 	public void configure() throws ConfigurationException {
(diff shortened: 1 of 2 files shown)


core/src/main/java/nl/nn/adapterframework/pipes/IncreaseIntegerPipe.java
```diff
@@ -39,6 +45,7 @@
 
 	private String sessionKey=null;
 	private int increment=1;
+	private final static String INCREMENT = "increment";
```
gvanbrakel, line 48: I would name this constant PARAMETER_INCREMENT, to indicate that it is not just the word 'increment', but meant as the name of a parameter

core/src/main/java/nl/nn/adapterframework/pipes/IncreaseIntegerPipe.java
```diff
@@ -68,7 +89,9 @@ public String getSessionKey() {
 		return sessionKey;
 	}
 
+	@Deprecated
 	@IbisDoc({"amount to increment the value", "1"})
+	@ConfigurationWarning("Please use the parameter ["+INCREMENT+"]")
```
gvanbrakel, line 94: It is not necessary to deprecate the attribute

core/src/main/java/nl/nn/adapterframework/pipes/IncreaseIntegerPipe.java
```diff
@@ -52,6 +59,20 @@ public PipeRunResult doPipe(Message message, IPipeLineSession session) throws Pi
...
+			try {
+				pvl = getParameterList().getValues(message, session);
+			} catch (ParameterException e) {
+				throw new PipeRunException(this, getLogPrefix(session) + "exception extracting parameters", e);
+			}
+		}
+		ParameterValue pv = pvl.getParameterValue(INCREMENT);
+		if(pv != null && pv.getValue() != null) {
+			increment = pv.asIntegerValue(increment);
+		}
```
gvanbrakel, line 74: The parameter value is meant to be used at runtime, so in doPipe(), and not in configure()

gvanbrakel (changes requested):

core/src/main/java/nl/nn/adapterframework/pipes/IncreaseIntegerPipe.java
nielsm5: You could check (during configure) if the setter has been set/changed and if a parameter has been supplied. If so the setter will be redundant.

core/src/main/java/nl/nn/adapterframework/pipes/IncreaseIntegerPipe.java
```diff
@@ -52,6 +57,20 @@ public PipeRunResult doPipe(Message message, IPipeLineSession session) throws Pi
...
+			try {
+				pvl = getParameterList().getValues(message, session);
+			} catch (ParameterException e) {
+				throw new PipeRunException(this, getLogPrefix(session) + "exception extracting parameters", e);
+			}
+		}
+		ParameterValue pv = pvl.getParameterValue(PARAMETER_INCREMENT);
+		if(pv != null && pv.getValue() != null) {
+			increment = pv.asIntegerValue(increment);
+		}
```
nielsm5, line 72: The != check is also present in `asIntegerValue`.
Right now `pv.asIntegerValue(increment);` will never use the argument `increment`.
I think it's better if you either remove the != null check, or replace this argument with 0 or something?

nielsm5 (changes requested):

core/src/main/java/nl/nn/adapterframework/pipes/IncreaseIntegerPipe.java
```diff
@@ -39,20 +43,37 @@
...
+			try {
+				pvl = getParameterList().getValues(message, session);
+			} catch (ParameterException e) {
+				throw new PipeRunException(this, getLogPrefix(session) + "exception extracting parameters", e);
+			}
+		}
+		ParameterValue pv = pvl.getParameterValue(PARAMETER_INCREMENT);
+		if(pv != null) {
+			incrementBy = pv.asIntegerValue(increment);
+		}
```
gvanbrakel, line 74: The first line will fail with NPE when parameterList == null. 
Either remove the test `` getParameterList() != null ``, or put these lines inside the guarded block

gvanbrakel (changes requested):

gvanbrakel (approved):

core/src/test/java/nl/nn/adapterframework/pipes/IncreaseIntegerPipeTest.java
```diff
@@ -45,4 +46,32 @@ public void testCannotIncreaseBy2AsNoSessionKey() throws Exception {
...
+		session.put(numberSession, "4");
+		Parameter inc = new Parameter();
+		inc.setName("increment");
+		inc.setValue(null);
+		pipe.addParameter(inc);
+		pipe.setSessionKey(numberSession);
+		pipe.configure();
+		doPipe(pipe, null, session);
+		assertEquals("5", session.get(numberSession));
+    }
```
nielsm5, line 75: For the sake of completeness, could you also add a test with an empty String?
`testEmptyIncrementParameter()`

nielsm5 (approved):

Issue #494, opened 2020-03-12
Health check endpoints are always protected, which is not the preferred method on AWS


nielsm5:
**Is your feature request related to a problem? Please describe.**
Aws does health checks on certain endpoints it would be nice if it could access the ibis health endpoint without authentication

**Describe the solution you'd like**
property to turn on authentication on health check endpoints

**Describe alternatives you've considered**
An nginx proxy that adds an authentication header

nielsm5:
endpoints are already set to PermitAll, since they are filtered via the web.xml, there is a global constraint on all /iaf/api/* requests. Once this will be loaded dynamically (spring/beans?) it will solve this issue

nielsm5:
Add new healthcheck servlet endpoint, which you can explicitly enable/disable

nielsm5:
No longer required. See [AWS health-enhanced](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/health-enhanced.html?icmpid=docs_elasticbeanstalk_console#health-enhanced-authz).

PR #1419, opened 2021-01-26
Authorized health check

Laurens-makel:
Possible solution for #494.

Current implementation has by default authorisation **disabled**  and checks if there is a remote user present, it does not check if the user has any roles.

A possible improvement could be to add a few properties which can be used to specify which roles to perform a check on.


jacodg:
We are working on another solution (@jjansenvr will create an issue)

PR #1434, opened 2021-02-03
Migrator implements autoClosable

### LINKED ISSUE #1433: Migrator leaves connection open
**Describe the bug**
With the migrator enabled long deployment times observed. The cause of the long deployments have been investigated. It is found that the migrator gets a connection and never closes it.

**Reporter**
ibis4oma

**Expected behavior**
Close connection after using the migrator.

**Screenshots**

**Environment:**
 - Application server: [e.g. Tomcat/WebSphere/JBOSS/Jetty]
 - DBMS: [e.g. Oracle, MSSQL, MariaDB, MySql,PostgreSQL]
 - Browser: [e.g. Chrome, Safari, Edge]
 - IAF Version: [e.g. 7.5, 7.6]

alisihab:
closes #1433

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/jdbc/migration/LiquibaseImpl.java
+++ b/core/src/main/java/nl/nn/adapterframework/jdbc/migration/LiquibaseImpl.java
@@ -25,7 +25,9 @@ import liquibase.Contexts;
 import liquibase.LabelExpression;
 import liquibase.Liquibase;
 import liquibase.changelog.ChangeSet;
+import liquibase.database.Database;
 import liquibase.database.jvm.JdbcConnection;
+import liquibase.exception.DatabaseException;
 import liquibase.exception.LiquibaseException;
 import liquibase.resource.ClassLoaderResourceAccessor;
 import nl.nn.adapterframework.configuration.Configuration;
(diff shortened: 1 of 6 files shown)


gvanbrakel (approved):

core/src/main/java/nl/nn/adapterframework/jdbc/migration/Migrator.java
```diff
@@ -38,7 +39,7 @@
  * @since	7.0-B4
  *
  */
-public class Migrator extends JdbcFacade {
+public class Migrator extends JdbcFacade implements AutoCloseable {
```
nielsm5, line 42: The Spring term would be `DisposableBean` instead of AutoClosable.
Either way, since it's a Singleton bean neither will be called automatically.
I think it would be wise to add a comment somewhere explaining this behavior and the problem that occurs when the JDBC connection isn't closed.

core/src/main/java/nl/nn/adapterframework/jdbc/migration/LiquibaseImpl.java
```diff
@@ -108,4 +110,13 @@ public Writer getUpdateScript(Writer writer) throws LiquibaseException {
...
 	}
+
+	public void close() throws DatabaseException {
+		if(liquibase != null) {
+			Database db = liquibase.getDatabase();
+			if(db != null) {
+				db.close();
+			}
+		}
+	}
```
nielsm5, line 121: Also add DisposableBean or AutoClosable here :)

nielsm5 (changes requested):

nielsm5 (approved):

PR #1446, opened 2021-02-08
Display user in gui

### LINKED ISSUE #1382: Display logged in user in GUI
**Reporter**
Ricardo

This morning I had to help with a deployment, but the user could not log in. Further investigation indicated that this was because a session was already active in his browser for a user that has no permissions and the session was re-used when logging in from the loginscreen with different credentials. 

Display the username in the GUI.

N/A

**Additional context**
Issue with sessions might be related to discussion Niels and I had regarding #1232, that there is a difference between the browser and the application server.

alisihab:
Closes #1382

### DIFF
--- a/webapp/src/main/webapp/iaf/gui/views/common/topnavbar.html
+++ b/webapp/src/main/webapp/iaf/gui/views/common/topnavbar.html
@@ -42,9 +42,10 @@
                     </li> -->
                 </ul>
             </li>
-            <li ng-if="::userName">{{::userName}}
+            <li ng-if="::userName">{{::userName}}</li>
+            <li ng-if="::loggedin">
                 <a ui-sref="logout">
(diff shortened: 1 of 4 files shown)


webapp/src/main/webapp/iaf/gui/views/common/topnavbar.html
```diff
@@ -42,9 +42,9 @@
                     </li> -->
                 </ul>
             </li>
-            <li ng-if="loggedin">
+            <li ng-if="::userName">{{::userName}}
```
nielsm5, line 45: I think the if statement should remain, the logout button is not directly related to the user. 
You can display the username here, when available of course :)

nielsm5 (changes requested):

nielsm5 (approved):

gvanbrakel (approved):

PR #1450, opened 2021-02-09
#1233 Receiver and ErrorStorage

### LINKED ISSUE #1233: Some preferred features for the error storage
**Reporter**
Laura

I'm Ops of the Day. Sometimes there's a problem in the ibis which gives a lot of error messages in the error storage. For example a "presumedTimeout" from another application. Sometimes I have almost 1000 the same errors in te error storage.
Everytime selecting 100 and pressing "resend selected", takes a lot of time, because the resending process is very slow (it takes more than 5 minutes for 100 messages). (Thread by tread)

- It would be nice if you can select and resend more then 100 messages in the errorstorage at once, so for example 1000
- The resend process goes very slow, it seems one by one. Is multithreaded resending a possibility?

**Additional context**

dropdown

gvanbrakel:
Closes #1233


core/src/main/java/nl/nn/adapterframework/webcontrol/api/TransactionalStorage.java
```diff
@@ -277,6 +284,48 @@ public Response resendReceiverMessages(
 		return Response.status(Response.Status.ACCEPTED).entity(errorMessages).build();
 	}
 
+	@POST
+	@RolesAllowed({"IbisDataAdmin", "IbisAdmin", "IbisTester"})
+	@Path("/adapters/{adapterName}/receivers/{receiverName}/errorstorage/changeProcessState")
```
gvanbrakel, line 289: Now this only works for errorStorage.
If a Hold storage is introduced (which is as simple as adding a statusValueHold attribute to a JdbcTableListener), then it should be possible to move messages from errorStorage to Available and to Hold, but also from Hold to Error.
I would expect the path to be something like:
``@Path("/adapters/{adapterName}/receivers/{receiverName}/storage/{processState}/changeto/{targetState}")``

webapp/src/main/webapp/iaf/gui/js/controllers.js
```diff
@@ -1453,6 +1455,21 @@ angular.module('iaf.beheerconsole')
...
 		});
 	}
+	$scope.changingProcessState = false;
+	$scope.changeProcessState = function(targetState) {
+		$scope.changingProcessState = true;
+		var data = getFormData();
+		data.append("targetState", targetState);
+		Api.Post($scope.base_url+"/changeProcessState", data, function() {
+			$scope.changingProcessState = false;
+			$scope.addNote("success", "Successfully moved messages to Available");
```
gvanbrakel, line 1465: 'Available' should reflect the targetState

core/src/main/java/nl/nn/adapterframework/webcontrol/api/TransactionalStorage.java
```diff
@@ -277,6 +284,48 @@ public Response resendReceiverMessages(
...
+			throw new ApiException("Adapter not found!");
+		}
+
+		Receiver<?> receiver = adapter.getReceiverByName(receiverName);
+		if(receiver == null) {
+			throw new ApiException("Receiver ["+receiverName+"] not found!");
+		}
+		ProcessState targetState = ProcessState.valueOf(resolveStringFromMap(input, "targetState"));
+		String[] messageIds = getMessages(input);
+		IMessageBrowser<?> errorStorageBrowser = receiver.getMessageBrowser(ProcessState.ERROR);
```
gvanbrakel, line 311: browser to be retrieved must be selected by path variable

core/src/main/java/nl/nn/adapterframework/webcontrol/api/TransactionalStorage.java
```diff
@@ -277,6 +284,48 @@ public Response resendReceiverMessages(
...
+
+		if(adapter == null){
+			throw new ApiException("Adapter not found!");
+		}
+
+		Receiver<?> receiver = adapter.getReceiverByName(receiverName);
+		if(receiver == null) {
+			throw new ApiException("Receiver ["+receiverName+"] not found!");
+		}
+		ProcessState targetState = ProcessState.valueOf(resolveStringFromMap(input, "targetState"));
```
gvanbrakel, line 309: targetState to be selected via path (or other input parameter)

gvanbrakel (commented):
The Issue also requests the option to have the number of message that is browsed to be increased to 1000. Did you address that too?

core/src/main/java/nl/nn/adapterframework/core/ProcessState.java
```diff
@@ -15,27 +15,39 @@
...
-	HOLD;
+	AVAILABLE("Available", "fa-server", "success"),
+	INPROCESS("InProcess", "fa-share", "success"),
+	DONE("Done", "fa-envelope-o", "success"),
+	ERROR("Error", "fa-times-circle", "danger"),
+	HOLD("Hold", "fa-pause-circle", "warning");
+
+	String name=null;
+	String iconName=null;
+	String type=null;
```
gvanbrakel, line 35: Could you comment briefly on what 'type' means/ is used for

core/src/main/java/nl/nn/adapterframework/core/ProcessState.java
```diff
@@ -50,4 +62,35 @@
...
+	public String getType() {
+		return type;
+	}
+
+	public void setName(String name) {
+		this.name = name;
+	}
+	public String getName() {
+		return name;
+	}
```
gvanbrakel, line 95: You could replace these getters and setters with Lombok field annotations ``@Getter`` and ``@Setter``

core/src/main/java/nl/nn/adapterframework/webcontrol/api/TransactionalStorage.java
```diff
@@ -277,9 +270,52 @@ public Response resendReceiverMessages(
...
+		Receiver<?> receiver = adapter.getReceiverByName(receiverName);
+		if(receiver == null) {
+			throw new ApiException("Receiver ["+receiverName+"] not found!");
+		}
+		String[] messageIds = getMessages(input);
+		IMessageBrowser<?> store = receiver.getMessageBrowser(ProcessState.getProcessStateFromName(processState));
+		List<String> errorMessages = new ArrayList<String>();
+		for(int i=0; i < messageIds.length; i++) {
+			try {
+				if(!receiver.changeProcessState(store.browseMessage(messageIds[i]), ProcessState.getProcessStateFromName(targetState), null)) {
```
gvanbrakel, line 302: You must check that targetState is really in the receiver.targetStates

webapp/src/main/webapp/iaf/gui/views/ShowConfigurationStatus.html
```diff
@@ -391,7 +383,7 @@ <h4>Senders</h4>
                                     <tbody>
                                         <tr ng-repeat="pipes in adapter.pipes" ng-if="pipes.sender">
                                             <td>{{::pipes.name}}</td>
-                                            <td>{{::pipes.sender}}<span ng-if="pipes.destination"> ({{::pipes.destination}})</span></td>
+                                            <td>{{::pipes.sender}}<span ng-if="pipes.destination"> ({{::pipes.destination|truncate:150}})</span></td>
```
nielsm5, line 386: Kan je ook `title='{{pipes.destination}}'` toevoegen zodat men bij een 'on-hover' alles ziet?

nielsm5 (changes requested):

core/src/main/java/nl/nn/adapterframework/core/ProcessState.java
```diff
@@ -15,27 +15,39 @@
...
+	
 	public static Set<ProcessState> getMandatoryKnownStates() {
 		Set<ProcessState> knownProcessStates = new LinkedHashSet<>();
 		knownProcessStates.add(AVAILABLE);
 		return knownProcessStates;
 	}
 
 	public static Map<ProcessState, Set<ProcessState>> getTargetProcessStates(Set<ProcessState> knownProcessStates) {
-		Map<ProcessState, Set<ProcessState>> targetProcessStates = new HashMap<>();
+		Map<ProcessState, Set<ProcessState>> targetProcessStates = new LinkedHashMap<>();
```
gvanbrakel, line 50: Turning this into a LinkedHashMap suggests that the insertion order is significant. Is that the case? Are there iterations over this map that should be in the right order?

core/src/main/java/nl/nn/adapterframework/receivers/Receiver.java
```diff
@@ -294,8 +295,8 @@
 
 	private EventHandler eventHandler=null;
 
-	private Set<ProcessState> knownProcessStates;
-	private Map<ProcessState,Set<ProcessState>> targetProcessStates = new HashMap<>();
+	private Set<ProcessState> knownProcessStates = new LinkedHashSet<ProcessState>();
+	private Map<ProcessState,Set<ProcessState>> targetProcessStates = new LinkedHashMap<>();
```
gvanbrakel, line 299: Turning this into a LinkedHashMap suggests that the insertion order is significant. Is that the case? Are there iterations over this map that should be in the right order?

gvanbrakel (commented):
I would like to suggest to improve the layout of the message browsers on screen:
- have 'select all'/'deselect all' as a checkbox in the header, aligned with the individual selection checkbox;
- have the actions (e.g. 'delete', 'view', 'move to available', 'move to Hold') as small icons on the right side
- use arrow icons for the 'move to' actions, to the left for a move to a 'lower' process state, to the right for the move to a 'higher' process state

nielsm5 (approved):

PR #1459, opened 2021-02-10
Use correct UriResolver when using Xerces Jaxp parser

nielsm5:
It turns out that Xerces doesn't necessarily maintain the correct EntityResolver.
It uses the `XMLEntityResolver` instead of the sax `EntityResolver`. See [xerces2 javadoc](
https://xerces.apache.org/xerces2-j/javadocs/xerces2/org/apache/xerces/impl/XMLEntityManager.html) for more info.

It's important that the `XMLEntityResolver` never returns `null` so that the parser **won't** try to resolve the entity using a default/fallback mechanism!

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/xml/ClassLoaderURIResolver.java
+++ b/core/src/main/java/nl/nn/adapterframework/xml/ClassLoaderURIResolver.java
@@ -75,12 +75,13 @@ public class ClassLoaderURIResolver implements URIResolver {
 		}
 
 		String ref=ref1;
-		Resource resource = Resource.getResource(scopeProvider, ref, protocol);
+		Resource resource = Resource.getResource(scopeProvider, ref);
 		if (resource==null && ref2!=null) {
 			if (log.isDebugEnabled()) log.debug("Could not resolve href ["+href+"] base ["+base+"] as ["+ref+"], now trying ref2 ["+ref2+"] protocol ["+protocol+"]");
 			ref=ref2;
-			resource = Resource.getResource(scopeProvider, ref, protocol);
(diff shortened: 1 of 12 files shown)


gvanbrakel:
> It turns out that Xerces doesn't necessarily maintain the correct EntityResolver.
> It uses the `XMLEntityResolver` instead of the sax `EntityResolver`. See [xerces2 javadoc](https://xerces.apache.org/xerces2-j/javadocs/xerces2/org/apache/xerces/impl/XMLEntityManager.html) for more info.
> 
> It's important that the `XMLEntityResolver` never returns `null` so that the parser **won't** try to resolve the entity using a default/fallback mechanism!

And what is the default/fallback mechanism? isn't that just the normal entity resolver?

core/src/main/java/nl/nn/adapterframework/xml/ClassLoaderURIResolver.java
```diff
@@ -37,7 +37,7 @@
  */
 public class ClassLoaderURIResolver implements URIResolver {
 	protected Logger log = LogUtil.getLogger(this);
-	
+	private String allowedProtocols = "file"; //It's possible that the base is an absolute path starting with file. Since we don't know the relative base, allow the FILE protocol
```
gvanbrakel, line 40: Volgens mij moet je zeggen: je mag wel relatief ten opzichte van de base, ook als die protcol file heeft, maar niet absoluut met protocol file.
Dus niet zomaar protocol file toestaan

gvanbrakel (changes requested):

gvanbrakel (approved):

PR #1486, opened 2021-02-17
Allow reading properties from the environment in Tomcat container images…

ricardovh:
… (see [Tomcat documentation](https://tomcat.apache.org/tomcat-8.5-doc/config/systemprops.html#Property_replacements)).

This will allow container-orchestration platforms to provide secrets and additional properties to both Tomcat and the application via the environment (see [Kubernetes](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables)). Both the properties in Tomcat-files such as context.xml and in configurations use the same ${...} syntax and are replaced from the environment.

It is still needed to set `properties.hide` to hide the actual values from the GUI.


gvanbrakel (approved):

nielsm5:
Ik zie liever een implementatie om secrets uit een AWS Secrets Manager, Azure Key Vault of misschien zelfs een kubernetes cluster te halen.

nielsm5 (approved):

PR #1529, opened 2021-03-02
FF! docker image with WildFly

alisihab:


docker/appserver/WildFly/example/src/scripts/add_modules.py
```diff
@@ -0,0 +1,22 @@
...
+
+for jarFile in jarList:
+	resources += path_to_jars+jarFile+resourceDelimiter
+
+resources=resources[0:-1]
+
+command="module add --name="+moduleName+" --resources="+resources+" --resource-delimiter="+resourceDelimiter
+
+os.system(jboss_cli+" --command='"+command+"'")
+
```
nielsm5, line 22: If the modules are also in the standalone.xml this isn't required is it?

nielsm5 (changes requested):

docker/appserver/WildFly/example/src/scripts/add_modules.py
alisihab: These are resources and configurations that are provided as jars. It is not possible to just put them in the lib folder to make them available in classpath. It is required to add those as modules to the server. This is a way to add them as modules.

nielsm5 (approved):

PR #1570, opened 2021-03-12
Fix IndexOutOfBoundsException on abort when ignoring report

jacodg:


gvanbrakel (commented):
When I run iaf-test on this branch, there is still a report in progress after all scenarios have been tested.

nielsm5 (approved):

gvanbrakel (approved):

jacodg:
I also have one report in progress but it is already in progress before running the test scenario's. It's the Larva adapter that is started automatically after starting iaf-test. I don't know why the report stays in progress at this moment, needs some more investigation.

PR #1573, opened 2021-03-15
Add unit tests for rollingFile

alisihab:


core/src/test/java/nl/nn/adapterframework/util/FileUtilsTest.java
```diff
@@ -279,4 +283,85 @@ public void testIsFileBinaryEqual() throws Exception {
...
+
+		test.setLastModified(new GregorianCalendar(2020, 11, 25).getTime().getTime()); // change the last modified date for the file to be deleted.
+		test = FileUtils.getRollingFile(tempDir.getAbsolutePath(), "", FileUtils.WEEKLY_ROLLING_FILENAME_DATE_FORMAT, "", 7, null);
+		Files.createFile(Paths.get(test.getAbsolutePath()));
+		assertEquals(1, tempDir.listFiles().length); // test number of files 
+	}
+
+	@Test
+	public void testGetRollingFileTheSamefile() throws IOException {
+		File tempDir = FileUtils.createTempDir();
```
nielsm5, line 352: I think it's safer to use a `TemporaryFolder` (junit thingie)
It creates a folder which is test independent. So you can create files etc without affecting other tests.
Have a look at the DirectoryListenerTest, it's used there.

nielsm5 (changes requested):

nielsm5 (approved):

gvanbrakel (approved):

PR #1575, opened 2021-03-15
add dockerfile and config for local testing with activemq

thomasverweij:


test/src/main/webapp/META-INF/context.xml
```diff
@@ -149,7 +149,7 @@
 		name="jms/qcf-activemq"
 		factory="org.apache.naming.factory.BeanFactory"
 		type="org.apache.activemq.ActiveMQXAConnectionFactory"
-		brokerURL="vm://localhost"
+		brokerURL="tcp://host.docker.internal:61616"
```
gvanbrakel, line 152: This change makes the use of the docker container required for testing with ActiveMQ.
In the current situation, it is enough to have the activemq-all.jar on the classpath.
I would suggest to leave this context.xml-entry unchanged, and add a new one, named 'jms/qcf-activemq-external'

gvanbrakel (changes requested):

PR #1591, opened 2021-03-18
Introduce removeNonEmptyDirectory attribute for file systems

alisihab:


core/src/main/java/nl/nn/adapterframework/filesystem/FileSystemActor.java
```diff
@@ -691,4 +690,12 @@ public void setExcludeWildCard(String excludeWildCard) {
 	public String getExcludeWildCard() {
 		return excludeWildCard;
 	}
+
+	@IbisDoc({"12", "If set to true then the directory and the content of the non empty directory will be deleted."})
+	public void setRemoveNonEmptyDirectory(boolean removeNonEmptyDirectory) {
```
gvanbrakel, line 695: Please use 'Folder' instead of 'Directory'

core/src/test/java/nl/nn/adapterframework/filesystem/FileSystemActorTest.java
```diff
@@ -1167,7 +1167,39 @@ public void fileSystemActorRmdirActionTest() throws Exception {
 		// test
 		assertFalse("Expected folder [" + folder + "] " + "not to be present", actual);
 	}
+	@Test
+	public void fileSystemActorRmNonEmptyDirActionTest() throws Exception {
```
gvanbrakel, line 1171: You should also test that if removeNonEmptyDirecty=false, that an exception is thrown if the folder is not empty.

gvanbrakel (changes requested):

gvanbrakel (approved):

nielsm5 (approved):

PR #1628, opened 2021-03-30
fix #1596 update flow image dynamically

### LINKED ISSUE #1596: Flow image doesn't update on configuration reload
**Describe the bug**
When developing locally and reloading your configuration in the console so see the changes made in the configuration xml file the flow image doesn't change. When using F5 after reloading the configuration the image does update.

**Reporter**
Jaco de Groot

**Category** 
 - Frank Console

**To Reproduce**
1. Start your Frank!
2. Select an adapter to edit and look at the flow image in the console
3. Edit the xml file of the adapter, e.g. change the name of the listener and add a pipe to the pipeline
4. Reload the configuration in the console
5. See that the name of the listener is changed but the flow image doesn't show the new pipe

**Expected behavior**
The flow image to reflect the pipeline as is has been loaded after a reload of the configuration

**Screenshots**
\-

**Environment:**
 - Application server: Tomcat
 - DBMS: H2
 - Browser: Chrome
 - IAF Version: 7.7-20210317.161627

**Additional context**
\-

alisihab:


webapp/src/main/webapp/iaf/gui/js/controllers.js
```diff
@@ -269,7 +269,7 @@ angular.module('iaf.beheerconsole')
 						adapter.status = "stopped";
 
 					//Add flow diagrams
-					adapter.flow = Misc.getServerPath() + 'iaf/api/adapters/' + adapter.name + "/flow";
+					adapter.flow = Misc.getServerPath() + 'iaf/api/adapters/' + adapter.name + "/flow?"+adapter.upSince;
```
gvanbrakel, line 272: It would be better to use a version tag, so that flow diagram cache will survive a restart

gvanbrakel (changes requested):

gvanbrakel (approved):

core/src/main/java/nl/nn/adapterframework/configuration/AdapterManager.java
```diff
@@ -113,7 +114,7 @@ public void removeStopAdapterThread(Runnable runnable) {
 	}
 
 	/**
-	 * Get a registered adapter by its name through {@link IAdapterService#getAdapter(String)}
```
nielsm5, line 116: Please limit the PR to the classes you modified.

nielsm5 (changes requested):

nielsm5 (approved):

PR #1705, opened 2021-04-13
Store SOAPAction in a session key

### LINKED ISSUE #1673: WebServiceListener SOAPAction
**Describe the issue**
I want to replace an application that receives SOAP messages. This application receives all messages on the same uri and uses the SOAPAction-header to route them to the correct logic. When using the WebServiceListener I do not have access to the SOAPAction (header in SOAP 1.1, part or the Content-Type header in SOAP 1.2), blocking routing based on action

**Reporter**
Ricardo

**To Reproduce**
1. Use the below configuration
2. Use SOAPUI to send a SOAP message with the SOAPAction-header to the listener
3. Open the message in Ladybug
4. No SOAPAction is available in the session keys

**Screenshots**
N/A

**Configuration**
```
<Module>
	<Adapter name="Test">
		<Receiver name="Test">
			<WebServiceListener name="Test"
				address="test" />
		</Receiver >
		<Pipeline firstPipe="Echo">
			<exits>
				<exit path="EXIT" state="success" />
			</exits>
			<pipe
				name="Echo" 
				className="nl.nn.adapterframework.pipes.EchoPipe">
				<forward name="success" path="EXIT" />
			</pipe>
		</Pipeline>
	</Adapter>
</Module>
```

**Input**
```
[Pipeline Test.zip](https://github.com/ibissource/iaf/files/6270325/Pipeline.Test.zip)
```

**Environment**
```
FF! 7.7-20210402.153336: frank4ozb-java
Running on LPZWNL000014943 using Apache Tomcat/9.0.39
Java Version: OpenJDK Runtime Environment (1.8.0_252-b09)
Heap size: 300M, total JVM memory: 492M
Free disk space: 79GB, total disk space: 237GB
Up since: 2021-04-07 10:54:33 (2m)
```

**Additional Environment**
 - DBMS: N/A
 - Browser: N/A

**Additional Context**

alisihab:
Closes #1673

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/http/cxf/SOAPProviderBase.java
+++ b/core/src/main/java/nl/nn/adapterframework/http/cxf/SOAPProviderBase.java
@@ -1,5 +1,5 @@
 /*
-   Copyright 2018, 2019, 2021 Nationale-Nederlanden
+   Copyright 2018-2021 Nationale-Nederlanden, 2021 WeAreFrank!
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
@@ -42,6 +42,7 @@ import javax.xml.ws.WebServiceProvider;
 import javax.xml.ws.handler.MessageContext;
 
 import org.apache.commons.lang3.StringUtils;
+import org.apache.cxf.binding.soap.SoapBindingConstants;
 import org.apache.logging.log4j.Logger;
 import org.apache.soap.util.mime.ByteArrayDataSource;
 import org.w3c.dom.Element;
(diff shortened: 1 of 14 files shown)


gvanbrakel (changes requested):
Could you add a unit test or iaf-test to prove the behavior?

gvanbrakel (approved):

nielsm5 (changes requested):

core/src/main/java/nl/nn/adapterframework/http/cxf/SOAPProviderBase.java
```diff
@@ -292,4 +307,18 @@ private XmlBuilder getMimeHeadersXml(Iterator<MimeHeader> mimeHeaders) {
 		}
 		return xmlMimeHeaders;
 	}
+	
+	private String findAction(int start, String contentType) {
```
nielsm5, line 311: You could determine the start index in this method
`contentType.indexOf("action=")`

core/src/test/java/nl/nn/adapterframework/http/cxf/SoapProviderTest.java
```diff
@@ -318,4 +312,54 @@ public void receiveAndSendMessageWithAttachmentsTest() throws Throwable {
...
+	}
+
+	@Test
+	public void noSoapActionInSessionKey() throws Throwable {
+		SOAPMessage request = createMessage("soapmsg1_2.xml");
+		webServiceContext.getMessageContext().put("Content-Type", "application/soap+xml; somethingelse");
+		SOAPProvider.invoke(request);
+		webServiceContext.getMessageContext().clear();
+		assertNull(SOAPProvider.getSession().get("SOAPAction"));
+	}
```
nielsm5, line 364: All these tests could also be simplified by just testing `findAction(contentType)` with all these different contentTypes.
I would much rather see a test with soap1.1 and a soap 1.2 header, and vice-versa.
Happy flow vs unhappy flow

test/src/main/configurations/HTTP/ConfigurationWebServiceListenerSender.xml
```diff
@@ -313,4 +313,40 @@
...
 	</adapter>
+	
+	<Adapter name="HttpSoapSender" description="Send Http request to the TestSoapActionListener.">
+		<receiver className="nl.nn.adapterframework.receivers.GenericReceiver"
+			name="HttpSoapSender">
+			<listener className="nl.nn.adapterframework.receivers.JavaListener"
+				serviceName="ibis4test-HttpSoapSender" />
+		</receiver>
+		<Pipeline firstPipe="sendRequest">
+			<exits>
```
nielsm5, line 324: I think now that we have junit tests these are no longer required. Maybe its better to remove them

nielsm5 (changes requested):

nielsm5 (approved):

PR #1755, opened 2021-04-30
Cleanup Generic receiver from iaf-test configurations

alisihab:


test/src/test/testtool/Validators/Schema/SoapValidator.wsdl
```diff
@@ -1,444 +1,543 @@
...
+			<xsd:element minOccurs="0" name="TransactionList" type="TransactionListType"/>
+		</xsd:sequence>
+	</xsd:complexType>
+	<xsd:complexType name="ProducingEndpointType">
+		<xsd:sequence>
+			<xsd:element name="Id" type="xsd:string">
+				<xsd:annotation>
+					<xsd:documentation>
+						Unique identification of the initiator.
+					</xsd:documentation>
```
gvanbrakel, line 26: Could you trim() these documentations, so that there is no useless whitespace?

test/src/test/testtool/Validators/Schema/SoapValidator.wsdl
```diff
@@ -1,444 +1,543 @@
...
+				<soap:body parts="Part_Item_Request" use="literal"/>
+			</wsdl:input>
+			<wsdl:output>
+				<soap:body parts="Part_Item_Response" use="literal"/>
+			</wsdl:output>
+		</wsdl:operation>
+	</wsdl:binding>
+	<wsdl:service name="Service_SoapValidator">
+		<wsdl:port name="Port_SoapValidator" binding="tns:Binding_SoapValidator">
+			<soap:address location="external address of ibis"/>
```
gvanbrakel, line 540: This is not an improvement, i.m.o.

gvanbrakel (changes requested):

core/src/main/java/nl/nn/adapterframework/receivers/Receiver.java
```diff
@@ -541,7 +541,7 @@ public void configure() throws ConfigurationException {
 			super.configure();
 			if (StringUtils.isEmpty(getName())) {
 				if (getListener()!=null) {
-					setName(ClassUtils.nameOf(getListener()));
+					setName(Misc.concatStrings(ClassUtils.nameOf(getListener()), " ", getListener().getName()));
```
nielsm5, line 544: What does this do? This overwrites the name of the receiver? do we wan tthis? what if it's already set?

core/src/main/java/nl/nn/adapterframework/receivers/Receiver.java
```diff
@@ -1937,9 +1937,6 @@ public String toString() {
 	@IbisDoc({"10", "The source of messages"})
 	public void setListener(IListener<M> newListener) {
 		listener = newListener;
-		if (StringUtils.isEmpty(listener.getName())) {
-			listener.setName("listener of ["+getName()+"]");
-		}
```
nielsm5, line 1942: This sets the name of the listener, only when it's not explicitly set. I think we should keep this

nielsm5 (changes requested):

core/src/main/java/nl/nn/adapterframework/receivers/Receiver.java
alisihab: If the receiver does not have a name and has a listener the name of the receiver will become the class name of listener + listenerName(in case it is not empty)

core/src/main/java/nl/nn/adapterframework/receivers/Receiver.java
alisihab: this breaks the logic. we have propagateName method which updates the name of the listener in case the name of the listener is empty.
If this code block stays then the listeners may have name like 'listener of [null]' in case the receiver does not have a name

test/src/test/testtool/Validators/Schema/SoapValidator.wsdl
alisihab: I'll cheat a bit here

gvanbrakel: I see. This is just copied from the input XSD.

gvanbrakel (approved):

nielsm5 (approved):

PR #1769, opened 2021-05-03
Use static success forward

alisihab:


core/src/main/java/nl/nn/adapterframework/core/PipeForward.java
```diff
@@ -34,6 +34,7 @@
  */
 public class PipeForward {
 
+	public final static String SUCCESS_FORWARD = PipeLineExit.EXIT_STATE_SUCCESS;
```
gvanbrakel, line 37: I would suggest to call this `SUCCESS_FORWARD_NAME`

core/src/main/java/nl/nn/adapterframework/webcontrol/action/AdapterHandler.java
```diff
@@ -77,6 +78,6 @@ public ActionForward executeSub(
 			saveErrors(request, errors);
 		} // Forward control to the specified success URI
 		log.debug("forward to success");
-		return (mapping.findForward("success"));
+		return (mapping.findForward(PipeForward.SUCCESS_FORWARD));
```
gvanbrakel, line 81: All the mappings in webcontrol/action/* are not PipeForwards, but ActionForwards

gvanbrakel (changes requested):

gvanbrakel (approved):

core/src/main/java/nl/nn/adapterframework/ldap/LdapChallengePipe.java
```diff
@@ -147,7 +148,7 @@ public PipeRunResult doPipe(Message msg, PipeLineSession pls) throws PipeRunExce
 			return new PipeRunResult(findForward("invalid"), msg);
 		}
 						
-		return new PipeRunResult(findForward("success"), msg);
```
nielsm5, line 150: isn't this the getSuccessForward() ?

nielsm5 (changes requested):

core/src/main/java/nl/nn/adapterframework/ldap/LdapChallengePipe.java
gvanbrakel: no, LdapChallengPipe is not based on FixedForwardPipe (I think for no good reason)

nielsm5 (approved):

PR #1772, opened 2021-05-04
Fix #1765: Create stub receiver even if adapter has no receivers

### LINKED ISSUE #1765: Stub4testtool adapter receiver-stub copies active attribute
**Describe the issue**
When stubbing an adapter on which all receivers have an active attribute that is not locally set, the resulting adapter has no stub-receiver. In 7.5 this was not a problem as the stub created a new receiver, while in >7.6 (https://github.com/ibissource/iaf/commit/edd3ab8488dbd2ea151d9be235142b45bcf7936a) it takes those attributes from the first receiver. Because of this, the active attribute is copied, resulting in no receiver being created.

**Reporter**
Ricardo/Laura

**To Reproduce**

**Screenshots**

**Configuration**
```
```

**Input**
```
```

**Environment**
```
```

**Additional Environment**

**Additional Context**

ricardovh:


ricardovh:
Situation can happen if all receivers have an active attribute that does not resolve to true, as the active.xsl which processes the configuration first removes all those receivers. Previous versions of the stub4testtool just created a new receiver based on the adapter, which was not copied over in #1311. This PR restores that functionality.

gvanbrakel (approved):

nielsm5 (approved):

PR #1775, opened 2021-05-04
Use static variables instead of success string

### LINKED ISSUE #1701: Enum for default forwards (success, failure, exception)
**Describe the issue**
Right now we have too many `getPipeRunResult().getState().equals("success");`'s
The most common pipe forwards should be an enum or something that can easily be asserted.

**Reporter**
Niels

**Additional Context**
- state should be case in-sensitive

alisihab:
Use method to do the comparion of the state
Closes #1701


gvanbrakel (approved):

core/src/main/java/nl/nn/adapterframework/pipes/MessageSendingPipe.java
```diff
@@ -556,7 +555,7 @@ public PipeRunResult doPipe(Message input, PipeLineSession session) throws PipeR
 			if (wrapResult==null) {
 				throw new PipeRunException(inputWrapper, "retrieved null result from inputWrapper");
 			}
-			if (!wrapResult.getPipeForward().getName().equals(PipeForward.SUCCESS_FORWARD_NAME)) {
```
nielsm5, line 559: YEA! These are the big long ugly things I was referring to 😀

nielsm5 (approved):

PR #1788, opened 2021-05-07
Base groups on Javadoc annotations

mhdirkse:
This pull request aims to improve the JSON file used for the Frank!Doc website. The issue being solved is that too many groups are shown. Before this pull request, every Java interface that appeared as the argument of a config child setter introduced a group in the Frank!Doc webapplication. As an example, there was a method SenderPipe.setListener(ICorrelatedPullingListener) that introduced a group CorrelatedPullingListener. We do not want group CorrelatedPullingListener, but only Listener.

We solve this issue by introducing a JavaDoc annotation @ff.group that always has a value that is the name of the group. This is the main purpose of this pull request, but there are some related changes:

* Some method names of FrankType and FrankMethod in the wrapper API were confusing. We had for example getDefaultValueFromJavadoc(). This name was confusing, because interpreting information obtained from Java reflection or from a doclet is a task of the model. The wrapper API should only use either doclets or Java reflection to see which classes, methods, Java 5 annotations and JavaDoc annotations exist. There are also renames to have a clear distinction between Java 5 annotations and JavaDoc annotations.
* The JSON had a "group" field to tell what members were allowed as children of an element. The allowed children depend on a Java interface however that is modeled by an ElementType. The JSON referenced a FrankDocGroup however, causing a link between ElementType and FrankDocGroup that we need to break. The "group" field is removed for now, but before we merge this pull request an alternative may be introduced.

### DIFF
--- a/frankDoc/src/main/java/nl/nn/adapterframework/frankdoc/FrankDocJsonFactory.java
+++ b/frankDoc/src/main/java/nl/nn/adapterframework/frankdoc/FrankDocJsonFactory.java
@@ -195,7 +195,6 @@ public class FrankDocJsonFactory {
 		}
 		result.add("multiple", child.isAllowMultiple());
 		result.add("roleName", child.getElementRole().getRoleName());
-		result.add("group", child.getElementRole().getElementType().getFrankDocGroup().getName());
 		addIfNotNull(result, "description", child.getDescription());
 		return result.build();
 	}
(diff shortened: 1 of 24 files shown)


mhdirkse:
This pull request is doing two different things. It corrects some method names of FrankMethod because the methods should show what methods, classes and annotations are available, not their purpose. Methods like "getDefaultValueFrom..." are corrected. But this is not necessary to do the second thing, namely reducing the number of groups in the JSON. To reduce the JSON groups we need a JavaDoc annotation on classes. This is not related to JavaDoc annotations on methods.

PR #1802, opened 2021-05-12
Use configured output validator in open api schema generation

### LINKED ISSUE #1771: The  "Full API Spec" is not consistent with the list of separate api specs
**Describe the issue**
In the list of openapi specs I see 
`
/hdn/getVXMessage | GET | Pipeline has no validator. Content in the response mappings will be empty in the generated schema.
`
When downloading the "Full API Spec" the paths are there

**Reporter**
JJansenvr

**To Reproduce**
1. Go to ibis4outsystems on test 
2. webservices
3. Check the full open api spec 

**Screenshots**

**Environment**
```
FF! 7.6-RC1: ibis4outsystems 25-1863c6cc
```

**Additional Context**
I think we should generate an openapi spec even if no validator is available

alisihab:
closes #1771


core/src/main/java/nl/nn/adapterframework/http/rest/ApiServiceDispatcher.java
```diff
@@ -332,19 +335,25 @@ private JsonObjectBuilder mapResponses(IAdapter adapter, MediaTypes contentType,
 		JsonObjectBuilder responses = Json.createObjectBuilder();
 
 		PipeLine pipeline = adapter.getPipeLine();
-		Json2XmlValidator validator = getJsonValidator(pipeline);
+		Json2XmlValidator inputValidator = getJsonValidator(pipeline, false);
+		Json2XmlValidator outputValidator = getJsonValidator(pipeline, true);
+
 		JsonObjectBuilder schema = null;
 		String ref = null;
```
gvanbrakel, line 342: What does this ref refer to? Could the name be more meaningful?

gvanbrakel (changes requested):

gvanbrakel (approved):

PR #1875, opened 2021-06-02
Auto detect namedParameters in the query

### LINKED ISSUE #1826: FixedQuerySender attribute 'useNamedParams' doesn't give warning and seems unnecessary
It seems as though the 'useNamedParams' attribute is unnecessary. I think it should be easily possible to detect named parameters and handle them accordingly, instead of going off of an attribute.

Furthermore, currently, when using named params without specifying so in the attribute, the sender will throw an error that isn't very descriptive, like this:

> nl.nn.adapterframework.core.PipeRunException: Pipe [PostForm] msgId [testmessage0a722205-5cada04a_1797ea3e3ce_-7ff4] caught exception: [nl.nn.adapterframework.jdbc.FixedQuerySender] [PostForm] cannot getQueryExecutionContext: (JdbcSQLSyntaxErrorException) SQLState [42000], errorCode [42000]: Syntax error in SQL statement "INSERT INTO FORMS (NAME, DESCRIPTION, STATUS, FORMDEF) VALUES (?{name}, ?{description}, ?{status} ,?{formDef})[*]" [42000-200]

This says there is a syntax error, which is probably true, because this shouldn't normally be in a query, but it would be way more helpful and logical if the framework prevented this by throwing a warning/error on startup of the adapter.

alisihab:
Closes #1826

### DIFF
--- a/core/src/test/java/nl/nn/adapterframework/jdbc/FixedQuerySenderTest.java
+++ b/core/src/test/java/nl/nn/adapterframework/jdbc/FixedQuerySenderTest.java
@@ -9,6 +9,7 @@ import java.sql.Connection;
 import org.junit.Before;
 import org.junit.Test;
 
+import nl.nn.adapterframework.core.SenderException;
 import nl.nn.adapterframework.jdbc.dbms.IDbmsSupport;
 import nl.nn.adapterframework.parameters.Parameter;
 import nl.nn.adapterframework.senders.SenderTestBase;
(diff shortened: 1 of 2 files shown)


core/src/test/java/nl/nn/adapterframework/jdbc/FixedQuerySenderTest.java
```diff
@@ -72,71 +73,166 @@ public void setUp() throws Exception {
...
+		Parameter param = new Parameter();
+		param.setName("param");
+		param.setValue("value");
+		sender.addParameter(param);
+
+		sender.configure();
+		sender.open();
+
+		Message result = sendMessage("dummy");
+		assertEquals("?{param}", result.asString());
```
gvanbrakel, line 139: Could you remove this assertEquals and move the exception.expect to just before the sendMessage()? 
As it is now it is confusing if you don't read very carefully.

core/src/test/java/nl/nn/adapterframework/jdbc/FixedQuerySenderTest.java
```diff
@@ -72,71 +73,166 @@ public void setUp() throws Exception {
...
+	public void testUseNamedParametersWithoutParam() throws Exception {
+		exception.expect(SenderException.class);
+		exception.expectMessage("Syntax error in SQL statement \"INSERT INTO TEMP (TKEY, TVARCHAR) VALUES ('1', ?{param})[*]\"");
+		sender.setQuery("INSERT INTO TEMP (TKEY, TVARCHAR) VALUES ('1', ?{param})");
+		sender.setUseNamedParams(true);
+		sender.configure();
+		sender.open();
+
+		Message result = sendMessage("dummy");
+		assertEquals("<result><rowsupdated>1</rowsupdated></result>", result.asString());
```
gvanbrakel, line 163: Could you remove this assertEquals and move the exception.expect to just before the sendMessage()? 
As it is now it is confusing if you don't read very carefully.

core/src/test/java/nl/nn/adapterframework/jdbc/FixedQuerySenderTest.java
```diff
@@ -72,71 +73,166 @@ public void setUp() throws Exception {
...
+	public void testNamedParamInQueryFlagFalse() throws Exception {
+		exception.expect(SenderException.class);
+		exception.expectMessage("Syntax error in SQL statement \"INSERT INTO TEMP (TKEY, TVARCHAR) VALUES ('1', ?{param})[*]\"");
+		sender.setQuery("INSERT INTO TEMP (TKEY, TVARCHAR) VALUES ('1', ?{param})");
+		sender.setUseNamedParams(false);
+		sender.configure();
+		sender.open();
+
+		Message result = sendMessage("dummy");
+		assertEquals("<result><rowsupdated>1</rowsupdated></result>", result.asString());
```
gvanbrakel, line 176: Could you remove this assertEquals and move the exception.expect to just before the sendMessage()? 
As it is now it is confusing if you don't read very carefully.

core/src/test/java/nl/nn/adapterframework/jdbc/FixedQuerySenderTest.java
```diff
@@ -72,71 +73,166 @@ public void setUp() throws Exception {
...
+	@Test
+	public void testInCompleteNamedParamInQuery() throws Exception {
+		exception.expect(SenderException.class);
+		exception.expectMessage("Syntax error in SQL statement \"INSERT INTO TEMP (TKEY, TVARCHAR) VALUES ('1', ?{param)[*]\"");
+		sender.setQuery("INSERT INTO TEMP (TKEY, TVARCHAR) VALUES ('1', ?{param)");
+		sender.configure();
+		sender.open();
+
+		Message result = sendMessage("dummy");
+		assertEquals("<result><rowsupdated>1</rowsupdated></result>", result.asString());
```
gvanbrakel, line 188: Could you remove this assertEquals and move the exception.expect to just before the sendMessage()? 
As it is now it is confusing if you don't read very carefully.

gvanbrakel (changes requested):

nielsm5 (approved):

gvanbrakel (approved):

PR #1906, opened 2021-06-14
Implement not operator also for properties other than "active"

mhdirkse:
Fix issue https://github.com/ibissource/iaf/issues/1898. The not operator was only implemented for property "active". This pull request implements it for all Boolean attributes. This makes the old implementation through active.xsl superfluous. This pull request will also include the required update to the Frank!Doc.

The old implementation is undone by partially undoing the following commit: 7e61fb5671a8577f8e84afddbbc2686a9ebb6cda "Simplify active.xsl (#1394)" by ricardovh on Wed Jan 20 12:22:20 2021 +0100. Only the changes to active.xsl are undone. The commit also had a small update to test code that was changed again after the mentioned commit; therefore that change of the commit was not undone.


mhdirkse:
Right now, this pull request is not ready. The following has to be added:

* When a boolean attribute is set to a non-boolean value, then there should be a console warning. This is not the case at the time of writing. This behavior is not yet in this pull request but will be added.
* The Frank!Doc has to be updated to accept Boolean attribute values with an !.
* The Frank!Doc has to be updated to not only accept "true" and "false" for Boolean attribute. The value should be case-insensitive.

core/src/main/java/nl/nn/adapterframework/configuration/digester/ValidateAttributeRule.java
```diff
@@ -34,6 +38,18 @@
...
+
+	private static Map<String, String> BOOLEAN_NEGATIONS;
+	static {
+		BOOLEAN_NEGATIONS = new HashMap<>();
+		for(Boolean b: Arrays.asList(Boolean.FALSE, Boolean.TRUE)) {
+			String s = b.toString();
+			BOOLEAN_NEGATIONS.put(s, s);
+			BOOLEAN_NEGATIONS.put("!" + s, Boolean.valueOf(! b).toString());
+		}
+	}
```
nielsm5, line 51: Je bent nu nieuwe functionaliteit aan het inbouwen. Is dit wel de bedoeling?

mhdirkse:
This pull request changes the way property values are read when configurations are parsed. There are other places in the code that read properties, however. As an example, properties can also be read from property files. To implement the not operator, all occasions of reading properties have to be updated. We do not want to do that, so we close this pull request. The attribute "active" then remains the only attribute for which the !-operator is supported.

PR #1929, opened 2021-06-21
Fix #1882 JsonPipe default version 2

### LINKED ISSUE #1882: JsonPipe evaluates empty element to json empty object
**Describe the issue**
When transforming an XML to a JSON format <a></a> ==> a: {} instead of a: ""

**Reporter**
jjansenvr / Swapil

**To Reproduce**
Implement a JSONPipe provide it with a xml with empty values.
 

**Screenshots**

**Configuration**
```
<pipe name="ConvertJson2Xml"
                  className="nl.nn.adapterframework.pipes.JsonPipe">
	           <forward name="success" path="verifyTitanResponse"/>
</pipe>
```

**Input**
```
<foo>
<bar>value</bar>
<bar1></bar1>
<bar2></bar2>
<bar3></bar3>
</foo>
```

**Environment**
```
IAF 7.6-20200421.151810: ibis4webforms test
```

**Additional Environment**
na

**Additional Context**
Workaround is to implement a Json validator instead, using a schema will give more control over output

The code of the JSONPipe needs to be refactored, it uses the wrong libraries

gvanbrakel:


nielsm5:
This is not a fix but a temporary solution. I think we should remove the `org.json` dependency and make this pipe work without the use of a transformer (and thus removing the version nonsense).

nielsm5 (approved):

PR #1984, opened 2021-07-05
Frank!Doc: Update compatibility.xsd to ignore the sequence of child elements

mhdirkse:


mhdirkse:
This commit changes the following:

* compatibility.xsd, the XSD used by the Frank!Framework to parse configs, should not check the sequence of elements. We want this because old configurations do not respect the element sequence prescribed by strict.xsd.
* Both strict.xsd and compatibility.xsd had empty sequences. These do not have any purpose so they are removed.
* More integration tests are added to DocWriterNewIntegrationTest to improve code coverage.

The new integration tests introduce input class hierarchies and expected XSD documents. The expected XSD documents should be produced by the Frank!Doc when the input class hierarchies are processed. It was difficult to write the correct expected values for these XSD documents. I checked each expected XSD by using it to verify and example XML document. I had an example XML document open in Eclipse that referenced an XSD. Then I checked the errors I saw while editing the XML.

mhdirkse:
I put this pull request back to "draft" because I want more test coverage. Here are some ideas about the coverage I want to add:

* When there are role names with multiple config children, I want to test that all config children go under the same <sequence><choice>, not just the config children of the role. I can test this by adding more config children.
* I want to test that attributes still appear at the right place. Therefore I have to add attributes.

mhdirkse:
Ik heb de twee XSDs die ik in de laatste commit heb aangepast weer tegen een XML getest.

gvanbrakel (approved):

nielsm5 (approved):

PR #1992, opened 2021-07-07
Ensure that the pipeline fails when Frank!Doc cannot be build

nielsm5:


webapp/src/main/webapp/iaf/frankdoc/views/element-attributes.html
```diff
@@ -1,5 +1,6 @@
 <div>
 	<h4>Attributes</h4>
+	<p>It is possible to configure the following attributes:</p>
```
gvanbrakel, line 3: Ik zou zeggen: The following attributes can be configured

webapp/src/main/webapp/iaf/frankdoc/views/element-children.html
```diff
@@ -1,9 +1,12 @@
 <div ng-controller="element-children">
 	<h4>Nested children</h4>
-	<div ng-repeat="child in element.children">
-		<div>
-			<b ng-attr-title="{{ getTitle(child) }}">{{child.roleName}}</b>
-			<span ng-if="child.description"><b>:</b> {{ child.description }}</span>
-		</div>
-	</div>
+	<p>It is possible to configure the following sub-elements:</p>
```
gvanbrakel, line 3: Ik zou zeggen: The following sub-elements can be configured

gvanbrakel (changes requested):

gvanbrakel (approved):

PR #2003, opened 2021-07-08
Fix WebServiceListener and WebServiceSender related warnings in config HTTP

Bibi0803:
1 - WebServiceListener calling webservices via de ServiceDispatcher_ServiceProxy is deprecated. Please specify an address or serviceNamespaceURI and modify the call accordingly
2 - WebServiceSender attribute [multipartResponse] is deprecated: Unless set explicitly multipart response will be detected automatically

### DIFF
--- a/test/src/main/configurations/HTTP/ConfigurationWebServiceListenerSender.xml
+++ b/test/src/main/configurations/HTTP/ConfigurationWebServiceListenerSender.xml
@@ -77,7 +77,7 @@
 			<pipe name="Send2WS"
 				className="nl.nn.adapterframework.pipes.SenderPipe">
 				<sender className="nl.nn.adapterframework.http.WebServiceSender"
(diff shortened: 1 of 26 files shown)


test/src/main/configurations/HTTP/ConfigurationWebServiceListenerSender.xml
```diff
@@ -2,7 +2,7 @@
 	<adapter name="WebServiceListener" description="Test the functioning of the WebServiceListener">
 		<receiver>
 			<listener className="nl.nn.adapterframework.http.WebServiceListener"
-				name="urn:ws" />
+				name="urn:ws" serviceNamespaceURI="urn:ws" />
```
nielsm5, line 5: Could you revert the changes made in this file, as we explicitly test the endpoints here. The other 2 files are ok!

nielsm5 (approved):

gvanbrakel (approved):

PR #2025, opened 2021-07-15
Fix warnign "WsdlXmlValidator [ndn] attribute [schemaLocation] for wsdl [/WsdlXmlValidator/NoDefaultNamespace/NoDefaultNamespace.wsdl] should only be set when addNamespaceToSchema=true"

Bibi0803:


test/src/main/configurations/MainConfig/ConfigurationWsdlXmlValidator.xml
```diff
@@ -33,8 +33,7 @@
 			<!-- ndn (NoDefaultNamespace) -->
 			<pipe name="ndn" className="nl.nn.adapterframework.pipes.WsdlXmlValidator"
 				soapBody="Request"
-				wsdl="/WsdlXmlValidator/NoDefaultNamespace/NoDefaultNamespace.wsdl"
-				schemaLocation="dummy schema1">
+				wsdl="/WsdlXmlValidator/NoDefaultNamespace/NoDefaultNamespace.wsdl">
```
gvanbrakel, line 36: The original configuration also tests that the operation does not break when the schemaLocation has been set, so if you remove that here, that test is lost.
If you still want to get rid of the warning here, you should extend the unit test you created to see that a pipe confgured as the original works as expected

gvanbrakel (changes requested):

gvanbrakel (approved):

nielsm5 (approved):

PR #2150, opened 2021-08-25
Convert file system actions to enum

alisihab:


core/src/main/java/nl/nn/adapterframework/filesystem/FileSystemActor.java
```diff
@@ -558,16 +593,19 @@ public MessageOutputStream provideOutputStream(PipeLineSession session, IForward
...
 
-	@IbisDoc({"1", "Possible values: "+ACTION_LIST+", "+ACTION_INFO+", "+ACTION_READ1+", "+ACTION_READ_DELETE+", "+ACTION_MOVE+", "+ACTION_COPY+", "+ACTION_DELETE+", "+ACTION_MKDIR+", "+ACTION_RMDIR+", "+ACTION_WRITE1+", "+ACTION_APPEND+", "+ACTION_RENAME+". If parameter ["+PARAMETER_ACTION+"] is set, then the attribute action value will be overridden with the value of the parameter.", "" })
+	@IbisDoc({"1", "If parameter ["+PARAMETER_ACTION+"] is set, then the attribute action value will be overridden with the value of the parameter.", "" })
 	public void setAction(String action) {
-		this.action = action;
+		this.action = EnumUtils.parse(FileSystemAction.class, action);
+	}
+	public FileSystemAction getActionEnum() {
+		return action;
 	}
```
gvanbrakel, line 606: Nowadays, you can use `public void setAction(FileSystemAction action)` and `public FileSystemAction getAction()`. The digester interprets the Enum, and the FrankDoc understands it too.

core/src/main/java/nl/nn/adapterframework/filesystem/FileSystemActor.java
```diff
@@ -198,38 +218,38 @@ public void configure(FS fileSystem, ParameterList parameterList, IConfigurable
...
 			if (!parameter1Set && !parameter2Set && !attributeSet) {
-				throw new ConfigurationException(ClassUtils.nameOf(owner)+" ["+owner.getName()+"]: the "+action+" action requires the parameter ["+parameter1+"] "+(parameter2!=null?"or parameter ["+parameter2+"] ":"")+(attributeName!=null?"or the attribute ["+attributeName+"] ": "")+"to be present");
+				throw new ConfigurationException(ClassUtils.nameOf(owner)+" ["+owner.getName()+"]: the ["+action+"] action requires the parameter ["+parameter1+"] "+(parameter2!=null?"or parameter ["+parameter2+"] ":"")+(attributeName!=null?"or the attribute ["+attributeName+"] ": "")+"to be present");
 			}
 		}
 	}
 
 	public void open() throws FileSystemException {
-		if (StringUtils.isNotEmpty(getInputFolder()) && !fileSystem.folderExists(getInputFolder()) && !ACTION_MKDIR.equals(getAction())) {
+		if (StringUtils.isNotEmpty(getInputFolder()) && !fileSystem.folderExists(getInputFolder()) && FileSystemAction.MKDIR!=getActionEnum()) {
```
gvanbrakel, line 252: I prefer to see `FileSystemAction.MKDIR!=getActionEnum()` like `getActionEnum() != FileSystemAction.MKDIR`,
as this is the more natural order.
The order the other way round is only to avoid NPEs for Strings that are null

core/src/main/java/nl/nn/adapterframework/filesystem/FileSystemActor.java
```diff
@@ -577,7 +615,7 @@ public void setCreateFolder(boolean createFolder) {
 		this.createFolder = createFolder;
 	}
 
-	@IbisDoc({"4", "when set to <code>true</code>, for actions "+ACTION_MOVE+", "+ACTION_COPY+" or "+ACTION_RENAME+", the destination file is overwritten if it already exists", "false"})
+	@IbisDoc({"4", "when set to <code>true</code>, for actions MOVE, COPY or RENAME, the destination file is overwritten if it already exists", "false"})
```
gvanbrakel, line 618: Would it be possible to use the refer to the values of the Enum constants here? E.g. by making them final?

gvanbrakel (changes requested):

core/src/main/java/nl/nn/adapterframework/filesystem/FileSystemUtils.java
```diff
@@ -97,8 +98,8 @@
 	}
 
 	public static <F> F moveFile(IBasicFileSystem<F> fileSystem, F file, String destinationFolder, boolean overwrite, int numOfBackups, boolean createFolders) throws FileSystemException {
-		checkSource(fileSystem, file, "move");
-		prepareDestination(fileSystem, file, destinationFolder, overwrite, numOfBackups, createFolders, "move");
+		checkSource(fileSystem, file, FileSystemAction.MOVE.getLabel());
```
nielsm5, line 101: in checkSource you're just validating if the action is allowed (if I'm not mistaken) you could do `xx.getLabel()` there.
That way on line 91 you'd also pass the enum, instead of a String

nielsm5 (changes requested):

PR #2163, opened 2021-08-30
Do not duplicate the stack trace in the log file

### LINKED ISSUE #2142: Deduplicate Log4j2 PatternLayout stacktrace
**Describe the issue**
The log4j2 stacktrace is extremely long and needs to be shortened.
See https://stackoverflow.com/questions/31384939/log4j2-including-library-name-in-stacktrace, https://logging.apache.org/log4j/2.x/manual/layouts.html#PatternLayout

alisihab:
Closes #2142

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/logging/IbisThrowablePatternConverter.java
+++ b/core/src/main/java/nl/nn/adapterframework/logging/IbisThrowablePatternConverter.java
@@ -1,0 +1,117 @@
+package nl.nn.adapterframework.logging;
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.net.URL;
+import java.security.CodeSource;
+
+import org.apache.logging.log4j.core.LogEvent;
+import org.apache.logging.log4j.core.config.Configuration;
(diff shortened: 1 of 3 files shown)


gvanbrakel (commented):
Does this have no impact on iaf-test?

alisihab:
> Does this have no impact on iaf-test?

ran the scenarios, none of them requires adjusting. This may not be the expected solution. You may see the the difference that came with the change in the attached files.
[before.txt](https://github.com/ibissource/iaf/files/7078119/before.txt)
[after.txt](https://github.com/ibissource/iaf/files/7078120/after.txt)

gvanbrakel (commented):
The javadoc of 'org.apache.logging.log4j.core.impl.ThrowableProxy' says: "Wraps a Throwable to add packaging information about each stack trace element." It also has a 'commonElementCount'. Doesn't this class do all that is necessary?

nielsm5 (changes requested):

gvanbrakel (approved):

nielsm5 (approved):

PR #2171, opened 2021-08-31
Frank!Doc: User manual for core developers

mhdirkse:


mhdirkse:
Gerrit, I saw your email. File FRANKDOC.md should be rewritten. The document now explains the behavior of the doclet code, but it does not provide the information to core developers as they need it. I made a start rewriting the text. I will add a note when I think I am done. I will only make this issue ready-for-review if some other PRs will have been merged. We cannot merge documentation that is not true yet.

mhdirkse:
I want to close this pull request without losing the work I did. In the
future I may rewrite some of the text. We need a guide on how to develop
the Frank!Framework. That is the place to explain how Java classes and
Java methods should be organized and named. That explanation should not
be combined with explaining the Frank!Doc like I was trying in this PR.

Issue #1390, opened 2021-01-12
CheckReload executed even when no config has DatabaseClassLoader


ricardovh:
**Reporter**
Ricardo

**Is your feature request related to a problem? Please describe.**
When having an Frank that has no configurations loaded using a DatabaseClassLoader, the CheckReload is still executed every minute and queries the database for active configurations. This might lead to errors if there is no database/no ibisconfig table (as is currently happening on https://ibis4example.ibissource.org/iaf/gui/#!/scheduler). 

**Describe the solution you'd like**
A check to be executed before the query to check if there are configurations for which a reload might make sense.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

nielsm5:
Solved in https://github.com/ibissource/iaf/pull/2726

Issue #1220, opened 2020-10-30
Use JdbcTableListener adapter instead of scheduled job for configuration reload


gvanbrakel:
**Reporter**
Gerrit

**Is your feature request related to a problem? Please describe.**
The log files of ibises that have configuration reload enabled are cluttered with logging of reload attempts, that are in most cases void. 

**Describe the solution you'd like**
Create an Adapter configuration with a JdbcTableListener, that polls the database for configurations to load, and performs the reload. 

**Describe alternatives you've considered**
Switching off the reload by default

**Additional context**

jacodg:
Listener as trigger for job execution to make it more generic

PR #2197, opened 2021-09-13
reduce iaf-test database reload interval to once per hour

gvanbrakel:


nielsm5 (approved):

nielsm5:
Temporarily fix for #1390 and #1220

Issue #2206, opened 2021-09-16
Missing index on IBISSTORE / Message store sender very slow when 100k+ records


Laurens-makel:
**Describe the issue**
On TST/ACC we had perfect performance of storing archive requests in the IBISSTORE using a Message Store Sender, averaging under 100 ms.

Although on production we saw averages of ~35 seconds, after some investigation we found that acceptance only has 1000 records and on production around 100.000 records.

On line number 406 an expensive query is executed to detect duplicate message id's.
https://github.com/ibissource/iaf/blob/master/core/src/main/java/nl/nn/adapterframework/jdbc/JdbcTransactionalStorage.java

Since there is no index on (slotid, messageid) it is a very costly query. 

(4th column is average)

Before applying index:
`Store_ArchiveRequests			2819	21	886938	32011.9	38081.1	90241620	6026	21	0%	0%	0%	32.8%	19928.2	162309.2	228526.4	268256.8`

After applying index:
`Store_ArchiveRequests | 96 | 9 | 392 | 20.6 | 51 | 1982 | 30 | 25`

Suggestion is to add an index on slotid, messageid to improve the performance.

**Reporter**
Laurens

**To Reproduce**

**Screenshots**

**Configuration**
```
Please provide the configuration of the Pipe or Receiver with the problem.
```

**Input**
```
Please provide an example of the input message. 
Alternatively, provide a Ladybug report.
```

**Environment**
```
FF! 7.7-20210525.083818: Ibis4XXXX
Running on srpzzapp0531.XXX.XXXXusing IBM WebSphere Application Server/8.5
Java Version: Java(TM) SE Runtime Environment (8.0.6.25 - pxa6480sr6fp25-20210115_01(SR6 FP25))
Heap size: 850M, total JVM memory: 4096M
Free disk space: 7GB, total disk space: 10GB
Up since: 2021-09-16 07:20:00 (3h)
```

**Additional Environment**
 - DBMS: [e.g. Oracle, MSSQL, MariaDB, MySql, PostgreSQL]
 - Browser: [e.g. Chrome, Safari, Edge, Firefox]

**Additional Context**
Add any other context about the problem here. (f.e. ladybug report / test adapter with larva test)

PR #2207, opened 2021-09-17
Impl #2206 add index for duplicates check

gvanbrakel:


ricardovh:
Shouldn't https://github.com/ibissource/iaf/blob/master/core/src/main/java/nl/nn/adapterframework/jdbc/JdbcTransactionalStorage.java#L249 also be updated to give a warning when the new index is missing?

nielsm5 (approved):

PR #2217, opened 2021-09-20
Make larva default timeout configurable

### LINKED ISSUE #1774: Create a property to set Larva timeout
**Describe the issue**
After a codechange, a lot of time is wasted adjusting the Larva results.
Each timeout in larva takes 30 seconds, which quickly add up.

I would like to have a property added which i can locally use to set the timeout length.

This can be done by adding a property for DEFAULT_TIMEOUT in TestTool.java

**Reporter**
Bonno Smeele

NA

**Screenshots**

**Configuration**
```
NA, can be any configuration
```

**Input**
```
NA, can be any input
```

**Environment**
```
Local
```

**Additional Environment**
 - DBMS:  NA
 - Browser: NA

**Additional Context**

alisihab:
Closes #1774

### DIFF
--- a/larva/src/main/java/nl/nn/adapterframework/testtool/TestTool.java
+++ b/larva/src/main/java/nl/nn/adapterframework/testtool/TestTool.java
@@ -110,7 +110,7 @@ public class TestTool {
 	private static Logger logger = LogUtil.getLogger(TestTool.class);
 	public static final String LOG_LEVEL_ORDER = "[debug], [pipeline messages prepared for diff], [pipeline messages], [wrong pipeline messages prepared for diff], [wrong pipeline messages], [step passed/failed], [scenario passed/failed], [scenario failed], [totals], [error]";
 	private static final String STEP_SYNCHRONIZER = "Step synchronizer";
-	protected static final int DEFAULT_TIMEOUT = 30000;
+	protected static final int DEFAULT_TIMEOUT = AppConstants.getInstance().getInt("larva.timeout", 30000);
 	protected static final String TESTTOOL_CORRELATIONID = "Test Tool correlation id";
 	protected static final String TESTTOOL_BIFNAME = "Test Tool bif name";
 	protected static final nl.nn.adapterframework.stream.Message TESTTOOL_DUMMY_MESSAGE = new nl.nn.adapterframework.stream.Message("<TestTool>Dummy message</TestTool>");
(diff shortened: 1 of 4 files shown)


test/src/main/resources/DeploymentSpecifics.properties
```diff
@@ -71,7 +71,8 @@ testdata.dir=${log.dir}/testdata
 fxf.dir=${testdata.dir}/fxf
 
 # Please mark that the backslashes below are double escaped: first for the value of the property, second time for use in regular expression
-log.dir.match=(?i)([cd]:[\\\\/]temp)|(${log.dir})
+
+log.dir.match=(?i)(([cd]:[\\\\/]temp)|(${log.dir}))
```
gvanbrakel, line 75: This has already been fixed in PR #2218.

gvanbrakel (changes requested):
Wouldn't it be useful/possible to be able to specify a timeout in a scenario step?

test/src/main/resources/DeploymentSpecifics.properties
gvanbrakel: Please revert the changes in log.dir.match and testdata.dir.match, merge master and test again.

gvanbrakel (changes requested):

gvanbrakel (approved):

larva/src/main/java/nl/nn/adapterframework/testtool/TestTool.java
```diff
@@ -158,28 +158,33 @@ public static void runScenarios(ServletContext application, HttpServletRequest r
...
 	 * 
-	 * @param paramScenarioTimeout 
 	 * @return negative: error condition
 	 * 		   0: all scenarios passed
 	 * 		   positive: number of scenarios that failed
 	 */
 	public static int runScenarios(IbisContext ibisContext, AppConstants appConstants, String paramLogLevel,
-			String paramAutoScroll, String paramExecute, String paramWaitBeforeCleanUp, String paramScenarioTimeout,
+			String paramAutoScroll, String paramExecute, String paramWaitBeforeCleanUp,
 			String realPath, String paramScenariosRootDirectory,
```
nielsm5: I think you misunderstood me when I mentioned this should be an Integer. Right now you're updating a static field which will remember the setting across multiple sessions. I believe the idea is that you can;
- Change the default with a property
- Override the default (once) with a setting in the GUI

nielsm5 (changes requested):

PR #2244, opened 2021-09-23
Fix copyright exclude filepath when running single maven module

mhdirkse:


mhdirkse:
I cannot build the master anymore. This fixes the issue for me

nielsm5:
License is already included in the file, perhaps you could remove the slash in front of the text instead.

nielsm5 (changes requested):
License is already included in the file, perhaps you could remove the slash in front of the text instead.

nielsm5 (approved):

pom.xml
```diff
@@ -706,5 +706,16 @@
 				</plugins>
 			</reporting>
 		</profile>
+		<profile>
+			<activation>
+				<file>
+					<exists>copyrightTemplateFile.txt</exists>
```
gvanbrakel, line 712: Toch een beetje raar om aan de hand van deze file te bepalen of het de root folder is.

gvanbrakel (approved):

PR #2291, opened 2021-10-05
Frank!Doc: Demonstrate that initial field value is not available for attributes

mhdirkse:
In Java, you can give an initial value within a field declaration. It would be nice if the Frank!Doc could grab this initial value. Then we would not need to duplicate that value in an ff.default JavaDoc tag. Alas, this is not possible. This PR demonstrates this. If we merge it, we will remember.


nielsm5:
Behoorlijk frustrerend dat `getConstantValue()` anders werkt dan gehoopt.

PR #2358, opened 2021-10-18
Add jdk 11 support

### LINKED ISSUE #1998: Add JDK11+ compatibility
**Describe the issue**
Some classes are not compatible with newer JDK versions.
Create a wrapper that tries the old, then new version then throws exception if it cannot find the class.

**Reporter**
Core-Team

alisihab:
Closes #1998


test/pom.xml
```diff
@@ -35,7 +35,7 @@
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-compiler-plugin</artifactId>
-				<version>3.1</version>
+				<version>3.8.1</version>
```
nielsm5, line 38: Can this be managed via pluginManagement?

test/pom.xml
```diff
@@ -115,7 +115,7 @@
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-war-plugin</artifactId>
-				<version>2.6</version>
+				<version>3.2.2</version>
```
nielsm5, line 118: Can this be managed via pluginManagement?

test/pom.xml
```diff
@@ -194,5 +194,23 @@
...
+						<groupId>org.apache.maven.plugins</groupId>
+						<artifactId>maven-compiler-plugin</artifactId>
+						<version>3.8.1</version>
+						<configuration>
+							<release>11</release>
+						</configuration>
+					</plugin>
+				</plugins>
+			</build>
+		</profile>
```
nielsm5, line 214: I would expect this in the parent pom, and other poms to inherit it from there. Does that not work?

ibm/pom.xml
```diff
@@ -1,65 +1,94 @@
...
+			<activation>
+				<jdk>11</jdk>
+			</activation>
+			<dependencies>
+				<dependency>
+					<groupId>com.ibm</groupId>
+					<artifactId>ws-runtime</artifactId>
+					<version>8.0.0</version>
+					<scope>provided</scope>
+				</dependency>
```
nielsm5, line 78: This doesn't have to be in a profile. You can upgrade the old was6 dependency

core/pom.xml
```diff
@@ -680,5 +680,75 @@
...
+					<version>${cxf.version}</version>
+					<exclusions>
+						<exclusion>
+							<groupId>com.sun.xml.messaging.saaj</groupId>
+							<artifactId>saaj-impl</artifactId>
+						</exclusion>
+						<exclusion>
+							<groupId>org.apache.geronimo.specs</groupId>
+							<artifactId>geronimo-jta_1.1_spec</artifactId>
+						</exclusion>
```
nielsm5, line 736: Is it possible to exclude these by default? Do we need the JTA spec compile time?

nielsm5 (changes requested):

nielsm5 (approved):

gvanbrakel (approved):

PR #2434, opened 2021-11-11
Add ability to upload a file to liquibase script generator

### LINKED ISSUE #1484: Allow to upload/edit databasechangeset in Liquibase script generator
**Reporter**
Gerrit

It would be nice to have the ability to view/edit/upload a liquibase databasechangeset to the Liquibase script generator. This would help application updates, as it would allow to do/prepare the database changes independent of the actual deployment.
It would also allow for a workaround for the problem that Liquibase checksums can change over Liquibase versions.

alisihab:
Closes #1484


webapp/src/main/webapp/iaf/gui/js/controllers.js
```diff
@@ -1782,7 +1749,7 @@ angular.module('iaf.beheerconsole')
 
 .controller('AddScheduleCtrl', ['$scope', 'Api', 'Misc', function($scope, Api, Misc) {
 	$scope.state = [];
-	$scope.addLocalAlert = function(type, message) {
```
nielsm5, line 1785: I think a merge went wrong here..

nielsm5 (changes requested):

core/src/main/java/nl/nn/adapterframework/jdbc/migration/LiquibaseLogger.java
```diff
@@ -0,0 +1,103 @@
...
+	public void setLogLevel(String logLevel, String logFile) {
+	}
+
+	@Override
+	public void setChangeLog(DatabaseChangeLog databaseChangeLog) {
+	}
+
+	@Override
+	public void setChangeSet(ChangeSet changeSet) {
+	}
```
nielsm5, line 98: Are you purposely overriding these? If so, please give a brief explanation as to why. Else please remove these methods, as I can imagine it's handy to know which file/changeset contains an error.

core/src/main/java/nl/nn/adapterframework/jdbc/migration/LiquibaseLogger.java
```diff
@@ -0,0 +1,103 @@
...
+	@Override
+	public void debug(String message, Throwable e) {
+		if(log.isDebugEnabled()) {
+			log.debug(message, e);
+		}
+	}
+
+	@Override
+	public void setLogLevel(String logLevel, String logFile) {
+	}
```
nielsm5, line 90: I can imagine when you delegate everything to log4j you either want to let log4j handle this (and set the loglevel default to INFO, as per the old log4j4ibis.xml).

core/src/main/java/nl/nn/adapterframework/jdbc/migration/LiquibaseLogger.java
```diff
@@ -0,0 +1,103 @@
...
+		if(log.isTraceEnabled()) {// prints too much unnecessary only print if trace enabled
+			log.debug(message);
+		}
+	}
+
+	@Override
+	public void debug(String message, Throwable e) {
+		if(log.isDebugEnabled()) {
+			log.debug(message, e);
+		}
```
nielsm5, line 85: You don't need these if statements, it's the first thing log4j does :)

core/src/main/java/nl/nn/adapterframework/jdbc/migration/LiquibaseLogger.java
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * @author alisihab
+ *
+ */
+public class LiquibaseLogger extends AbstractLogger {
+	private static final Logger log = LogUtil.getLogger(LiquibaseLogger.class);
+
+	@Override
+	public void setName(String name) {
+	}
```
nielsm5, line 34: Give a brief explanation that calling this wont do much as the loggers name has already been defined upon creating the logger

core/src/main/java/nl/nn/adapterframework/jdbc/migration/LiquibaseLogger.java
```diff
@@ -0,0 +1,103 @@
...
+
+	@Override
+	public void severe(String message) {
+	}
```
nielsm5: I think these should also be logged?

core/src/main/java/nl/nn/adapterframework/jdbc/migration/LiquibaseImpl.java
```diff
@@ -71,6 +72,16 @@ public LiquibaseImpl(DataSource datasource, Configuration configuration, String
 		}
 		connection = new JdbcConnection(datasource.getConnection());
 
+		// Override liquibase internal logger
+		LogFactory.setInstance(new LogFactory() {
```
nielsm5, line 76: Please override getLog(String) instead to be able to fetch the correct logger.

core/src/main/java/nl/nn/adapterframework/jdbc/migration/LiquibaseLogger.java
```diff
@@ -0,0 +1,103 @@
...
+import nl.nn.adapterframework.util.LogUtil;
+
+import org.apache.logging.log4j.Logger;
+
+/**
+ * @author alisihab
+ *
+ */
+public class LiquibaseLogger extends AbstractLogger {
+	private static final Logger log = LogUtil.getLogger(LiquibaseLogger.class);
```
nielsm5, line 30: Please give each logger a name.
Right now all log statements come from the migration.LiquibaseLogger package, which hides it's origin.
You can set the final field in a constructor where you may pass the name of the logger.

nielsm5 (changes requested):

nielsm5 (approved):

gvanbrakel (approved):

PR #2446, opened 2021-11-15
Compare Frank!Doc with previous version

mhdirkse:


mhdirkse:
As this PR is now, the build can fail if the generated Frank!Doc has changed. This happens when text is removed. I think it is better to have the build succeed regardless the changes in the Frank!Doc. Shall I update this PR for that? We can still produce a diff file with the changes.

nielsm5:
Yes, the build should continue

mhdirkse:
The diffs are produced successfully now, but I think they are truncated in the output of the GitHub action. Shall I write the diffs to files and make them available as artifacts?

mhdirkse:
I saw in the git log that this PR has been superseeded. Shall I close this PR?

nielsm5:
> I saw in the git log that this PR has been superseeded. Shall I close this PR?

Yup, closing this.

PR #2452, opened 2021-11-16
Document that FrankDocGroup behaves differently on classes and interfaces

mhdirkse:


FRANKDOC.md
```diff
@@ -18,13 +19,11 @@ The Frank!Doc provides reference information for Frank developers as explained i
...
 
-**Annotation @FrankDocGroup:**: Has fields `name` and `order` (integer), which is used by the webapp to sort the groups. You can define a group by setting multiple `@FrankDocGroup` annotations with the same `name`. Only one `@FrankDocGroup` annotation of a group should have its `order` field set to avoid ambiguity.
+**Annotation @FrankDocGroup:**: Has fields `name` and `order` (integer), which is used by the webapp to sort the groups. You can define a group by setting multiple `@FrankDocGroup` annotations with the same `name`. Only one `@FrankDocGroup` annotation of a group should have its `order` field set to avoid ambiguity. This annotation behaves differently on classes and interfaces. When a Java class implements an interface that has a `@FrankDocGroup` annotation, then the class is put in the group non-exclusively. A class can belong to multiple groups when it implements multiple interfaces with different groups. However, when a class has a `@FrankDocGroup` annotation, then the class only belongs to that group. `@FrankDocGroup` annotations on classes are inherited by descendant classes, so descendant classes are also in the specified group exclusively.
```
mhdirkse, line 23: It is an error to put a @FrankDocGroup annotation on a class that is not compatible with any interface-inherited @FrankDocGroup. @FrankDocGroup-s on classes can serve two purposes:
* For a non-interface config child, put its member in a group. An example is the `@FrankDocGroup` on class `InputfieldsPart`.
* Restrict a member of an interface-based config child to only one group. It is an error to choose a group that is not inherited by any interface.

Shall I update this PR to add this, or is it too much detail?

gvanbrakel: Please do not introduce breaking tests for situations that cannot be identified as a real threat to something.

mhdirkse: This test is there for a reason. If you use a @FrankDocGroup to restrict the type of a class, you can mistype the group name.  Without a test, a new group would be created. The new group would be empty, because it would not possess types. The annotated class would be excluded from all inherited types. It would become unreachable as the member of a config child. This would apply only for the web application, not the XSD.

gvanbrakel: In my opinion, it should not be a problem to set the @FrankDocGroup annotation to a new or other group.
For instance, it should not be a problem to set `@FrankDocGroup(name = "Batch")` to StreamTransformerPipe. If that was done, StreamTransformerPipe and it's descendant BatchFileTransformerPipe should just appear in the 'Batch' group.
It should not harm the ability to use those elements as pipes in a configuration.

mhdirkse: This feature is complicated because @FrankDocGroup annotations have a relation with config children. The webapp shows for a config child a list of the possible child classes. Such a list always corresponds to an ElementType in the Frank!Doc model. The JSON has an entry "types" that contains all types along with their members. The members of a "type" are the members of an ElementType.

The @ff.ignoreTypeMembership annotation we are removing could be put on a class to remove it from a type. Then it was both removed from the related group and it was removed as a possible config child. I implemented the same using the @FrankDocGroup annotation on a class. So the @FrankDocGroup annotation on MessageStoreSender removes that class from webapp group TransactionalStorages. It also removes MessageStoreSender from the possibilities for config child SenderPipe | messageLog, to name an example.

Considering this confusion, I think we should re-introduce @ff.ignoreTypeMembership and remove the new feature of @FrankDocGroup. Then the annotations and tags explain themselves better.

mhdirkse: Gerrit, we continued this discussion orally. The conclusion was that annotation @FrankDocGroup should remain the same. The story that links groups, element types and the limitation of @FrankDocGroup we are hitting here is too complicated to document in the [FRANKDOC.md](./FRANKDOC.md) file or in the README file of the frank-doc project. I updated a log message in frank-doc PR https://github.com/ibissource/frank-doc/pull/33. I hope that is enough to document the issue.

FRANKDOC.md
```diff
@@ -18,13 +19,11 @@ The Frank!Doc provides reference information for Frank developers as explained i
...
```
gvanbrakel, line 22: Ik denk dat het overzichtelijker is om dit te splitsen in een '@FrankDocGroup on a interface' en een '@FrankDocGroup on a class'

gvanbrakel (changes requested):

FRANKDOC.md
```diff
@@ -5,7 +5,8 @@ The Frank!Doc provides reference information for Frank developers as explained i
 
 | Kind | Name | Appears on | Function |
 | ---- | ---- | ---------- | -------- |
-| Java annotation | `@FrankDocGroup` | Class or interface | Define group as shown in top-left of webapp, fields `name` and `order`.|
+| Java annotation | `@FrankDocGroup` | Interface | Define group as shown in top-left of webapp. Implementations added to group. They can belong to othe groups as well. |
```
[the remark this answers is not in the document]
mhdirkse: Gefixed.

FRANKDOC.md
```diff
@@ -5,7 +5,8 @@ The Frank!Doc provides reference information for Frank developers as explained i
...
+| Java annotation | `@FrankDocGroup` | Class | Add class and descendants to specified group. They are excluded from other groups. |
```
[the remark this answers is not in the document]
mhdirkse: Gedaan. Nu past elk record weer op één regel.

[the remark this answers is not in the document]
mhdirkse: Verderop vraag je om toevoegingen van één regel te integreren in de tabel. Dat heb ik gedaan. Nu zijn er weer tabel records die meerdere regels beslaan.

FRANKDOC.md
```diff
@@ -18,13 +19,11 @@ The Frank!Doc provides reference information for Frank developers as explained i
...
+**Annotation @FrankDocGroup:**: Has fields `name` and `order` (integer), which is used by the webapp to sort the groups. You can define a group by setting multiple `@FrankDocGroup` annotations with the same `name`. Only one `@FrankDocGroup` annotation of a group should have its `order` field set to avoid ambiguity. This annotation behaves differently on classes and interfaces. When a Java class implements an interface that has a `@FrankDocGroup` annotation, then the class is put in the group non-exclusively. A class can belong to multiple groups when it implements multiple interfaces with different groups. However, when a class has a `@FrankDocGroup` annotation, then the class only belongs to that group. `@FrankDocGroup` annotations on classes are inherited by descendant classes, so descendant classes are also in the specified group exclusively.
 
 **JavaDoc tag @ff.parameter:** First argument is name of parameter, second argument is description of that parameter.
 
 **JavaDoc tag @ff.forward:** First argument is name of forward, second argument is description.
```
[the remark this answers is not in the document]
mhdirkse: Gedaan.

FRANKDOC.md
mhdirkse: Gedaan.

gvanbrakel (changes requested):
Als je de merge conflicts oplost, is hij daarna ok!

mhdirkse:
Ik heb het merge conflict opgelost. Daarbij heb ik README hetzelfde gelaten als hij voor de merge was in deze PR.

gvanbrakel (approved):

nielsm5 (approved):

PR #2491, opened 2021-11-26
Fix raw parameter to target datatype resolution

### LINKED ISSUE #2256: PutParametersInSession with xpathExpression with type=domdoc results in "Content is not allowed in prolog"
**Describe the issue**
After storing (a part of) an XML in a session key using a PutParametersInSession and type="domdoc", further xpathExpressions on that session key result in "Content is not allowed in prolog" errors.

Setting type="xml"  works as intended, but I would expect type="domdoc" to not have different behavior.

The issue is present in 7.5.2, 7.6 and 7.7. 
In 7.5.2 the error is `Parameter [xmlMessageChild2] exception on transformation to get parametervalue: (ClassCastException) org.apache.xerces.dom.DocumentImpl cannot be cast to java.lang.String"`
In 7.6 and 7.7 the error is `Parameter [xmlMessageChild2] exception on transformation to get parametervalue: (XPathException): (SAXParseException) line [1] column [1]: Content is not allowed in prolog."`

**Reporter**
Ricardo

**To Reproduce**

**Screenshots**

**Configuration**
```
<module>
	<adapter name="Repro" description="Repro">
		<receiver name="Repro"
			className="nl.nn.adapterframework.receivers.GenericReceiver">
			<listener name="Repro"
				className="nl.nn.adapterframework.receivers.JavaListener"/>
		</receiver>

		<pipeline firstPipe="Store">
			<exits>
				<exit path="READY" state="success"/>
			</exits>

			<pipe name="Store"
				className="nl.nn.adapterframework.pipes.PutParametersInSession">
				<param name="xmlMessageChild" xpathExpression="Test/Child[1]" type="domdoc"/>
				<forward name="success" path="Store2" />
			</pipe>

			<pipe name="Store2"
				className="nl.nn.adapterframework.pipes.PutParametersInSession">
				<param name="xmlMessageChild2" sessionKey="xmlMessageChild" xpathExpression="Child/name"/>
				<forward name="success" path="READY" />
			</pipe>
		</pipeline>
	</adapter>
</module>
```

**Input**
```
<?xml version="1.0" encoding="UTF-8"?>
<Test>
	<Child><name>X</name></Child>
	<Child><name>Y</name></Child>
	<Child><name>Z</name></Child>
</Test>
```

**Environment**
```
FF! 7.7-20210928.133035: ibis4repro
Running on LPZWNL000014943 using Apache Tomcat/9.0.50
Java Version: OpenJDK Runtime Environment (1.8.0_292-b10)
Heap size: 261M, total JVM memory: 450M
Free disk space: 58GB, total disk space: 237GB
Up since: 2021-09-28 15:56:47 (14m)
```

**Additional Environment**

**Additional Context**

### LINKED ISSUE #2489: Some of the datatypes are not converted to domdoc although the parameter definition has type="domdoc"
**Describe the issue**
During an upgrade of the framework it is observed that the some object types(like byte[], PathMessage, URL) are not converted to domdoc. Therefore the Xslt processor throws an exception with the following message

 Required item type of first operand of '/' is node(); supplied value has item type javatype:nl.nn.adapterframework.stream.PathMessage--Lambda-234/534890610&#xD;
	at nl.nn.adapterframework.senders.XsltSender.sendMessage(XsltSender.java:304)&#xD;

**Reporter**
Ali

alisihab:
Closes #2489 
Closes #2256


core/src/main/java/nl/nn/adapterframework/stream/Message.java
```diff
@@ -395,6 +398,14 @@ public String asString(String defaultCharset) throws IOException {
 		if (request instanceof String) {
 			return (String)request;
 		}
+		if(request instanceof Document) {
```
gvanbrakel, line 401: How would the request become a document? There is no constructor for that.

core/src/test/java/nl/nn/adapterframework/parameters/ParameterTest.java
```diff
@@ -114,13 +114,18 @@ public void testPatternUnknownSessionVariableOrParameter() throws ConfigurationE
...
-		
+
 		ParameterValueList alreadyResolvedParameters=new ParameterValueList();
+
+		try {
+			p.getValue(alreadyResolvedParameters, null, session, false);
+			Assert.fail("supposed to throw an exception");
+		} catch(ParameterException e) {
+			assertEquals("Parameter or session variable with name [unknown] in pattern [{unknown}] cannot be resolved", e.getMessage());
+		}
```
gvanbrakel, line 127: Was it necessary to change the code this way? I would say the previous implementation is more readable

core/src/test/java/nl/nn/adapterframework/parameters/ParameterTest.java
```diff
@@ -190,6 +195,7 @@ public void testParameterValueMessageStream() throws Exception {
 		Message message = new Message("fakeMessage");
 
 		Object result = p.getValue(alreadyResolvedParameters, message, session, false);
+		System.err.println(result.getClass());
```
gvanbrakel, line 198: please remove, or change into log statement

core/src/test/java/nl/nn/adapterframework/pipes/FilePipeTest.java
```diff
@@ -44,13 +42,9 @@ public static void before() throws Exception {
 
     @Test
     public void doTestSuccess() throws Exception {
```
gvanbrakel, line 44: Why were all parameters removed from this test?

gvanbrakel (changes requested):

core/src/test/java/nl/nn/adapterframework/pipes/FilePipeTest.java
alisihab: Because it was non sense. The parameter is defined with a value and a sessionKey. The sessionkey refers to parameter itself. Parameter toString will be something like 'Parameter name=[p1] defaultValue=[null] sessionKey=[key]'. Then this will be converted(an attempt to convert) to Integer since the type is integer. Morevover, the parameter won't be used anywhere.

core/src/main/java/nl/nn/adapterframework/stream/Message.java
alisihab: Message.asMessage(Object)

core/src/main/java/nl/nn/adapterframework/stream/Message.java
```diff
@@ -395,6 +397,14 @@ public String asString(String defaultCharset) throws IOException {
 		if (request instanceof String) {
 			return (String)request;
 		}
+		if(request instanceof Node) {
+			try {
```
gvanbrakel, line 401: I think it would be wise to not limit the support for Node to the asString() method, and extend it to the other as...() methods.
When you do that, please add tests for that in MessageTest.java

core/src/main/java/nl/nn/adapterframework/stream/Message.java
```diff
@@ -395,6 +397,14 @@ public String asString(String defaultCharset) throws IOException {
...
+				request = XmlUtils.nodeToString((Node)request, true);
+			} catch (TransformerException e) {
+				log.warn("Could not convert Document to String", e);
```
gvanbrakel, line 404: Please rethrow the exception, instead of only logging it.

core/src/main/java/nl/nn/adapterframework/parameters/Parameter.java
```diff
@@ -577,6 +509,89 @@ public Object getValue(ParameterValueList alreadyResolvedParameters, Message mes
...
+	/** Converts raw data to configured parameter type */
+	private Object getTypeValue(Object message, boolean namespaceAware) throws ParameterException {
+		Object result = message;
+		if(!(result instanceof String)) {
+			try {
+				result = Message.asMessage(result).asString();
+			} catch (IOException e) {
+				throw new ParameterException("Could not convert parameter ["+getName()+"] to string", e);
+			}
+		}
```
gvanbrakel, line 521: It is a bit sorry to convert all messages to string here, and then later convert them into something else.
Please use message as the main type here.

gvanbrakel (changes requested):

core/src/test/java/nl/nn/adapterframework/parameters/ParameterTest.java
```diff
@@ -218,5 +223,170 @@ public void testParameterValueList() throws Exception {
...
+
+		ParameterValueList alreadyResolvedParameters=new ParameterValueList();
+		Message message = new Message("fakeMessage");
+
+		Object result = p.getValue(alreadyResolvedParameters, message, null, false);
+		assertEquals("a", Message.asMessage(result).asString());
+	}
+	
+	@Test
+	public void testParameterNumberBooelan() throws Exception {
```
nielsm5, line 243: Typo

core/src/main/java/nl/nn/adapterframework/parameters/Parameter.java
```diff
@@ -107,6 +108,8 @@
 	public static final String FIXEDUID ="0a1b234c--56de7fa8_9012345678b_-9cd0";
 	public static final String FIXEDHOSTNAME ="MYHOST000012345";
 
+	private List<ParameterType> typeConversionExclusionList = Arrays.asList(ParameterType.STRING, ParameterType.XML, ParameterType.BINARY, ParameterType.BYTES, ParameterType.INPUTSTREAM, ParameterType.LIST, ParameterType.MAP);
+
```
nielsm5, line 112: Isn't is easier if this would be a field in the enum ParameterType?

core/src/main/java/nl/nn/adapterframework/parameters/Parameter.java
nielsm5: To further elaborate Gerrit's comment here, the `XmlUtils.buildDomDocument` method can also take a Reader for example. Instead of (many) conversions from binary data to character data (or whatever) it can also be directly parsed with the correct reader, improving performance.
If you were to create a new method to convert the Message to a String, you would also no longer need the casts `(String) result` nor `result.toString()` when parsing the value.

core/src/main/java/nl/nn/adapterframework/stream/Message.java
nielsm5: Is this used anywhere? If not it might be best to not implement Node at all...

nielsm5 (changes requested):

core/src/main/java/nl/nn/adapterframework/parameters/Parameter.java
```diff
@@ -488,93 +497,115 @@ public Object getValue(ParameterValueList alreadyResolvedParameters, Message mes
...
 						if (transformerPoolRemoveNamespaces != null) {
-							result = transformerPoolRemoveNamespaces.transform((String)result, null);
+							request = new Message(transformerPoolRemoveNamespaces.transform(request, null));
+						}
+						Object requestObject = request.asObject();
+						if(requestObject instanceof Document) {
+							return requestObject;
 						}
-						result=XmlUtils.buildDomDocument((String)result,namespaceAware);
+						result=XmlUtils.buildDomDocument(request.asString() ,namespaceAware);
```
nielsm5, line 551: Why not buildDomDocument(message.asReader(), namespaceAware) ?

core/src/main/java/nl/nn/adapterframework/parameters/Parameter.java
```diff
@@ -488,93 +497,115 @@ public Object getValue(ParameterValueList alreadyResolvedParameters, Message mes
...
+						}
+						Object requestObject = request.asObject();
+						if(requestObject instanceof Document) {
+							return ((Document)requestObject).getDocumentElement();
 						}
-						result=XmlUtils.buildNode((String)result,namespaceAware);
+						if(requestObject instanceof Node) {
+							return requestObject;
+						}
+						result=XmlUtils.buildNode(request.asString(), namespaceAware);
```
nielsm5, line 536: buildNode indirectly calls buildDomDocument, you could use a reader instead

core/src/main/java/nl/nn/adapterframework/parameters/Parameter.java
```diff
@@ -488,93 +497,116 @@ public Object getValue(ParameterValueList alreadyResolvedParameters, Message mes
...
+						}
+						Object requestObject = request.asObject();
+						if(requestObject instanceof Document) {
+							return ((Document)requestObject).getDocumentElement();
 						}
-						result=XmlUtils.buildNode((String)result,namespaceAware);
+						if(requestObject instanceof Node) {
+							return requestObject;
+						}
+						result=XmlUtils.buildDomDocument(request.asReader(), namespaceAware).getDocumentElement();
```
gvanbrakel, line 536: Please use request.asSource() here.

core/src/main/java/nl/nn/adapterframework/parameters/Parameter.java
```diff
@@ -488,93 +497,116 @@ public Object getValue(ParameterValueList alreadyResolvedParameters, Message mes
...
 						if (transformerPoolRemoveNamespaces != null) {
-							result = transformerPoolRemoveNamespaces.transform((String)result, null);
+							request = new Message(transformerPoolRemoveNamespaces.transform(request, null));
+						}
+						Object requestObject = request.asObject();
+						if(requestObject instanceof Document) {
+							return requestObject;
 						}
-						result=XmlUtils.buildDomDocument((String)result,namespaceAware);
+						result=XmlUtils.buildDomDocument(request.asReader(), namespaceAware);
```
gvanbrakel, line 551: Please use request.asSource() here.

gvanbrakel (changes requested):

core/src/main/java/nl/nn/adapterframework/stream/Message.java
```diff
@@ -395,6 +424,13 @@ public String asString(String defaultCharset) throws IOException {
 		if (request instanceof String) {
 			return (String)request;
 		}
+		if(request instanceof Node) {
+			try {
+				return XmlUtils.nodeToString((Node)request, true);
```
nielsm5, line 429: Hier een log.warn -> performance impact

nielsm5 (changes requested):

nielsm5 (approved):

gvanbrakel (approved):

PR #2492, opened 2021-11-27
Add Webjars support

jacodg:


jacodg:
We will use Servlet 3

PR #2513, opened 2021-12-06
Multipassword keystore

gvanbrakel:


core/src/test/java/nl/nn/adapterframework/http/HttpSenderTest.java
```diff
@@ -831,4 +831,84 @@ public void paramsWithoutValue() throws Throwable {
...
+		HttpSender sender = getSender();
+		sender.setKeystore(pfxCertificate);
+		sender.setKeystoreAuthAlias("ks_alias");
+		sender.setKeyAlias("2nd");
+		sender.setKeystoreAliasAuthAlias("key2");
+
+		sender.setMethodType(HttpMethod.GET);
+		sender.configure();
+
+	}
```
nielsm5, line 913: Could you fix the test names and also add tests for keystores?

nielsm5 (changes requested):

core/src/main/java/nl/nn/adapterframework/http/HttpSenderBase.java
```diff
@@ -367,13 +373,17 @@ public void configure() throws ConfigurationException {
 			javax.net.ssl.SSLSocketFactory socketfactory = (javax.net.ssl.SSLSocketFactory) javax.net.ssl.SSLSocketFactory.getDefault();
 			sslSocketFactory = new SSLConnectionSocketFactory(socketfactory, hostnameVerifier);
```
nielsm5, line 374: The alias password now only works when a keystoreUrl is provided, but what if you want to use the default keystore?

core/src/main/java/nl/nn/adapterframework/filesystem/FileSystemPipe.java
```diff
@@ -43,6 +43,12 @@
  * 
  * @see FileSystemActor
  * 
+ * @ff.parameter action overrides attribute <code>action</code>
```
nielsm5, line 46: punt

nielsm5 (approved):

PR #2523, opened 2021-12-10
Use ApiListener for example configurations

jacodg:


gvanbrakel (commented):
I would like to suggest to not only change the listener types, but let the adapters implement something that looks like a use case of the framework.

jacodg:
I agree that better examples can be created and we should certainly look into it. But apart from that I think ApiListeners will already make the current examples a bit better. It will enable you to download the OpenApi spec and call /api and get a result. And for new people an ApiListeners will say more than a JavaListener

gvanbrakel (approved):

PR #2542, opened 2021-12-15
Fix LiquibaseScriptGenerator

### LINKED ISSUE #2518: Liquibase script generator in Frank!Console produces error on ibis4example
**Describe the issue**
When I browse to https://ibis4example.ibissource.org/ and run the Liquibase script generator, I see an error. See picture below:

image

I got this error after pressing the button "Generate SQL"

**Reporter**
Martijn Dirkse

**To Reproduce**
1. Browse to https://ibis4example.ibissource.org/.
2. Click JDBC | Liquibase Script.
3. Press "Generate SQL"

**Screenshots**
See above.

**Configuration**
```
```

**Input**
```
```

**Environment**
```
```

**Additional Environment**
 - Google Chrome.

**Additional Context**

alisihab:
Fixes #2518


core/src/main/java/nl/nn/adapterframework/webcontrol/api/ShowLiquibaseScript.java
```diff
@@ -79,7 +79,7 @@ public Response downloadScript() throws ApiException {
 
 		for(Configuration config : getIbisManager().getConfigurations()) {
 			DatabaseMigratorBase databaseMigrator = config.getBean("jdbcMigrator", DatabaseMigratorBase.class);
-			if(databaseMigrator.isEnabled()) {
```
nielsm5, line 82: I would personally prefer a big fat warning that somebody enabled the migrator but didn't supply the migration script.
I think the real issue here, is that liquibase can be enabled from the 'src/main/resources' AppConstants instance. It should only check in it's own context (just like the migration script.

alisihab: It is not necessary to set a property to use this tool. Having a script in the configuration classpath is sufficient.

core/src/main/java/nl/nn/adapterframework/jdbc/migration/DatabaseMigratorBase.java
```diff
@@ -135,4 +135,18 @@ public final ApplicationContext getApplicationContext() {
...
+	public boolean hasLiquibaseScript(Configuration config) {
+		AppConstants appConstants = AppConstants.getInstance(config.getClassLoader());
+		String changeLogFile = appConstants.getString("liquibase.changeLogFile", "DatabaseChangelog.xml");
+		LiquibaseResourceAccessor resourceAccessor = new LiquibaseResourceAccessor(config.getClassLoader());
+		if(resourceAccessor.getResource(changeLogFile) == null) {
+			log.debug("No liquibase script ["+changeLogFile+"] found in the classpath of configuration ["+config.getName()+"]");
+			return false;
+		}
+		return true;
+	}
```
nielsm5, line 151: Can you then please override `isEnabled` in LiquibaseMigrator and call the method `getChangeLogFile`? The method returns a FNFE, I think the easiest is to catch the exception and return false. That way the isEnabled works uniformly, and there is no duplicate code.

nielsm5 (approved):

gvanbrakel (approved):

PR #2546, opened 2021-12-16
Enable JBOSS container

ricardovh:
This PR fixes JBOSS containers by upgrading the container and changing the build. First commit should still work against the old image version.

This PR works only when reactive-streams are added as dependency on the webapp (see #2539), so on its own this PR will not work.

Additional information:
Not all tests pass (Oracle has 6 tests not passing, MSSQL has a lot of tests not passing because of IDENTITY_INSERTS failing), ActiveMQ is not tested, and the WildFly server properties result in a container that does not work.


ricardovh:
Replaced by #3436 and #3315

PR #2580, opened 2021-12-31
Add ability to specify message property from the gui for jms messages

alisihab:


webapp/src/main/webapp/iaf/gui/js/controllers.js
```diff
@@ -2196,6 +2196,8 @@ angular.module('iaf.beheerconsole')
 			return;
 		}
 
+		if(formData.propertyKey && formData.propertyKey != "" && formData.propertyValue && formData.propertyValue)
```
nielsm5, line 2199: I think your forgot to add `!= ""` :)

webapp/src/main/webapp/iaf/gui/views/SendJmsMessage.html
```diff
@@ -59,6 +59,18 @@ <h4>Send JMS Message</h4>
 								<span class="form-control" style="border: none;"><input icheck type="checkbox" name="lookupDestination" ng-model="form.lookupDestination" /></span>
 							</div>
 						</div>
+						<div class="row form-group">
+							<label class="col-sm-7 col-md-3 control-label label-height-30">Message Property</label>
+							<div class="col-sm-7 col-md-3">
+								<input class="form-control" name="propertyKey" type="text" ng-model="form.propertyKey"/>
+							</div>
+							<div class="col-sm-7 col-md-1">
+								<label class="control-label label-height-30 pull-right">Value</label>
```
nielsm5, line 68: Why is there a `pull-right` over here?
Please ensure that when using col-*-n the total 'per line' value never crosses 12.

nielsm5 (changes requested):

webapp/src/main/webapp/iaf/gui/views/SendJmsMessage.html
alisihab: it puts too many space between the Value text and the text box. Line 86 has the same expression.

nielsm5: Not quite, like 86 has `hidden-xs` thus it only does pull-right when the screen is larger then 992px.
Right now it always does the pull right, which might look ok when you're on a wide screen, but not when it's smaller

alisihab: punt

nielsm5 (approved):

gvanbrakel (approved):

PR #2581, opened 2022-01-03
Change jmsRealm deprecation text

nielsm5:


core/src/main/java/nl/nn/adapterframework/jdbc/JdbcFacade.java
```diff
@@ -202,7 +202,7 @@ public void iterateOverStatistics(StatisticsKeeperIterationHandler hski, Object
 
 	@Override
 	@Deprecated
-	@ConfigurationWarning("Please use attribute dataSourceName instead")
+	@ConfigurationWarning("We discourage the use of jmsRealms for datasources, please use the datasourceName attribute directly instead of specifying a realm")
```
gvanbrakel, line 205: May be you could even add: ''to specify another datasource than the default"

gvanbrakel (approved):

PR #2685, opened 2022-01-27
Use Maven CI friendly versions

ricardovh:
As described on https://maven.apache.org/maven-ci-friendly.html

This should also fix the problems with Websphere container image dependencies not being resolved (it does locally).

TODO: 
- Check if this does not accidentally break something in CICD/release plugin workings
- Contributing for required maven version


nielsm5:
You've now turned all parent references into a property defined in the parent. How does the system know what value to resolve the property to if it cannot find the parent?

docker/appserver/Jboss-WildFly/pom.xml
```diff
@@ -16,7 +16,6 @@
 		<dependency>
 			<groupId>org.ibissource</groupId>
 			<artifactId>ibis-adapterframework-ear</artifactId>
-			<version>${iaf.version}</version>
```
nielsm5, line 19: <version>${project.version}</version>

nielsm5:
https://medium.com/outbrain-engineering/faster-release-with-maven-ci-friendly-versions-and-a-customised-flatten-plugin-fe53f0fcc0df

nielsm5:
Ik denk dat het zo goed is.

gvanbrakel (approved):

PR #2691, opened 2022-01-28
Add support for having parameters in xpath in XmlIf pipe

### LINKED ISSUE #2284: The XmlIfPipe can't work with parameters
**Describe the issue**
The XmlIfPipe can't work with parameters. If an xpathExpression in the XmlIfPipe refers to a parameter that is defined in the pipe, it will throw an error and it won't start up.

**Reporter**
Lenard van der Maas

**To Reproduce**
1. Create an XmlIfPipe in an adapter, with an xpath that calls a parameter using `$paramName`
2. Start the Frank
3. Go to the adapter in the adapter overview
4. ERROR: error initializing pipeline: Pipe [AdapterName] could not create transformer from xpathExpression [example/subElement[text() = $paramName]], target expressionValue [null]: (TransformerConfigurationException): (XPathException) line [1]: Variable paramName has not been declared

**Configuration**
```
<pipe name="CheckSwitchIBANs"
	className="nl.nn.adapterframework.pipes.XmlIf"
	xpathExpression="contains(result/rowset/row/FIELD[name='VALUE']/text(), $iban)">
	<param name="iban" sessionKey="IBAN" defaultValue="|NonExistentIBAN|"/>
	<forward name="then" path="GetPaylink ApiSender"/>
	<forward name="else" path="GetPaylink QueueSender"/>
</pipe>
```
**Additional Context**
A different version of this error will also occur when using a stylesheet instead of an xpathExpression.
I suggest changing the way transformations are implemented to how it's done in the XsltSender. That should solve the problem.

alisihab:
Closes #2284

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/pipes/XmlIf.java
+++ b/core/src/main/java/nl/nn/adapterframework/pipes/XmlIf.java
@@ -1,5 +1,5 @@
 /*
-   Copyright 2013, 2020 Nationale-Nederlanden, 2021 WeAreFrank!
+   Copyright 2013, 2020 Nationale-Nederlanden, 2021, 2022 WeAreFrank!
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


core/src/main/java/nl/nn/adapterframework/pipes/XmlIf.java
```diff
@@ -116,35 +126,25 @@ public PipeRunResult doPipe(Message message, PipeLineSession session) throws Pip
...
 
-		// log.debug(getLogPrefix(session) + "input value is [" + sInput + "]");
-		
 		if (tp!=null) {
 			try {
-				forward = tp.transform(sInput,null, isNamespaceAware());
+				Map<String,Object> parametervalues = null;
+				ParameterList parameterList = getParameterList();
+				if (!parameterList.isEmpty()) {
+					message.preserve();
```
gvanbrakel, line 134: `message.preserve()` should only be done if `parameterList.requiresInputValueForResolution()`

gvanbrakel (changes requested):

gvanbrakel (approved):

nielsm5 (approved):

PR #2696, opened 2022-02-01
Preserve message if a parameter requires message

alisihab:


core/src/main/java/nl/nn/adapterframework/soap/SoapWrapperPipe.java
```diff
@@ -229,8 +229,10 @@ public PipeRunResult doPipe(Message message, PipeLineSession session) throws Pip
 					payload = new Message(outputNamespaceTp.transform(payload.asSource()));
 				}
 				Map<String,Object> parameterValues = null;
-				if (getParameterList()!=null && (soapHeaderTp != null || soapBodyTp != null)) {
-					payload.preserve();
+				if (!getParameterList().isEmpty() && (soapHeaderTp != null || soapBodyTp != null)) {
+					if(getParameterList().parameterEvaluationRequiresInputMessage()) {
```
gvanbrakel, line 233: It would be good to do this evaluation only once, in `configure()`

gvanbrakel (changes requested):

gvanbrakel (approved):

nielsm5 (approved):

PR #2730, opened 2022-02-08
Add dockerfile for activemq-artemis

alisihab:


core/src/main/resources/META-INF/services/javax.json.spi.JsonProvider
```diff
@@ -0,0 +1 @@
+org.glassfish.json.JsonProviderImpl
```
nielsm5, line 1: Goede werken! Misschien deze ook in de webapp module zetten.

nielsm5 (changes requested):

test/src/main/webapp/META-INF/context.xml
```diff
@@ -180,5 +180,13 @@
...
 		type="org.apache.activemq.ActiveMQXAConnectionFactory"
 		brokerURL="tcp://host.docker.internal:61616"
-	/>	
+	/>
+
+	<!-- Activemq-artemis resource configuration -->
+	<Resource name="jms/qcf-activemq-artemis"
+		factory="org.apache.naming.factory.BeanFactory"
+		type="org.apache.activemq.artemis.jms.client.ActiveMQXAConnectionFactory"
+		brokerURL="tcp://host.docker.internal:61615"
```
gvanbrakel, line 189: Here you use port 61615, while you published port 61616. I don't think that is close enough!

gvanbrakel (changes requested):

gvanbrakel (approved):

nielsm5 (approved):

PR #2768, opened 2022-02-16
Make larva scenario file parameters available as FileMessage in session

alisihab:


larva/src/main/java/nl/nn/adapterframework/testtool/TestTool.java
```diff
@@ -4203,138 +4205,123 @@ public static String replaceRegularExpression(String string, String from, String
 	 */
 	private static Map<String, Object> createParametersMapFromParamProperties(Properties properties, String property, Map<String, Object> writers, boolean createParameterObjects, PipeLineSession session) {
 		debugMessage("Search parameters for property '" + property + "'", writers);
-		Map<String, Object> result = new HashMap<String, Object>();
+		final String _name = ".name";
+		final String _param = ".param";
+		final String _type = ".type";
```
nielsm5, line 4210: I hate to say it but this does make it more difficult to figure out what has actually been changed :)

larva/src/main/java/nl/nn/adapterframework/testtool/TestTool.java
```diff
@@ -4203,138 +4205,123 @@ public static String replaceRegularExpression(String string, String from, String
...
 						} else {
-							String inputStreamFilename = properties.getProperty(property + ".param" + i + ".valuefileinputstream.absolutepath");
+							String inputStreamFilename = properties.getProperty(property + _param + i + ".valuefileinputstream.absolutepath");
 							if (inputStreamFilename != null) {
-								try {
-									value = new FileInputStream(inputStreamFilename);
-								} catch(FileNotFoundException e) {
-									errorMessage("Could not read file '" + inputStreamFilename + "': " + e.getMessage(), e, writers);
-								}
+								errorMessage("valuefileinputstream is no longer supported use valuefile instead", writers);
```
nielsm5, line 4290: I hate to say it, but perhaps we could introduce a 'warningMessage' which is orange (ish).
I'm ok with a TODO here for now.

nielsm5 (changes requested):

larva/src/main/java/nl/nn/adapterframework/testtool/TestTool.java
gvanbrakel: yes, but later on it helps to find out where it is used!

larva/src/main/java/nl/nn/adapterframework/testtool/TestTool.java
gvanbrakel: We could still retain the valuefileinputstream for case where one really wants a non-repeatable stream

larva/src/main/java/nl/nn/adapterframework/testtool/TestTool.java
```diff
@@ -4203,138 +4205,123 @@ public static String replaceRegularExpression(String string, String from, String
...
-				if (value != null && value instanceof String) {
-					if ("node".equals(properties.getProperty(property + ".param" + i + ".type"))) {
+				if(typesRequireStringType.contains(type) && value instanceof Message) {
+					try {
+						value = ((Message)value).asString();
+					} catch (IOException e) {
+						errorMessage("Could not convert parameter '" + name + "' with value: " + value +" to string", e, writers);
+					}
+				}
+				if (value instanceof String) {
```
gvanbrakel, line 4302: Instead of introducing `typesRequireStringType` you could just do `value.asString()` instead of `(String)value`

gvanbrakel (changes requested):

larva/src/main/java/nl/nn/adapterframework/testtool/TestTool.java
alisihab: In this case, valuefileinputsteam usage is removed. therefore an errormessage fits better here

gvanbrakel (approved):

nielsm5 (approved):

PR #2795, opened 2022-02-21
Make XPath/XSLT version configurable by a property

alisihab:


core/src/main/resources/AppConstants.properties
```diff
@@ -180,6 +180,7 @@ xml.namespaceAware.default=false
 xslt.streaming.default=true
 xslt.auto.reload=false
 xslt.bufsize=4096
+xslt.default.version=2
```
gvanbrakel, line 183: In the current (unwritten) convention it would seem more logical to use the name 'xslt.version.default'

gvanbrakel (changes requested):

gvanbrakel (approved):

nielsm5 (approved):

PR #2805, opened 2022-02-24
Create suppress key for xslt streaming warning

### LINKED ISSUE #2775: No option to suppress warning "ForEachChildElementPipe [] XsltProcessor xsltVersion [2] currently does not support streaming XSLT, might lead to memory problems for large message"

**Describe the issue**
There is no option to suppress the warning "ForEachChildElementPipe [] XsltProcessor xsltVersion [2] currently does not support streaming XSLT, might lead to memory problems for large messages".

Usually XSLT v1 has enough functionality for elementXPathExpression, but XPath 2 has more functions and sometimes those are needed to filter certain elements out correctly. For small XMLs it is then fine to set xsltVersion 2, but there is no option to suppress the warning.

**Reporter**
Ricardo

**To Reproduce**

**Screenshots**

**Configuration**
```
```

**Input**
```
```

**Environment**
```
```

**Additional Environment**

**Additional Context**

alisihab:
Closes #2775

### DIFF
--- a/core/src/main/resources/AppConstants.properties
+++ b/core/src/main/resources/AppConstants.properties
@@ -184,7 +184,7 @@ xslt.bufsize=4096
 # The ForEachChildElementPipe usually works with large files and thus is streaming preferred.
 # Currently streaming is only supported with XSLT version 1.
 # Therefore this property will not change the XSLT version in the ForEachChildElementPipe.
-# If desired, the XSLT version can be change for each individual ForEachChildElementPipe using xsltVersion=2.
+# If desired, the XSLT version can be changed for each individual ForEachChildElementPipe using xsltVersion=2.
 xslt.version.default=2
 
 xmlValidator.warn=true
--- a/core/src/main/resources/AppConstants.properties
+++ b/core/src/main/resources/AppConstants.properties
@@ -180,6 +180,10 @@ xml.namespaceAware.default=false
 xslt.streaming.default=true
 xslt.auto.reload=false
 xslt.bufsize=4096
+# ForEachChildElement pipe usually works with large files and uses streaming.
+# Currently streaming is only supported with xslt version 1.
+# Therefore this property will not change the xslt version in ForEachChildElementPipe.
+# If desired, xsltVersion can be set for the individual pipe.
 xslt.version.default=2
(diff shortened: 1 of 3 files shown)


core/src/main/resources/AppConstants.properties
```diff
@@ -180,6 +180,10 @@ xml.namespaceAware.default=false
 xslt.streaming.default=true
 xslt.auto.reload=false
 xslt.bufsize=4096
+# ForEachChildElement pipe usually works with large files and uses streaming.
+# Currently streaming is only supported with xslt version 1.
+# Therefore this property will not change the xslt version in ForEachChildElementPipe.
+# If desired, xsltVersion can be set for the individual pipe.
```
gvanbrakel, line 186: This is a confusing comment if you put is this way. Please first explain/document the property as it is, and then note the exception

gvanbrakel (changes requested):

core/src/main/resources/AppConstants.properties
nielsm5, line 186:
```suggestion
# Sets the default XSLT version throughout the entire application except the ForEachChildElementPipe.
# The ForEachChildElementPipe usually works with large files and thus is streaming preferred.
# Currently streaming is only supported with XSLT version 1.
# Therefore this property will not change the XSLT version in the ForEachChildElementPipe.
# If desired, the XSLT version can be change for each individual ForEachChildElementPipe using xsltVersion=2.
```

nielsm5, line 186:
```suggestion
# Sets the default XSLT version throughout the entire application except the ForEachChildElementPipe.
# The ForEachChildElementPipe usually works with large files and thus is streaming preferred.
# Currently streaming is only supported with XSLT version 1.
# Therefore this property will not change the XSLT version in the ForEachChildElementPipe.
# If desired, the XSLT version can be change for each individual ForEachChildElementPipe using xsltVersion=2.
```

gvanbrakel:
Could you provide a backport to 7.7 too?

gvanbrakel (approved):

nielsm5 (approved):

PR #2842, opened 2022-03-09
Improve message store and small fixes

### LINKED ISSUE #2762: filter on payload of errorstore/message log not user friendly in new GUI

**Describe the issue**
In "old GUI"  has the possibility to filter/search on the content of the messages in the errorstore. there is no search field for message in the new GUI.
the requested functionality is to be able to filter messages on based on the existence of a certain string in the message 

UPDATE
we were testing with 7.6 and the extra field is added.
the functionality works but is not clear to the user
- If you open a message and push the back button the filter is cleared (also in other interaction the filter is cleared
- there is no obvious trigger moment to start the filtering (should be more clear how this works)
- the processing message is not in your screen, if you scroll down you see a message not similar to the other "progress bar" images
- the total numbers do not change based on the filtered documents (fixed in 7.4) 

as we cannot upgrade and want to investigate a production issue a work around would be appreciated

** if available please provide workaround via SQL statement **

**Reporter**
jjansenvr on behalve of Fop

**To Reproduce**
1. Go to an adapter with errostor or messagelog
2. open filter section
3. no message filed to do a filter on message content

**Screenshots**
image

**Configuration**
```
n/a
```

**Input**
```
```

**Environment**
```
IAF 7.6.x 
```

**Additional Environment**
 - DBMS:  Oracle
 - 
**Additional Context**

alisihab:
Closes #2762


core/src/main/java/nl/nn/adapterframework/webcontrol/api/TransactionalStorage.java
```diff
@@ -85,21 +93,29 @@ public Response browseReceiverMessage(
...
+				throw new ApiException("Receiver ["+storageSourceName+"] not found!");
+			}
+			storage = receiver.getMessageBrowser(ProcessState.getProcessStateFromName(processState));
+			message = getMessage(storage, receiver.getListener(), messageId);
 		}
 
-		IMessageBrowser<?> storage = receiver.getMessageBrowser(ProcessState.getProcessStateFromName(processState));
 		try {
-			// messageId is double URLEncoded, because it can contain '/' in ExchangeMailListener
-			messageId = Misc.urlDecode(messageId);
```
gvanbrakel, line 96: What happened to this?

core/src/main/java/nl/nn/adapterframework/webcontrol/api/TransactionalStorage.java
```diff
@@ -85,21 +93,29 @@ public Response browseReceiverMessage(
...
+			}
+			storage = pipe.getMessageLog();
+			message = getMessage(storage, messageId);
+		} else {
+			Receiver<?> receiver = adapter.getReceiverByName(storageSourceName);
+			if(receiver == null) {
+				throw new ApiException("Receiver ["+storageSourceName+"] not found!");
+			}
+			storage = receiver.getMessageBrowser(ProcessState.getProcessStateFromName(processState));
+			message = getMessage(storage, receiver.getListener(), messageId);
```
gvanbrakel, line 113: Thanks for combining this code!
This code, that is repeated similarly below, could benefit from defining an interface IMessageStorageOwner or so.
Then it would/should be enough to call
```
IMessageStorageOwner storageOwner = getMessageStorageOwner(storageSource, storageSourceName);
storageOwner.<doTheThingsRequired>;
```

gvanbrakel (changes requested):

core/src/main/java/nl/nn/adapterframework/receivers/JavaListener.java
```diff
@@ -57,7 +57,7 @@
  *
  * @author  Gerrit van Brakel
  */
-public class JavaListener implements IPushingListener<String>, RequestProcessor, HasPhysicalDestination {
+public class JavaListener implements IPushingListener<Message>, RequestProcessor, HasPhysicalDestination {
```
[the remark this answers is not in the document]
alisihab: This is or a different approach is necessary since we store Message in database and try to extract data while using message filter in error store. Currently, it throws ClassCastException since the extractMessage expects string when we provide Message. Another approach to solve this could be to pass string instead of message to a javaListener extractMessage method. But I think it is better to start using Message type in JavaListener

core/src/main/java/nl/nn/adapterframework/webcontrol/api/TransactionalStorage.java
alisihab: I reordered some of the lines here. This line is moved to line 98.

core/src/main/java/nl/nn/adapterframework/receivers/JavaListener.java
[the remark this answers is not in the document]
gvanbrakel: But currently JavaListener is of type IPushingListener<String>, and should work like that. This change appears to me as a workaround for another problem.
I think the problem is in MessageBrowsingFilter.getRawMessage().
1) It does not return a rawMessage, but returns messagebody-as-text. This method should be called something like getMessageText 
2) on line 176, the result of extractMessage, that is a Message, is assigned to a String.

This problem becomes clear when the signature of getRawMessage is specified better, like:
`private String getRawMessage(IMessageBrowser<?> messageBrowser, IListener<?> listener, String messageId)

core/src/main/java/nl/nn/adapterframework/webcontrol/api/TransactionalStorage.java
gvanbrakel: OK, but please keep the comment about the double URLEncoding

gvanbrakel (changes requested):

gvanbrakel (approved):

core/src/main/java/nl/nn/adapterframework/webcontrol/api/TransactionalStorage.java
```diff
@@ -67,14 +73,16 @@
 public class TransactionalStorage extends Base {
 
 	protected static final TransactionDefinition TXNEW = new DefaultTransactionDefinition(TransactionDefinition.PROPAGATION_REQUIRES_NEW);
+	private static final String STORAGESOURCE_TYPE_PIPE = "pipes";
 
 	@GET
 	@RolesAllowed({"IbisDataAdmin", "IbisAdmin", "IbisTester"})
-	@Path("/adapters/{adapterName}/receivers/{receiverName}/stores/{processState}/messages/{messageId}")
+	@Path("/adapters/{adapterName}/{storageSource}/{storageSourceName}/stores/{processState}/messages/{messageId}")
```
nielsm5, line 80: storagesource can be one of two things correct?
receivers or pipes
Perhaps you can turn this into an enum?
``` 
public enum StorageSource {
		RECEIVERS,PIPES;

		public static StorageSource fromString(String value) {
			return EnumUtils.parse(StorageSource.class, value);
		}
	}
```

webapp/src/main/webapp/iaf/gui/js/controllers.js
```diff
@@ -1441,39 +1463,67 @@ angular.module('iaf.beheerconsole')
...
+		host: searchSession ? searchSession["host"] : "",
+		messageId: searchSession ? searchSession["messageId"] : "",
+		correlationId: searchSession ? searchSession["correlationId"] : "",
+		comment: searchSession ? searchSession["comment"] : "",
+		label: searchSession ? searchSession["label"] : "",
+		message: searchSession ? searchSession["message"] : ""
+	};
+
+	$scope.clearSearch = function() {
+		$scope.clear = true;
```
nielsm5, line 1504: perhaps rename this to clearLadda or clearSearchLadda

nielsm5, line 1504: perhaps rename this to clearLadda or clearSearchLadda

webapp/src/main/webapp/iaf/gui/js/controllers.js
```diff
@@ -1565,180 +1640,26 @@ angular.module('iaf.beheerconsole')
...
-			var row = $(row);// .children("td:first").addClass("m-b-xxs");
-			row.children("td.date").each(function(_, element) {
-				var time = $(this).text();
-				if(time)
-					$(element).attr({"to-date": "", "time": time });
-			});
-			var scope = $scope.$new();
-			scope.message = data;
-			$compile(row)(scope);
-		},
```
nielsm5, line 1661: You've removed this

nielsm5 (changes requested):

nielsm5 (changes requested):

nielsm5 (approved):

PR #2867, opened 2022-03-15
Unescape escaped message in messagestorelistener (7.7)

alisihab:


nielsm5 (approved):

gvanbrakel (approved):

nielsm5:
Perhaps you could add a test to prove that this works?

PR #2893, opened 2022-03-18
Pass project version to frank-doc

mhdirkse:


gvanbrakel:
Could you compare to the master into this PR? The current head of master already contains this addition

mhdirkse:
Done. @gvanbrakel you are right that this change was already in. Now there is no change left in this PR.

PR #2898, opened 2022-03-21
Fix MSSQL container init by not waiting fixed sleep time that might be insufficient

ricardovh:


docker/dbms/MsSqlServer/dockerInit.sh
```diff
@@ -4,16 +4,21 @@
...
+	
+	echo "Running create_user script"
+	for i in {1..120};
+		do
+		    if /opt/mssql-tools/bin/sqlcmd -S localhost -U sa -P MssqlPass123 -d master -i /create_user.sql
+		    then
+		        echo "create_user completed"
+		        break
+		    else
+		        sleep 1
```
gvanbrakel, line 19:
```suggestion
		        if [[ $i == 120 ]]
		        then
		                echo "WARN: timeout in create_user"
		                break
		        else         
		                sleep 1
		        fi
```

gvanbrakel (changes requested):

gvanbrakel (approved):

nielsm5 (approved):

PR #2908, opened 2022-03-22
Resolve properties both in read and write steps in Larva

### LINKED ISSUE #2886: Larva does not replace properties in expected result

**Describe the issue**
When using properties in files used by Larva for .read/.write steps, property replacement is done only on the write steps, which can result in failures if such a property is also in a response that is compared in a .read step.

**Reporter**
Ricardo

**To Reproduce**

**Screenshots**

**Configuration**
```
```

**Input**
```
```

**Environment**
```
FF! 7.8-20220315.180221: ibis4repro
Running on LPZWNL000033867 using Apache Tomcat/9.0.56
Java Version: OpenJDK Runtime Environment (1.8.0_292-b10)
Heap size: 275M, total JVM memory: 491M
Free disk space: 79GB, total disk space: 237GB
Up since: 2022-03-18 08:27:47 (6m)
```

**Additional Environment**

**Additional Context**
[src.zip](https://github.com/ibissource/iaf/files/8302471/src.zip)

alisihab:
Closes #2886

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/soap/SoapValidator.java
+++ b/core/src/main/java/nl/nn/adapterframework/soap/SoapValidator.java
@@ -37,6 +37,12 @@ import nl.nn.adapterframework.validation.RootValidations;
  * Before the <code>outputSoapBody</code> attribute was introduced, two validators were used for a request-reply pattern (an inputValidator for the request and an outputValidator for the reply).
  * These inputValidator and outputValidator were identical except for the child element of the SOAP body. Because validators use relatively a lot of memory, the <code>outputSoapBody</code> attribute was added which replaces the outputValidator.
  * Both the request and the reply are then validated by the inputValidator.
+ * <p>To generate a wsdl with a soap action included one of the following properties must be set</p>
+ * <table border="1">
+ * <tr><td>wsdl.${name}.${listenerName}.soapAction</td></tr>
+ * <tr><td>wsdl.${name}.soapAction</td></tr>
+ * <tr><td>wsdl.soapAction</td></tr>
+ * </table>
  *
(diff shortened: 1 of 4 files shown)


core/src/main/java/nl/nn/adapterframework/soap/SoapValidator.java
```diff
@@ -37,6 +37,12 @@
  * Before the <code>outputSoapBody</code> attribute was introduced, two validators were used for a request-reply pattern (an inputValidator for the request and an outputValidator for the reply).
  * These inputValidator and outputValidator were identical except for the child element of the SOAP body. Because validators use relatively a lot of memory, the <code>outputSoapBody</code> attribute was added which replaces the outputValidator.
  * Both the request and the reply are then validated by the inputValidator.
+ * <p>To generate a wsdl with a soap action included one of the following properties must be set</p>
+ * <table border="1">
+ * <tr><td>wsdl.${name}.${listenerName}.soapAction</td></tr>
+ * <tr><td>wsdl.${name}.soapAction</td></tr>
```
gvanbrakel, line 43: Please specify what name. I assume it is adapterName

gvanbrakel (changes requested):

gvanbrakel:
On the issue it was commented:
"We have discussed this in pr https://github.com/ibissource/iaf/pull/730 where we concluded that this would break the save functionality. If such a property would be in the response the correct thing to do would be to ignore it using one of the many ignore/replace options."
I suggest to remove the actual code of the PR, but keep the changes to the WSDL SoapAction configuration/documentation.

gvanbrakel (approved):

nielsm5 (approved):

PR #2916, opened 2022-03-23
Search recursively in frankdoc

### LINKED ISSUE #2860: Frank!Doc search does not search data of inherited classes

**Describe the issue**
When searching the Frank!Doc I ran into the issue that I could not find anything based on search key "xPathLoggingKeys" which is located in the JmsListenerBase.

the search results in 0 results. 

**Reporter**
JJansenvr

**To Reproduce**

1. go to https://frankdoc.frankframework.org/
2. type xPathLoggingKeys in the search field
3. observe you see zero results
4. clear search field and click listeners and select PushingJmsListener
5. observe you see the attribute xPathLoggingKeys on the JmsListenerBase

**Screenshots**

**Configuration**
```
```

**Input**
```
```

**Environment**
```
```

**Additional Environment**

**Additional Context**

alisihab:
Fixes #2860


webapp/src/main/webapp/iaf/frankdoc/js/app.js
```diff
@@ -102,6 +102,16 @@ angular.module('iaf.frankdoc').config(['$stateProvider', '$urlRouterProvider', f
...
+					} else { // search in parent
+						let elementParent = elements[element].parent;
+						while(elementParent){
+							let parentObj = elements[elementParent];
+							if(JSON.stringify(parentObj).replace(/"/g, '').toLowerCase().indexOf(searchTextLC) > -1) {
+								r[element] = obj;
+								break;
+							}
+							elementParent = elements[elementParent].parent;
+						}
```
nielsm5, line 114: Right now this loops through all elements, and then all parent elements of each element recursively. This makes this loop very inefficient, and very complex.

It would be better to loop through the `$scope.elements` once and filter out which elements should be returned.

nielsm5 (changes requested):

webapp/src/main/webapp/iaf/frankdoc/js/app.js
```diff
@@ -95,17 +95,33 @@ angular.module('iaf.frankdoc').config(['$stateProvider', '$urlRouterProvider', f
...
+			let element = groupMembers[i];
+			let obj = elements[element];
+			if(searchTextLC) {
+				if(JSON.stringify(obj).replace(/"/g, '').toLowerCase().indexOf(searchTextLC) > -1) {
 					r[element] = obj;
+				} else { // search in parent (accessing children is an expensive operation)
+					let elementParent = elements[element].parent;
+					while(elementParent) {
+						let parentObj = elements[elementParent];
+						if(matchedParents[elementParent]) { // if parent matched already leave the loop
```
gvanbrakel, line 110: You should not only leave the loop if the parent is already matched, but also if it is already *not* matched.
To do so you could also keep a list of unmatched parents

gvanbrakel, line 110: You should not only leave the loop if the parent is already matched, but also if it is already *not* matched.
To do so you could also keep a list of unmatched parents

gvanbrakel (changes requested):

gvanbrakel (approved):

nielsm5 (approved):

PR #2922, opened 2022-03-24
Fix warning attribute name must appear on XmlInputValidator

mhdirkse:


mhdirkse:
Should fix Frank!Doc issue https://github.com/ibissource/frank-doc/issues/91.

core/src/main/java/nl/nn/adapterframework/pipes/XmlValidator.java
```diff
@@ -492,6 +493,7 @@ public String getName() {
 		}
 
 		@IbisDoc({"name of the pipe", ""})
+		@Optional
```
gvanbrakel, line 496: This change should not be necessary, as the ` @Optional` annotation should be inherited from `IValidator`

gvanbrakel (changes requested):

gvanbrakel:
problem will be solved in frank doc generator

PR #2935, opened 2022-03-28
LCM upgrade jackson dependencies

alisihab:


nielsm5 (approved):

core/pom.xml
```diff
@@ -570,7 +570,7 @@
 		<dependency>
 			<groupId>com.fasterxml.jackson.core</groupId>
 			<artifactId>jackson-databind</artifactId>
-			<version>${jackson.version}</version>
+			<version>2.13.2.1</version> <!-- patch release -->
```
gvanbrakel, line 573: I think I prefer to have a property `jackson.databind.version`, to assure that future versions will keep in sync too

gvanbrakel (changes requested):

gvanbrakel (approved):

PR #2945, opened 2022-03-29
Log a warning in case a javalistener does not start in time limit

alisihab:


gvanbrakel:
I think that only a warning must be issued, and the adapter should be allowed to start

core/src/main/java/nl/nn/adapterframework/senders/IbisLocalSender.java
```diff
@@ -141,6 +141,9 @@ public void open() throws SenderException {
 						throw new SenderException(e);
 					}
 				}
+				if(loops == 0 && (listener==null || !listener.isOpen())) {
+					log.warn("Unable to open JavaListener ["+getJavaListener()+"] in "+getDependencyTimeOut()+" seconds. Make sure that the listener ["+getJavaListener()+"] exists or increase the timeout so that the sub-adapter may start before timeout limit.");
+				}
```
nielsm5, line 146: Perhaps this can be an adapter warning? That way it's more visible to the developer.

nielsm5 (approved):
Not possible to retrieve the MessageKeeper at this time.

PR #2967, opened 2022-04-01
Allow Exits and GlobalForwards in compatibility parse mode (7.7)

gvanbrakel:


nielsm5 (approved):

core/src/test/java/nl/nn/adapterframework/configuration/digester/ConfigurationDigesterTest.java
```diff
@@ -44,6 +48,24 @@ public void testNewCanonicalizer() throws Exception {
 		MatchUtils.assertXmlEquals(expected, result);
 	}
 
+	@Test
+	public void testSingleExitInContainerElementSchemaParsingDisabled() throws Exception {
```
nielsm5, line 52: I think you could also override `parseAndResolveEntitiesAndProperties` and then call the super with a different configurationResource. Then you wouldn't have to mock statics nor set the property in appconstants.

nielsm5 (changes requested):

nielsm5 (approved):

PR #2984, opened 2022-04-04
Fix test adapter names and ensure only bound to address when provided

nielsm5:
The WebServiceListener used to bind to both the listeners name (which is determined based on the receiver name) AND the ` address` attribute.


gvanbrakel (approved):

nielsm5:
Might fix https://github.com/ibissource/iaf/issues/2965

PR #2997, opened 2022-04-05
Remove Serializable from Message

gvanbrakel:


nielsm5 (approved):

gvanbrakel:
Serialization of Message is used in MessageWrapper, already for a long time. We cannot just disable it.

PR #3145, opened 2022-05-09
Fix running Jetty using Maven on the command prompt

mhdirkse:


mhdirkse:
I tried to run the F!F with Jetty as explained in CONTRIBUTING.md. It did not work. I asked Niels. He came up with the solution that is applied here. I tested that it worked.

nielsm5 (approved):

gvanbrakel (approved):

nielsm5:
Broke when I forgot to replace the jaas entry in the example pom in #2985.
Thanks Martijn!

PR #3181, opened 2022-05-16
Refactor fixedResultPipe

alisihab:


core/src/main/java/nl/nn/adapterframework/pipes/FixedResultPipe.java
```diff
@@ -131,57 +129,55 @@ public PipeRunResult doPipe(Message message, PipeLineSession session) throws Pip
...
-			URL xsltSource = ClassUtils.getResourceURL(this, styleSheetName);
+		if (StringUtils.isNotEmpty(getStyleSheetName())) {
+			URL xsltSource = ClassUtils.getResourceURL(this, getStyleSheetName());
 			if (xsltSource!=null) {
 				try{
-					String xsltResult = null;
 					Transformer transformer = XmlUtils.createTransformer(xsltSource);
-					xsltResult = XmlUtils.transformXml(transformer, result);
-					result = xsltResult;
+					result = XmlUtils.transformXml(transformer, result);
```
nielsm5, line 171: It would perform a whole lot better if you were to use a transformerpool  -hint hint- 😄

core/src/main/java/nl/nn/adapterframework/pipes/FixedResultPipe.java
```diff
@@ -131,57 +129,54 @@ public PipeRunResult doPipe(Message message, PipeLineSession session) throws Pip
...
+		if (isSubstituteVars()) {
+			result=StringResolver.substVars(result, session, appConstants);
 		}
 
-		if (StringUtils.isNotEmpty(styleSheetName)) {
-			URL xsltSource = ClassUtils.getResourceURL(this, styleSheetName);
+		if (transformerPool != null) {
+			URL xsltSource = ClassUtils.getResourceURL(this, getStyleSheetName());
 			if (xsltSource!=null) {
 				try{
```
gvanbrakel, line 169: This 'if' is redundant now

alisihab: xsltSource can still be null?

gvanbrakel (approved):

core/src/main/java/nl/nn/adapterframework/pipes/FixedResultPipe.java
```diff
@@ -283,16 +251,10 @@ public void setStyleSheetName (String styleSheetName){
 	public void setLookupAtRuntime(boolean b){
 		lookupAtRuntime=b;
 	}
```
nielsm5, line 253: I don't think this does anything? Perhaps the setter can be deprecated and the field can be removed.

alisihab: Used in configure to verify the existence of the file in configure time or in runtime.

nielsm5: Correct me if I'm wrong but it's a classpath file. It's not like people can write a file somewhere and for this pipe to use it.. The file has to be in the jar/war.

I would expect the resource retrieval in configure (and save the URL), and when using the filenameSessionkey it would do it runtime.

gvanbrakel: I think it is meant as a development-time option

nielsm5 (approved):

PR #3190, opened 2022-05-17
Create and accommodate configuration2mermaid.xsl and make changes to adapter2mermaid.xsl

lenardvandermaas:
Add an xsl that will create a mermaid configuration for Frank configurations. Java code has been changed to use this new xsl instead of the xsl that is used for adapters when the input doesn't start with `<adapter`. This change will prevent from generating detailed images that are too big to normally view when the input is the entire configuration instead of just an adapter, and instead generates a minimal overview of the listeners and senders and connections between adapters.

This pull request also contains a new version of adapter2mermaid.xsl which includes some small bug fixes, and support for new features such as:
- Exits with a name attribute instead of path
- No exits specified
- Showing extra info about adapters


tnleeuw (approved):

jkosternl (approved):
Test class MermaidFlowGeneratorTest.java is not the prettiest, but probably functionally Ok to continue and merge.

PR #3269, opened 2022-05-30
Throw exception when file to read does not exist in fileSystems (7.7)

### LINKED ISSUE #3266: LocalFileSystemPipe does not follow exception forward

**Describe the issue**
FilePipe in a frank is replaced by a localFilesystemPipe. In a larva scenario, it is expected to throw an exception since the file does not exists. However, when base64 attribute is not present in the pipe configuration pipeline does not follow exception forward. On the other hand, with base64 attribute being set to encode it follows exception forward.

**Reporter**
Laura

FF 7.7.1 not tested with the latest

alisihab:
Closes #3266

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/filesystem/FileSystemActor.java
+++ b/core/src/main/java/nl/nn/adapterframework/filesystem/FileSystemActor.java
@@ -336,6 +336,9 @@ public class FileSystemActor<F, FS extends IBasicFileSystem<F>> implements IOutp
 				}
 				case READ: {
 					F file=getFile(input, pvl);
+					if(!fileSystem.exists(file)) {
+						throw new FileSystemException("File to read does not exist.");
+					}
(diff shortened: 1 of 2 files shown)


nielsm5 (approved):

core/src/main/java/nl/nn/adapterframework/filesystem/FileSystemActor.java
```diff
@@ -336,6 +336,9 @@ public Object doAction(Message input, ParameterValueList pvl, PipeLineSession se
 				}
 				case READ: {
 					F file=getFile(input, pvl);
+					if(!fileSystem.exists(file)) {
+						throw new FileSystemException("File to read does not exist.");
```
gvanbrakel, line 340: It would be good to include the filename in the error message

gvanbrakel (changes requested):

gvanbrakel (approved):

PR #3411, opened 2022-06-23
Fix handling of pattern {fixedDate} when parameter type is set

gvanbrakel:


core/src/main/java/nl/nn/adapterframework/parameters/Parameter.java
```diff
@@ -839,8 +843,15 @@ private Object getValueForFormatting(ParameterValueList alreadyResolvedParameter
 				Object fixedDateTime = session.get(PutSystemDateInSession.FIXEDDATE_STUB4TESTTOOL_KEY);
 				if (fixedDateTime==null) {
 					fixedDateTime = PutSystemDateInSession.FIXEDDATETIME;
+					DateFormat df = new SimpleDateFormat(DateUtils.FORMAT_GENERICDATETIME);
+					try {
+						fixedDateTime = df.parse(Message.asString(fixedDateTime));
+					} catch (ParseException | IOException e) {
+						log.warn("Could not parse FIXEDDATETIME", e);
+						fixedDateTime = e;
```
nielsm5, line 851: is dit handig?

nielsm5 (approved):

PR #3539, opened 2022-07-27
Fix resumeTransaction always being forced when tx sync is inactive

nielsm5:
- Disable streaming xslt when attribute streamingActive is set to false
- Improve error handling when XsltSender cannot parse XML (and the ExceptionCatchingFilter is never called)


core/src/main/java/nl/nn/adapterframework/jta/TransactionConnectorCoordinator.java
```diff
@@ -128,21 +128,21 @@ public void suspendTransaction() {
...
 	}
-	
+
 	public void resumeTransaction() {
 		resumeTransaction(false);
 	}
 	public void resumeTransaction(boolean force) {
 		if (suspended || force) {
 			log.debug("resumeTransaction() resuming transaction of parent thread [{}], current thread [{}]", ()->parentThread.getName(), ()->Thread.currentThread().getName());
-			if (!force || !TransactionSynchronizationManager.isSynchronizationActive()) {
```
nielsm5, line 142: !?

gvanbrakel (approved):

PR #3542, opened 2022-08-03
Only detect Content-Type (and Charset) when type = DETECT

nielsm5:


core/src/main/java/nl/nn/adapterframework/http/rest/MediaTypes.java
```diff
@@ -16,23 +16,30 @@
...
+import org.springframework.util.MimeType;
+import org.springframework.util.MimeTypeUtils;
 
 import nl.nn.adapterframework.util.StreamUtil;
 import nl.nn.adapterframework.util.EnumUtils;
 
 public enum MediaTypes {
 
 	ANY("*/*", null),
+	DETECT("*/*"),
```
nielsm5, line 32:
```suggestion
	/** (Only for produces) Attempts to detect the MimeType as well as charset when not known */
	DETECT("*/*"),
```

gvanbrakel (approved):

PR #3923, opened 2022-10-26
Provide handlers to converted documents after conversion to PDF

Laurens-makel:
In the current situation, all converted files are written to the filesystem after conversion.

Ideally this should not happen because it requires additional steps to read the contents of the converted documents into session. 

This PR is an attempt to phase out the usage of the FS in the Aspose module.

Currently 10 tests are failing, when checking these by hand, it seems that the conversion is succesfull but some things are a little bit different.

Changes:
- Write file contents to session instead of filesystem for further direct usage.


gvanbrakel (changes requested):

aspose/src/main/java/nl/nn/adapterframework/extensions/aspose/pipe/PdfPipe.java
```diff
@@ -164,6 +173,21 @@ public PipeRunResult doPipe(Message input, PipeLineSession session) throws PipeR
...
+		if(StringUtils.isNotEmpty(result.getResultFilePath())){
+			FileMessage document = new FileMessage(new File(result.getResultFilePath()));
+			message.getContext().put(CONVERTED_DOCUMENTS_CONTEXT_KEY+index, document);
+		}
+
+		List<CisConversionResult> attachmentList = result.getAttachments();
+		if (attachmentList != null && !attachmentList.isEmpty()) {
+			for (int i = 0; i < attachmentList.size(); i++) {
+				index++;
+				populateContext(attachmentList.get(i), message, index);
```
tnleeuw, line 186: @Laurens-makel Did you test this recursion? It looks to me like this will not correctly update the index for all nested documents.

aspose/src/main/java/nl/nn/adapterframework/extensions/aspose/pipe/PdfPipe.java
```diff
@@ -164,6 +173,21 @@ public PipeRunResult doPipe(Message input, PipeLineSession session) throws PipeR
 		}
 	}
 
+	private void populateContext(CisConversionResult result, Message message, int index) {
+		if(StringUtils.isNotEmpty(result.getResultFilePath())){
+			FileMessage document = new FileMessage(new File(result.getResultFilePath()));
+			message.getContext().put(CONVERTED_DOCUMENTS_CONTEXT_KEY+index, document);
```
tnleeuw, line 179: I'm not sure that the message-context is the best place to put these documents? Rather put them into the session?

aspose/src/main/java/nl/nn/adapterframework/extensions/aspose/pipe/PdfPipe.java
```diff
@@ -154,8 +158,13 @@ public PipeRunResult doPipe(Message input, PipeLineSession session) throws PipeR
 					XmlBuilder main = new XmlBuilder("main");
 					cisConversionResult.buildXmlFromResult(main, true);
 
-					session.put("documents", main.toXML());
-					return new PipeRunResult(getSuccessForward(), main.toXML());
+					Message message = new Message(main.toXML());
+					populateContext(cisConversionResult, message, 0);
+
+					session.put("documents", message);
```
tnleeuw, line 164: It would be nice to have this session-key configurable.

aspose/src/main/java/nl/nn/adapterframework/extensions/aspose/pipe/PdfPipe.java
tnleeuw: I've done some debugging both in my branch and in your branch and I never see an attachment that has a file that is written out.
That makes me concerned whether we are doing this right, or if the Aspose module is actually broken and we never noticed because our tests are too dumb.

aspose/src/main/java/nl/nn/adapterframework/extensions/aspose/pipe/PdfPipe.java
tnleeuw: Actually on further review why not use the already existing configuration-option `mainDocumentSessionKey` which is used to store the result of combining PDFs?

aspose/src/main/java/nl/nn/adapterframework/extensions/aspose/pipe/PdfPipe.java
tnleeuw: Update: I now see that we have an option to save separate PDFs but that currently we never test this. 🙄 
That explains a lot.

nielsm5:
Superseded by #5640.

PR #4008, opened 2022-11-16
Fix returning inputstream session keys

gvanbrakel:


core/src/main/java/nl/nn/adapterframework/receivers/JavaListener.java
```diff
@@ -283,31 +257,27 @@ public void setLocal(String name) {
 
 	@Deprecated
 	public void setIsolated(boolean b) {
-		isolated = b;
+		throw new RuntimeException("function of attribute 'isolated' is replaced by 'transactionAttribute' on PipeLine");
```
nielsm5, line 260: IllegalStateException, NotImplementedException of IllegalArgumentException?

nielsm5 (changes requested):

gvanbrakel:
Er zijn nog iaf-test failures

nielsm5 (approved):

PR #4035, opened 2022-11-19
Fix and optimize Sonatype Lift builds

nielsm5:
Big thanks to @mattjohnson for helping out!


philipsens (approved):

mattjohnson:
It looks like it still failed at the end 😞 The good news is it has gotten past all of the build stages, and the tools all run. I'll make sure to get this fix next week.

Issue #4048, opened 2022-11-22
Browser does not set 'If-None-Match' headers because of 'Cache-Control: no-store, no-cache' 


Laurens-makel:
**❗ please do not add sensitive information in issues, you can provide extra information via email using issue number as reference ❗**

**Describe the issue**
When using the `updateEtag=true` attribute on a ApiListener, the listener will generate an etag value over the response of the adapter and will set the value to response header 'ETAG'.  A consumer should use this ETAG value in a 'If-None-Match' header in subsequent requests to the same URL to prevent the backend calculating the same response in case the etag values match.

Although my browser refuses to send a 'If-None-Match'  because in the first call to the adapter, another response header named 'Cache-Control: no-store, no-cache' is present, which enforces the browser to not apply any form of caching and retrieve the same response again.

https://stackoverflow.com/questions/15900548/why-browser-does-not-send-if-none-match-header
![image](https://user-images.githubusercontent.com/37303445/203265410-4a79c326-25bd-4941-8f55-c4133ce592d2.png)

This header seems to be added to the response by the CacheControlFilter in the web.xml
https://github.com/ibissource/iaf/blob/master/webapp/src/main/webapp/WEB-INF/web.xml
![image](https://user-images.githubusercontent.com/37303445/203265812-21454918-a18a-4934-9c5d-30725cc25c1f.png)

https://github.com/ibissource/iaf/blob/master/core/src/main/java/nl/nn/adapterframework/http/CacheControlFilter.java
![image](https://user-images.githubusercontent.com/37303445/203266026-9be605c3-8a2b-4eb8-bded-7f48e536a51e.png)

I'd expect that when an ETAG header is present, the Cache-Control should not be set to 'no-store, no-cache'.

**Reporter**
Laurens

PR #4050, opened 2022-11-22
Prevent setting cache control headers when etag header has been set

Laurens-makel:
Partially solves #4048


core/src/main/java/nl/nn/adapterframework/http/CacheControlFilter.java
```diff
@@ -49,9 +49,15 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
 		HttpServletRequest req = (HttpServletRequest) request;
 		//resp.setHeader("Expires", "Tue, 03 Jul 2001 06:00:00 GMT");
 		resp.setDateHeader("Last-Modified", new Date().getTime());
```
nielsm5, line 51: Moet deze niet in de if komen?

PR #4300, opened 2023-01-06
Resolve values

mhdirkse:


mhdirkse:
Please merge https://github.com/ibissource/frank-doc/pull/146 before this PR.

gvanbrakel (approved):

mhdirkse:
I discussed with @nielsm5. It seems that all {@value ...} occurrences refer to public Java elements. If this is the case, there is no need to merge this PR. The Frank!Doc PR mentioned above modifies the Frank!Doc such that it can work with private Java elements. If there appear to be references to private constants, we can merge this PR.

mhdirkse:
Indeed, no private fields are needed.

mhdirkse:
We need a private Java element to have attribute Adapter.msgLogLevel in the Frank!Doc. Therefore we do need to merge this PR. @gvanbrakel, we discussed about the enum type that is the argument of the attribute setter. The attribute could be included by making that enum public, but that is not the solution because then any place in the Java code can access the enum.

gvanbrakel (approved):

nielsm5:
ibissource/frank-doc#146 has been merged :)

PR #4324, opened 2023-01-11
Make no-docker profile for appservers and use properties for dbms hostname in iaf/test

philipsens:
Added a profile to each appserver pom to disable the Docker build. This way we can use Kaniko instead.
Changed the tag of the base appserver for Tomcat, so Kaniko can push and pull it. Users can still build it locally but won't be able to push it.
Changed all iaf/test database hostnames to a property, so a different one can be used in K8s.


docker/appserver/JBoss/pom.xml
```diff
@@ -183,4 +183,23 @@
...
+							<execution>
+								<id>prepare-package</id>
+								<phase>none</phase>
+							</execution>
+						</executions>
+					</plugin>
+				</plugins>
+			</build>
+		</profile>
+	</profiles>
```
nielsm5, line 204: zie https://github.com/ibissource/iaf/blob/master/example/pom.xml#L138

VBPham (approved):

PR #4357, opened 2023-01-17
SQL injections warning not suppressed

tnleeuw:
This change fixes the SQL Injection Warning that is triggered from the Security Items page; I have tested this using IAF Example (with and without the change) however I'm not sure if adding a unit-test is worth the effort?

NB: The bug only applies to 7.8 and earlier. It does not apply to 7.9, as the code for the Security Items has changed and no longer triggers this warning.

### DIFF
--- a/copyrightExcludeFile.txt
+++ b/copyrightExcludeFile.txt
@@ -4,6 +4,8 @@ src/test/
 src\test\
 webapp/
 webapp\
+overlays/
+overlays\
 docker/
 docker\
 .metadata/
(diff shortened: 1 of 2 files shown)


gvanbrakel (commented):
The issue states: '_On Localhost I don't see the warning, but in DEV/TEST it's present._' It might be good to check the cause of that.

tnleeuw:
> The issue states: '_On Localhost I don't see the warning, but in DEV/TEST it's present._' It might be good to check the cause of that.

This could simply be caused by someone having looked at the Security Items page on DEV/TEST before. As far as I could see in my local tests, the warning did not appear until I had opened that page.

Otherwise the warning could come from other adapters that are deployed on the same DEV/TEST environment, but I don't know if there would be any. Other adapters would need to have the `DirectQuerySender` be part of a multi-sender like `ParallelSender` or `ShadowSender`.

gvanbrakel (approved):

nielsm5 (approved):

PR #4386, opened 2023-01-20
When searching forward find exit too (7.8) partial backport of #3727

gvanbrakel:
Partial backport of #3727


nielsm5 (commented):
Wanneer treed dit op? Kan je de betreffende iaf-test scenario's ook backporten?

nielsm5 (approved):

PR #4412, opened 2023-01-25
Move SCP Filter to the console module

nielsm5:


console/src/main/java/nl/nn/adapterframework/webcontrol/CspFilter.java
```diff
@@ -35,7 +35,7 @@
 
 public class CspFilter implements Filter {
 	private ContentSecurityPolicyHeaderWriter cspWriter;
-	private static final boolean REPORT_ONLY = AppConstants.getInstance().getBoolean("cspheader.reportOnly", false);
+	private static final boolean REPORT_ONLY = AppConstants.getInstance().getBoolean("cspheader.reportOnly", true);
```
nielsm5, line 38: Right now the use of the SCP header breaks the upload file functionality.

gvanbrakel (approved):

PR #4414, opened 2023-01-25
fix: locally building tomcat docker images

philipsens:
I changed the base image for the tomcat images to use the nexus URL by default. When building with maven, the images will use the other base image.

Result when running `mvn clean install`

```console
REPOSITORY           TAG           IMAGE ID       CREATED          SIZE
iaf-example-tomcat   latest        64c12a1ec5ea   9 seconds ago    614MB
iaf-webapp-tomcat    latest        e7c8192b4ad6   10 seconds ago   584MB
iaf-test-tomcat      latest        94b8ad0924eb   15 seconds ago   563MB
iaf-as-tomcat        latest        e53be137bab6   18 seconds ago   382MB
tomcat               8.5.83-jdk8   fad50471d40b   2 months ago     336MB
```


philipsens:
The naming scheme of the tags is a bit confusing and not coherent. "iaf-as-tomcat" could better be "tomcat-base" or something in that direction.
"iaf-webapp-tomcat" should be "frank-framework" to reflect the name on nexus.

Changing the base image to a variable is done, so the local names can be used, while Kaniko (dockerless daemon in CI) can use the published variants. The tags for local use are defined in the pom.xml, so everything should go accordingly while running `mvn clean install`. 

The docker images could also all be tagged with the same name as the published variant. For example: "iaf-as-tomcat" would be tagged as "nexus.frankframework.org/private/ff-test/as/tomcat". This could cause confusion because you can't be certain if the image used to run a container is a locally build one or a published one.

philipsens:
I'm reorganizing all the docker images by the request of @nielsm5

philipsens:
#4440 fixes this issue.

PR #4424, opened 2023-01-27
Fix CSP header blocking file uploads

nielsm5:


gvanbrakel (commented):
Is dit 'Fix CSP header' of toch meer 'Fix file upload'?

gvanbrakel (approved):

PR #4474, opened 2023-02-08
Enable message preserve for AmazonS3FileSystem

gvanbrakel:


aws/src/main/java/nl/nn/adapterframework/filesystem/AmazonS3FileSystem.java
```diff
@@ -464,29 +479,18 @@ public Date getModificationTime(S3Object f) throws FileSystemException {
...
-//	 *            The name of the bucket where the file is stored in.
-//	 * @param fileName
-//	 * 			  The name of the file that is processed.
-//	 */
-//	public void fileDoesNotExist(String bucketName, String fileName) throws SenderException {
-//		if (!s3Client.doesObjectExist(bucketName, fileName))
-//			throw new SenderException(" file with fileName [" + fileName + "] does not exist, please specify the name of an existing file");
-//	}
+
+	public ClientConfiguration getProxyConfig() {
```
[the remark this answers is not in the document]
gvanbrakel: It is not a comment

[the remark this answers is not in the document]
gvanbrakel: it is not a comment

aws/src/main/java/nl/nn/adapterframework/filesystem/AmazonS3FileSystem.java
```diff
@@ -120,13 +121,20 @@ public void close() throws FileSystemException {
...
 		S3Object object = new S3Object();
-		object.setKey(filename);
+		int separatorPos = filename.indexOf(BUCKET_OBJECT_SEPARATOR);
+		if (separatorPos<0) {
+			object.setBucketName(bucketName);
+			object.setKey(filename);
+		} else {
+			object.setBucketName(filename.substring(0,separatorPos));
+			object.setKey(filename.substring(+1));
+		}
```
nielsm5, line 131: Hier zit nog een foutje

nielsm5 (changes requested):

nielsm5 (approved):

test-integration/src/test/java/nl/nn/adapterframework/filesystem/AmazonS3FileSystemTestHelper.java
```diff
@@ -56,38 +60,40 @@ public AmazonS3FileSystemTestHelper(String accessKey, String secretKey, boolean
...
+				// might not return all of the objects in the first listing. Check to
+				// see whether the listing was truncated. If so, retrieve the next page of objects 
+				// and delete them.
+				if (objectListing.isTruncated()) {
+					objectListing = s3Client.listNextBatchOfObjects(objectListing);
+				} else {
+					break;
+				}
+			}
+			//s3Client.deleteBucket(bucketName);
```
tnleeuw, line 96: Why is the bucket no longer deleted?

PS: Perhaps it would be cleaner to create the bucket in a `@BeforeAll` setup method and only delete it in a `@AfterAll` cleanup method instead of creating/deleting it with every single test?
And then reuse the bucket for all tests.
We would need to be sure of course, that bucket-name does not change between tests.

test-integration/src/test/java/nl/nn/adapterframework/filesystem/AmazonS3FileSystemTestHelper.java
```diff
@@ -183,7 +188,23 @@ public boolean _folderExists(String folderName) throws Exception {
...
+			// If the bucket contains many objects, the listObjects() call
+			// might not return all of the objects in the first listing. Check to
+			// see whether the listing was truncated. If so, retrieve the next page of objects 
+			// and delete them.
+			if (objectListing.isTruncated()) {
+				objectListing = s3Client.listNextBatchOfObjects(objectListing);
+			} else {
+				break;
+			}
+		}
```
tnleeuw, line 207: This whole loop seems duplicated with the teardown-method. Perhaps a good idea to extract the duplicate code into a method.

test-integration/src/test/java/nl/nn/adapterframework/filesystem/SelfContainedBasicFileSystemTest.java
```diff
@@ -59,30 +59,30 @@ public void testFiles()  throws Exception{
...
 		try(DirectoryStream<F> ds = fileSystem.listFiles(sourceOfMessages_folder)) {
 			Iterator<F> it = ds.iterator();
-			assertTrue("there must be at least one messsage in the sourceOfMessages_folder ["+sourceOfMessages_folder+"]", it!=null && it.hasNext());
+			assertTrue(it!=null && it.hasNext(), "there must be at least one messsage in the sourceOfMessages_folder ["+sourceOfMessages_folder+"]");
 
 			sourceFile =  it.next();
-			assertTrue("file retrieved from folder should exist", fileSystem.exists(sourceFile));
-			//assertFalse("name of source file should not appear in just created folder", fileSystem.filenameExistsInFolder(folderName, fileSystem.getName(sourceFile)));
+			assertTrue(fileSystem.exists(sourceFile), "file retrieved from folder should exist");
+			//assertFalse(fileSystem.filenameExistsInFolder(folderName, fileSystem.getName(sourceFile)), "name of source file should not appear in just created folder");
```
tnleeuw, line 72: Why do we keep these commented-out assertions? (Here, and below). Shouldn't we just remove those?

tnleeuw (changes requested):

core/src/test/java/nl/nn/adapterframework/filesystem/BasicFileSystemTest.java
```diff
@@ -139,6 +142,31 @@ public void basicFileSystemTestRead() throws Exception {
...
+		// test
+		existsCheck(filename);
+
+		F file = fileSystem.toFile(filename);
+		// test
+		Message in = fileSystem.readFile(file, null);
+
+		in.preserve();
+
+		String actual1 = StreamUtil.readerToString(in.asReader(), null);
```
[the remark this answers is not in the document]
gvanbrakel: It tests if message can actually be read multiple times, without turning it explicitly into a String or byte array.
This will fail if a message declared that it was repeatable, but actually was not repeatable.

test-integration/src/test/java/nl/nn/adapterframework/filesystem/AmazonS3FileSystemTestHelper.java
```diff
@@ -56,38 +60,40 @@ public AmazonS3FileSystemTestHelper(String accessKey, String secretKey, boolean
 	}
 
 	@Override
-	@Before
+	@BeforeAll
```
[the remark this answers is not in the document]
gvanbrakel: Actually the Before and After annotations do not appear to work anymore if not in a test-class. They are now called from these test classes.

test-integration/src/test/java/nl/nn/adapterframework/filesystem/AmazonS3FileSystemTestHelper.java
gvanbrakel: Because it no longer has a random name

PR #4480, opened 2023-02-09
Fix Appconstants property parsing to use the correct file encoding

nielsm5:


tnleeuw (approved):

core/src/main/java/nl/nn/adapterframework/configuration/BuildInfoValidator.java
```diff
@@ -69,7 +74,10 @@ private void read() throws IOException, ConfigurationException {
 					if(buildInfoFilename.equals(fileName)) {
 						name = FilenameUtils.getPathNoEndSeparator(entryName);
 						Properties props = new Properties();
-						props.load(zipInputStream);
+						try(Reader reader = StreamUtil.getCharsetDetectingInputStreamReader(zipInputStream)) {
+							props.load(reader);
+							LOG.info("Application constants loaded from url []");
```
nielsm5, line 79:
```suggestion
							LOG.info("properties loaded from archive, filename [{}]", name);
```

PR #4481, opened 2023-02-09
Make connection pools configurable

gvanbrakel:


core/src/main/java/nl/nn/adapterframework/jdbc/JdbcFacade.java
```diff
@@ -123,7 +123,7 @@ protected DataSource getDatasource() throws JdbcException {
 	public String getDatasourceInfo() throws JdbcException {
 		String dsinfo=null;
 		if(getDatasource() instanceof TransactionalDbmsSupportAwareDataSourceProxy) {
-			return ((TransactionalDbmsSupportAwareDataSourceProxy) getDatasource()).getInfo();
+			return ((TransactionalDbmsSupportAwareDataSourceProxy) getDatasource()).toString();
```
nielsm5, line 126: Dit is niet handig, de tostring geeft meer/minder data terug afhankelijk van het log level.

core/src/main/java/nl/nn/adapterframework/jta/btm/BtmDataSourceFactory.java
```diff
@@ -23,16 +23,26 @@
...
+import lombok.Getter;
+import lombok.Setter;
 import nl.nn.adapterframework.jndi.JndiDataSourceFactory;
 
 public class BtmDataSourceFactory extends JndiDataSourceFactory implements DisposableBean {
 
+	private @Getter @Setter int minPoolSize=0;
+	private @Getter @Setter int maxPoolSize=20;
+	private @Getter @Setter int maxIdleTime=60;
+	private @Getter @Setter int maxLifeTime=0;
```
nielsm5, line 35: Is het niet mooier hier `AppConstants.getInstance().getProperty(...);` van te maken, dan blijft de spring configuration xml kleiner, maar is het toch instelbaar?

nielsm5 (approved):

nielsm5 (approved):

core/src/main/java/nl/nn/adapterframework/jta/narayana/NarayanaDataSourceFactory.java
```diff
@@ -29,21 +29,31 @@
...
 public class NarayanaDataSourceFactory extends JndiDataSourceFactory {
 	protected static Logger log = LogUtil.getLogger(NarayanaDataSourceFactory.class);
 
+	private @Getter @Setter int maxPoolSize=20;
+
 	private @Setter NarayanaJtaTransactionManager transactionManager;
 
+	public NarayanaDataSourceFactory() {
+		AppConstants appConstants = AppConstants.getInstance();
+		maxPoolSize = appConstants.getInt(MIN_POOL_SIZE_PROPERTY, maxPoolSize);
```
tnleeuw, line 47: Is this intentional, or a typo?
If intentional, it would be good to explain in a comment.

If a typo, can we please have unit tests around these things, with the actual implementations and not just an IAF-test with some mocks? Might help catch typos like this.

core/src/main/java/nl/nn/adapterframework/jta/narayana/NarayanaDataSourceFactory.java
```diff
@@ -29,21 +29,31 @@
...
 import nl.nn.adapterframework.util.LogUtil;
 
 public class NarayanaDataSourceFactory extends JndiDataSourceFactory {
 	protected static Logger log = LogUtil.getLogger(NarayanaDataSourceFactory.class);
 
+	private @Getter @Setter int maxPoolSize=20;
+
 	private @Setter NarayanaJtaTransactionManager transactionManager;
 
+	public NarayanaDataSourceFactory() {
```
tnleeuw, line 45: Is it possible to override `toString()` for this class as well and print out the actual pool size? Or is that not supported by Narayana?

tnleeuw:
This might be my ignorance, but does the `NarayanaConnectionFactoryFactory` also need to be updated to allow configuration of the pool size?

tnleeuw (changes requested):

tnleeuw:
NB: I presume we need to also add this to documentation somewhere?

tnleeuw (approved):

tnleeuw (approved):

tnleeuw:
Now that we know the problematic application is using Narayana transaction manager, I think we should add JMS connection pooling to the Narayana Connection Factory as well.

tnleeuw (changes requested):
Now that we know the problematic application is using Narayana transaction manager, I think we should add JMS connection pooling to the Narayana Connection Factory as well.

nielsm5:
> Now that we know the problematic application is using Narayana transaction manager, I think we should add JMS connection pooling to the Narayana Connection Factory as well.

I completely agree which is why I created issue #4493. Shall we merge this first?

tnleeuw:
> > Now that we know the problematic application is using Narayana transaction manager, I think we should add JMS connection pooling to the Narayana Connection Factory as well.
> 
> I completely agree which is why I created issue #4493. Shall we merge this first?

OK, should #4493 then be added to release 7.8 as well?

tnleeuw (approved):
Approved given that adding JMS Connection Pooling to Narayana TX Manager will be solved in a different ticket.

PR #4485, opened 2023-02-09
Document ignoreUnknownNamespaces for elementFormDefault=unqualified

gvanbrakel:


core/src/main/java/nl/nn/adapterframework/validation/XSD.java
```diff
@@ -250,6 +250,7 @@ public int compareToByContents(XSD x) {
 			if (diff.similar()) {
 				return 0;
 			}
+			// TODO: check necessity of this compare. If Diff says they are different, is it useful to check again for the plain contents?
```
tnleeuw, line 253: Good question. If we do a full string compare, why do we do a diff first?

What would it take to answer the question and settle this TODO now, instead of postponing this for the future?

nielsm5: Sounds like something we could create a few tests for. Also because the Diff dependency is a bit iffy and this is the only place (I believe) we use it.

nielsm5 (approved):

core/src/main/java/nl/nn/adapterframework/validation/XSD.java
nielsm5: Needs to be fixed, a new issue has been created to resolve this (#4497).

tnleeuw (approved):

PR #4498, opened 2023-02-14
Fix imports from the credential-provider project

tnleeuw:
Fix imports of `Misc` that were importing from the `credential-provider` module to instead import the version in `core` module.

NB: Moving `Misc` and some other util classes to a separate Maven module will be another PR, first I want to get the imports right.


nielsm5 (approved):
Some of these imports have already been corrected in #4481 I think?

tnleeuw:
> Some of these imports have already been corrected in #4481 I think?

That has already been merged, right? I had some imports from my shelved changes give conflicts with master due to them already having been fixed (but not sorted) so I think it all should be in order.

gvanbrakel (approved):

PR #4517, opened 2023-02-16
Ensure connections are closed properly when using Narayana

nielsm5:


core/src/main/java/nl/nn/adapterframework/jta/narayana/NarayanaTransactionHelper.java
```diff
@@ -0,0 +1,63 @@
...
+
+	private final TransactionManager transactionManager;
+
+	public NarayanaTransactionHelper(TransactionManager transactionManager) {
+		super(transactionManager);
+		this.transactionManager = transactionManager;
+	}
+
+	/**
+	 * Connections are not always closed. {@link ConnectionProxy#close() } and {@link SessionProxy#close() }
```
gvanbrakel, line 40:
```suggestion
	 * Connections were not always closed, because the super implementation of this method returns false too often. {@link ConnectionProxy#close() } and {@link SessionProxy#close() } 
```

gvanbrakel (changes requested):

core/src/main/java/nl/nn/adapterframework/jta/btm/BtmConnectionFactoryFactory.java
```diff
@@ -43,7 +43,7 @@ public BtmConnectionFactoryFactory() {
 	}
 
 	@Override
-	protected ConnectionFactory augment(ConnectionFactory connectionFactory, String connectionFactoryName) {
+	protected ConnectionFactory augmentConnectionFactory(ConnectionFactory connectionFactory, String connectionFactoryName) {
```
tnleeuw, line 46: Why does overriding `augment` not work? Why instead is it made `final` in the super-class and delegating to this new method that is overridden here?

tnleeuw (changes requested):

core/src/main/java/nl/nn/adapterframework/jta/btm/BtmConnectionFactoryFactory.java
tnleeuw: Sounds familiar 🙈 

😅

tnleeuw (approved):

gvanbrakel (approved):

PR #4552, opened 2023-02-21
Improve suppressability of XSD warnings

gvanbrakel:


core/src/main/java/nl/nn/adapterframework/validation/XercesValidationErrorHandler.java
```diff
@@ -0,0 +1,68 @@
...
+				ConfigurationWarnings.add(source, log, e.toString(), suppressKey);
+			}
+		} else {
+			ConfigurationWarnings.add(source, log, e.toString());
+		}
+
+		// In case the XSD doesn't exist throw an exception to prevent the adapter from starting.
+		if (e.getMessage() != null && e.getMessage().startsWith("schema_reference.4: Failed to read schema document '")) {
+			throw e;
+		}
```
tnleeuw, line 66: I feel it would be cleaner to move this to a separate method, separating out the logging from the "potentially throwing".

Or perhaps rename the method to something a bit higher level, like `handleError(..)`, instead of describing the implementation ( `warnAndPotentiallyThrow(...)` )?

tnleeuw (approved):

core/src/main/java/nl/nn/adapterframework/validation/XercesValidationErrorHandler.java
```diff
@@ -0,0 +1,68 @@
...
+		throw e;
+	}
+
+	private void handleException(XMLParseException e, SuppressKeys suppressKey) {
+		if (suppressKey!=null) {
+			if (allowConsoleWarnings) {
+				ConfigurationWarnings.add(source, log, e.toString(), suppressKey);
+			}
+		} else {
+			ConfigurationWarnings.add(source, log, e.toString());
```
tnleeuw, line 60: Should `allowConsoleWarnings` still be ignored when no `suppressKey` is passed?

tnleeuw (approved):

core/src/main/java/nl/nn/adapterframework/validation/XercesValidationErrorHandler.java
gvanbrakel: When no suppressKey is passed, the warnings or errors should apparently not be suppressable. 
You could simplify matters by removing the 'allowConsoleWarnings' feature all together.

tnleeuw: I don't see that in any requirements nor in the discussion in the ticket so that surprises me? But it seems that there is no more call where `suppressKeys` is `null` anyway so then it doesn't matter.

nielsm5 (approved):

PR #4579, opened 2023-03-02
Fixing build with maven 3.9.0

tnleeuw:
Automatic update on Github to maven 3.9.0 broke the build.

Try some ways to get this fixed.


tnleeuw:
It appears that the combination of Frank!Doc and Maven 3.9.0 is broken.

Using the Maven wrapper to specify which Maven version is used, will get around this problem for now until we fix the Frank!Doc plugin.

nielsm5 (approved):

PR #4596, opened 2023-03-06
Refactor code that uses Commons Upload dependency

nielsm5:


tnleeuw:
On the face of it this looks good but I see 0 test coverage on this class.
(And for some reason I do not see CodeCov in the checks on this PR so I had to run this locally to check test-coverage ).

tnleeuw (approved):

PR #4637, opened 2023-03-15
Fix Http (Multipart-) Message content length

nielsm5:


core/src/main/java/nl/nn/adapterframework/stream/Message.java
```diff
@@ -877,6 +874,10 @@ public long size() {
 			return (long) context.get(MessageContext.METADATA_SIZE);
 		}
 
+		if (request instanceof String) {
+			return ((String) request).length();
+		}
```
nielsm5, line 879: If the length is available in the MessagContext, that should be used

core/src/test/java/nl/nn/adapterframework/http/HttpMessageEntityTest.java
```diff
@@ -27,12 +31,12 @@ public class HttpMessageEntityTest {
 	private static Message NON_REPEATABLE_MESSAGE;
 	private static Message REPEATABLE_TS_MESSAGE;
 
-	@BeforeClass
+	@BeforeAll
```
tnleeuw, line 34: I know you did not write this test, but I don't like the way this test is set up with `@BeforeAll` instead of regular non-static `@Before`. There are things inside message class that change its state when accessing streams and that means that potentially, each test is executed with a slightly different state of the `Message` instance(s).

core/src/main/java/nl/nn/adapterframework/stream/Message.java
tnleeuw: If the size of the message equal to the length of the string? 🤔
In which places does the difference matter between binary size, and character length?

core/src/main/java/nl/nn/adapterframework/stream/Message.java
```diff
@@ -877,6 +874,10 @@ public long size() {
 			return (long) context.get(MessageContext.METADATA_SIZE);
```
tnleeuw, line 874: Cannot this check be moved further up, above the check `if (request instance FileInputStream)`? That's a potentially expensive check when the same is already in the `MessageContext`.
Also, any reason why a size once calculated is not cached in the `MessageContext`?

tnleeuw (changes requested):

core/src/main/java/nl/nn/adapterframework/stream/Message.java
nielsm5: I've updated the documentation!

tnleeuw (approved):

PR #4667, opened 2023-03-22
Don't pollute log file when Console cannot determine errorstore count

nielsm5:


core/src/main/java/nl/nn/adapterframework/webcontrol/api/ShowConfigurationStatus.java
```diff
@@ -719,15 +719,17 @@ private Map<String, Object> mapAdapter(Adapter adapter) {
...
 				}
 			}
 			IMessageBrowser<?> mlmb = rcv.getMessageBrowser(ProcessState.DONE);
 			if(mlmb != null) {
 				try {
 					messageLogMessageCount += mlmb.getMessageCount();
 				} catch (ListenerException e) {
-					log.warn("Cannot determine number of messages in messagelog of ["+rcv.getName()+"]", e);
+					if(log.isInfoEnabled()) log.warn("Cannot determine number of messages in errorstore of [{}]", rcv.getName(), e);
+					else log.warn("Cannot determine number of messages in errorstore of [{}]: {}", rcv::getName, e::getMessage);
```
tnleeuw, line 732: I had to look 3x at this to figure out the difference between the 2 cases and figure out why you log a warn when info is enabled.

I think it would be helpful to add a comment to the code to explain.

Perhaps another idea is to keep a flag if stacktrace should be logged or not, set to not log stacktraces after first error and perhaps log again after message count has been successfully loaded again.

tnleeuw (approved):

core/src/main/java/nl/nn/adapterframework/webcontrol/api/ShowConfigurationStatus.java
```diff
@@ -719,15 +719,17 @@ private Map<String, Object> mapAdapter(Adapter adapter) {
 				try {
 					errorStoreMessageCount += esmb.getMessageCount();
 				} catch (ListenerException e) {
```
nielsm5, line 721:
```suggestion
				} catch (ListenerException e) {
						// Only log the stacktrace when loglevel == INFO. Otherwise it will pollute the log too much.
```

core/src/main/java/nl/nn/adapterframework/webcontrol/api/ShowConfigurationStatus.java
```diff
@@ -719,15 +719,17 @@ private Map<String, Object> mapAdapter(Adapter adapter) {
...
-					log.warn("Cannot determine number of messages in errorstore of ["+rcv.getName()+"]", e);
+					if(log.isInfoEnabled()) log.warn("Cannot determine number of messages in errorstore of [{}]", rcv.getName(), e);
+					else log.warn("Cannot determine number of messages in errorstore of [{}]: {}", rcv::getName, e::getMessage);
 				}
 			}
 			IMessageBrowser<?> mlmb = rcv.getMessageBrowser(ProcessState.DONE);
 			if(mlmb != null) {
 				try {
 					messageLogMessageCount += mlmb.getMessageCount();
 				} catch (ListenerException e) {
```
nielsm5, line 730:
```suggestion
				} catch (ListenerException e) {
						// Only log the stacktrace when loglevel == INFO. Otherwise it will pollute the log too much.
```

tnleeuw (approved):

PR #4673, opened 2023-03-23
Create bean to validate database connectivity

nielsm5:


tnleeuw:
Why do you not need the same dependency-excludes in the 7.8 branch?

tnleeuw (approved):

PR #4697, opened 2023-03-27
When standard port omit it in host header

jacodg:


nielsm5 (approved):

tnleeuw (approved):

tnleeuw:
@nielsm5 The change itself is good but do we perhaps want to see a unit test added for this kind of behaviour?

tnleeuw (approved):

PR #4701, opened 2023-03-28
Create Console WAR module

nielsm5:
Add SpringBoot runner for the Console backend


management-gateway/src/main/java/nl/nn/adapterframework/management/bus/BusMessageUtils.java
```diff
@@ -39,55 +41,68 @@ public class BusMessageUtils {
 
 	private static final Logger LOG = LogManager.getLogger(BusMessageUtils.class);
 
-	public static String getHeader(Message<?> message, String headerName) {
+	public static final String HEADER_PREFIX = "meta-";
+	public static final String HEADER_PREFIX_PATTERN = "meta-*";
```
tnleeuw, line 45: Isn't it customary to start custom header names with "X-"? So the prefix would be `X-meta-`?

nielsm5: The IETF guidance to prefix X- to experimental headers or to other extension field types has been deprecated since June 2012, per [RFC 6648](https://tools.ietf.org/html/rfc6648).

tnleeuw: Didn't know that!

console/backend/src/main/java/nl/nn/adapterframework/management/web/TestPipeline.java
```diff
@@ -117,6 +117,8 @@ private String getPayload(Message<?> response) {
 		Object payload = response.getPayload();
 		if(payload instanceof String) {
 			return (String) payload;
+		} else if(payload instanceof byte[]) {
+			return new String((byte[]) payload);
```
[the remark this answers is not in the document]
nielsm5: Technically it's possible however it's always UTF-8 and I plan on refactoring this bit of code soon. Since the new gateways It's become clear to me that we need to handle our responses better, instead of using a raw *T* type, and this ugly method..

tnleeuw (approved):

PR #4713, opened 2023-03-29
Attempt to reopen BtmDiskJournal

nielsm5:


tnleeuw (approved):
I wonder if we may need to try re-opening on every IOException, regardless of the message-text.
I'm rather unhappy we cannot better test this and worry we will continue to apply "fix" after "fix" without ever being sure we really fix anything.

So a conditional approval.

nielsm5:
If it fails it needs to roll back everything (as much as it can) and retry it later. Which should trigger another recovery.

tnleeuw:
You have my blessing to merge it.

PR #4741, opened 2023-04-04
Workaround for bug that manual retry does not extract msg from CSV row

tnleeuw:
Ugly fixes and ugly test-mocks for ugly code.

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/jdbc/MessageStoreListener.java
+++ b/core/src/main/java/nl/nn/adapterframework/jdbc/MessageStoreListener.java
@@ -101,11 +101,7 @@ public class MessageStoreListener<M> extends JdbcTableListener<M> {
 	public void configure() throws ConfigurationException {
 		super.configure();
 		if (sessionKeys != null) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/nl/nn/adapterframework/jdbc/MessageStoreListener.java
```diff
@@ -101,11 +101,7 @@
 	public void configure() throws ConfigurationException {
 		super.configure();
 		if (sessionKeys != null) {
-			sessionKeysList = new ArrayList<>();
-			StringTokenizer stringTokenizer = new StringTokenizer(sessionKeys, ",");
-			while (stringTokenizer.hasMoreElements()) {
-				sessionKeysList.add(stringTokenizer.nextToken());
-			}
+			extractSessionKeyList();
```
tnleeuw, line 104: Code extracted to a public method for testability and being able to avoid calling `super.configure()`, which crashes without having more configuration and a database -- that is more hassle than I want to add for the tests in the 7.8 branch, this close to release.

nielsm5 (approved):

PR #4750, opened 2023-04-06
Fix loadOnStartup check in CmisProxyServlet

nielsm5:
I've done a few UC and LC tests and it turns out the char[] conversion is faster:
char[]  1278314600
substr 1382596200
strbuf 1373397800

```
	public static String charArr(String input) {
		char[] c = input.toCharArray();
		c[0] = Character.toLowerCase(c[0]);
		return new String(c);
	}

	private static String subStr(String input) {
		return input.substring(0, 1).toLowerCase() + input.substring(1);
	}

	private static String strBuff(String input) {
		StringBuffer sb = new StringBuffer(input.length());
		sb.append(Character.toLowerCase(input.charAt(0)));
		sb.append(input.substring(1));
		return sb.toString();
	}
```

### DIFF
--- a/iaf-commons/src/main/java/nl/nn/adapterframework/util/StringUtil.java
+++ b/iaf-commons/src/main/java/nl/nn/adapterframework/util/StringUtil.java
@@ -221,8 +221,6 @@ public class StringUtil {
 		return count;
 	}
 
-	// Not sure what's better here...
-	// return input.substring(0, 1).toLowerCase() + input.substring(1);
 	/**
 	 * Turns the first Char into lower case.
 	 */
--- a/iaf-commons/src/main/java/nl/nn/adapterframework/util/StringUtil.java
+++ b/iaf-commons/src/main/java/nl/nn/adapterframework/util/StringUtil.java
@@ -220,4 +220,24 @@ public class StringUtil {
 			count++;
 		return count;
 	}
+
+	// Not sure what's better here...
+	// return input.substring(0, 1).toLowerCase() + input.substring(1);
+	/**
+	 * Turns the first Char into lower case.
+	 */
+	public static String lcFirst(String input) {
+		char[] c = input.toCharArray();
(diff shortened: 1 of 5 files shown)


iaf-commons/src/main/java/nl/nn/adapterframework/util/StringUtil.java
```diff
@@ -220,4 +220,24 @@ public static int countRegex(String string, String regex) {
 			count++;
 		return count;
 	}
+
+	// Not sure what's better here...
+	// return input.substring(0, 1).toLowerCase() + input.substring(1);
```
tnleeuw, line 225: I don't think that it makes much of a difference here, what is better. I would perhaps slightly prefer this form but the code is short and low level, and both feel kind of ugly to me in different ways.

iaf-commons/src/test/java/nl/nn/adapterframework/util/StringUtilTest.java
```diff
@@ -95,4 +95,19 @@ public void testCountRegex() {
 		assertEquals(4, regexCount);
 	}
 
+	@Test
```
tnleeuw, line 98: Make these tests parameterised, perhaps?

tnleeuw (approved):

iaf-commons/src/test/java/nl/nn/adapterframework/util/StringUtilTest.java
nielsm5: I tried but it became a bit bloated (using the CsvSource).
One big long annotation wont do the readability any good, so I opted for the 4 assertions instead.

tnleeuw: Putting each combination on a new line that never bothers me personally but that is also a matter of personal preference of course!

Issue #2763, opened 2022-02-15
"attribute [correlationIdToHexPrefix] has no value" - warning


laura-dimitra:
After upgrading the iaf version in ibis4xxx to 7.7-RC1 I got the following warning:  

**Configuration [xxx] - JmsSender [JMS Sender to send document data to xxx] attribute [correlationIdToHexPrefix] has no value**

I've already read the documentation, and here's written it's allowed to be empty:
![MicrosoftTeams-image (18)](https://user-images.githubusercontent.com/30295058/154094928-05488df3-56b5-478e-b89f-5e21f5ea31db.png)

I've already tried to add the attribute: correlationIdToHex = "true"  (the original configuration only has the attributecorrelationIdToHexPrefix), but the warning doesn't disappear.

![Knipsel](https://user-images.githubusercontent.com/30295058/154095934-b206e3cc-1f33-46dd-86da-6670253ef3ce.PNG)

PR #4758, opened 2023-04-10
Make convertToType more generic, and reduce empty value console warnings

nielsm5:
Make `ClassUtils#convertToType` more generic
Reduce empty value console warnings
Add tests to cover #2763
Revert iaf-test jms queue suffix removal #4749

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/util/ClassUtils.java
+++ b/core/src/main/java/nl/nn/adapterframework/util/ClassUtils.java
@@ -34,6 +34,8 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 
+import javax.annotation.Nonnull;
+
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.Logger;
 
@@ -299,7 +301,7 @@ public abstract class ClassUtils {
 	 */
 	public static void invokeSetter(Object bean, Method method, String valueToSet) {
 		if(!method.getName().startsWith("set") || method.getParameterTypes().length != 1) {
-			throw new IllegalArgumentException("method must start with [set] and may only contain [1] parameter");
+			throw new IllegalStateException("method must start with [set] and may only contain [1] parameter");
 		}
 
 		try {//Only always grab the first value because we explicitly check method.getParameterTypes().length != 1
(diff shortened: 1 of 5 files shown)


core/src/main/java/nl/nn/adapterframework/util/ClassUtils.java
```diff
@@ -299,7 +301,7 @@ public static String classNameOf(Object o) {
 	 */
 	public static void invokeSetter(Object bean, Method method, String valueToSet) {
 		if(!method.getName().startsWith("set") || method.getParameterTypes().length != 1) {
-			throw new IllegalArgumentException("method must start with [set] and may only contain [1] parameter");
+			throw new IllegalStateException("method must start with [set] and may only contain [1] parameter");
```
tnleeuw, line 304: I'd argue that `IllegalArgumentException` was more appropriate, what was the motivation to change?

core/src/test/java/nl/nn/adapterframework/util/ClassUtilsTest.java
```diff
@@ -311,19 +311,23 @@ private IScopeProvider getBytesClassLoaderProvider() throws Exception {
...
+		assertEquals("7", ClassUtils.convertToType(String.class, "7"));
+		assertEquals(true, ClassUtils.convertToType(boolean.class, "true"));
+		assertEquals(true, ClassUtils.convertToType(Boolean.class, "true"));
+		assertEquals(false, ClassUtils.convertToType(Boolean.class, "niet true"));
+		assertEquals(TestEnum.ONE, ClassUtils.convertToType(TestEnum.class, "one"));
+
+		assertThrows(IllegalArgumentException.class, ()->ClassUtils.convertToType(Object.class, "dummy"));
+		assertThrows(IllegalArgumentException.class, ()->ClassUtils.convertToType(Long.class, "dummy"));
+
+		assertThrows(IllegalArgumentException.class, ()->ClassUtils.convertToType(int.class, "")); //Empty string
```
tnleeuw, line 331: Could be wrapped in a call to `assertAll`

tnleeuw (approved):

core/src/main/java/nl/nn/adapterframework/util/ClassUtils.java
nielsm5: I agree, however in this case the IllegalArgumentExceptions are meant for the `valueToSet` and not the `method` being invalid.

core/src/test/java/nl/nn/adapterframework/util/ClassUtilsTest.java
nielsm5: Ah ja, nog een beetje nieuw voor mij.

PR #4773, opened 2023-04-11
Cleanup monitoring code

nielsm5:
- Rename Enums
- Move GET endpoints to the BUS
- Change the API so it doesn't send redundant information


tnleeuw (approved):
Looks good, a couple of Codacy & Liftbot things that could perhaps be addressed now.

PR #4836, opened 2023-04-28
Create JacksonUtils for ObjectMapping

nielsm5:


management-gateway/src/main/java/nl/nn/adapterframework/util/JacksonUtils.java
```diff
@@ -0,0 +1,63 @@
...
+		try {
+			return MAPPER.writeValueAsString(payload);
+		} catch (JacksonException e) {
+			throw new BusException("unable to convert response to JSON", e);
+		}
+	}
+
+	public static <T> T convertToDTO(Object payload, Class<T> dto) {
+		try {
+			if(payload instanceof String) {
```
tnleeuw, line 50: Why no overloaded methods -- is it not possible to distinguish at the call-site between payload types `String` or `byte[]`?

management-gateway/src/main/java/nl/nn/adapterframework/util/JacksonUtils.java
```diff
@@ -0,0 +1,63 @@
...
+				return MAPPER.readValue((String) payload, dto);
+			} else if(payload instanceof byte[]) {
+				return MAPPER.readValue((byte[]) payload, dto);
+			}
```
tnleeuw, line 54: `else throw ...` right here would be clearer code to me, then the `throw` statement below, because it is not connected to the if/if-else here.

tnleeuw (approved):

management-gateway/src/main/java/nl/nn/adapterframework/util/JacksonUtils.java
nielsm5: No, the payload is an Object

management-gateway/src/main/java/nl/nn/adapterframework/util/JacksonUtils.java
```diff
@@ -51,13 +51,13 @@ public static <T> T convertToDTO(Object payload, Class<T> dto) {
 				return MAPPER.readValue((String) payload, dto);
 			} else if(payload instanceof byte[]) {
 				return MAPPER.readValue((byte[]) payload, dto);
+			} else {
+				throw new NotImplementedException("unhandled payload type ["+payload.getClass()+"]");
```
tnleeuw, line 55: Potential NPE here? (Perhaps not a big issue since it's a throw exception either way).

tnleeuw (approved):

PR #4880, opened 2023-05-11
Refactor collectors

nielsm5:
Reverts some changes made in #3928


core/src/main/java/nl/nn/adapterframework/collection/CollectorPipeBase.java
```diff
@@ -0,0 +1,134 @@
...
+			Collection<C, P> collection = getCollection(session);
+			switch(action) {
+				case OPEN: {
+					if(collection != null) {
+						throw new CollectionException("collection [" + getCollectionName() + "] is already open, unable to create a new one");
+					}
+					collection = new Collection<>(createCollector());
+					session.scheduleCloseOnSessionExit(collection, ClassUtils.nameOf(this));
+					session.put(getCollectionName(), collection);
+					break; //should also be able to write?
```
tnleeuw, line 104: Should writing from OPEN then be optional?
Needs an extra parameter in that case to decide if to break, or fall thru to WRITE?

Let's decide and then remove this comment.

core/src/main/java/nl/nn/adapterframework/compression/MessageZipEntry.java
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	@Override
+	public void setSize(long size) {
+		if(size > 0) {
+			super.setSize(size);
+		}
+	}
+
+	/** Doesn't do anything is the size was already known. Computes if unknown. */
```
tnleeuw, line 60: "is" -> "if"

core/src/main/java/nl/nn/adapterframework/compression/MessageZipEntry.java
```diff
@@ -0,0 +1,89 @@
...
+		this(message, String.valueOf(message.getContext().get(MessageContext.METADATA_NAME)));
+	}
+
+	public MessageZipEntry(Message message, String filename) {
+		super(filename);
+
+		if (message!=null) {
+			this.message = message;
+
+			setSize(message.size());
```
tnleeuw, line 46: Why `message.size()` instead of `computeSize()`?

core/src/main/java/nl/nn/adapterframework/compression/ZipWriter.java
```diff
@@ -15,284 +15,103 @@
...
-				@Override
-				public void afterClose() throws Exception {
-					stream.flush();
-					closeEntry();
+	public Message build(List<MessageZipEntry> parts) throws IOException {
+		File collectorsTempFolder = FileUtils.getTempDirectory("collectors");
+		File tempFile = File.createTempFile("msg", ".zip", collectorsTempFolder);
+
+		try (FileOutputStream fos = new FileOutputStream(tempFile)) {
+			try (ZipOutputStream zipoutput = new ZipOutputStream(fos)) {
```
tnleeuw, line 101: volgens mij is dit in 1 try-with-resources regel te combineren door een `;` te zetten achter de eerste declaratie.

core/src/main/java/nl/nn/adapterframework/compression/ZipWriter.java
```diff
@@ -15,284 +15,103 @@
...
-	public static ZipWriter createZipWriter(PipeLineSession session, String handlekey, OutputStream resultStream, boolean closeOnExit) {
-		ZipWriter handle=new ZipWriter(resultStream,closeOnExit);
-		session.put(handlekey,handle);
-		handle.log.debug("opened new zipstream [{}]", handlekey);
-		return handle;
-	}
+	private final boolean includeFileHeaders;
 
-	public static void configure(Action action, ParameterList parameterList) throws ConfigurationException {
+	static void configure(Action action, ParameterList parameterList) throws ConfigurationException {
```
tnleeuw, line 48: - Why is this method static?
- Is `configure` the right name? Perhaps `validate` is better, since that's what it really  seems to do? Or `validateConfiguration`?

tnleeuw (changes requested):
Great cleanup and simplification of the code.

Some small changes / clarifications that I'd like to see.

core/src/main/java/nl/nn/adapterframework/collection/CollectorPipeBase.java
nielsm5: I think it's better not to. I would be nice but allows parameters to be set on the OPEN action. I don't think this extra complexity is worth it. The LAST action already allows the combining of 2 actions, so you will always need a minimum of 2 actions to create a collection.

core/src/main/java/nl/nn/adapterframework/compression/MessageZipEntry.java
nielsm5: Size is an optional field, if it's unknown and not required, there is no need to compute it.

core/src/main/java/nl/nn/adapterframework/compression/MessageZipEntry.java
```diff
@@ -0,0 +1,89 @@
...
+
+	public void computeCrc() {
+		Long crc32 = MessageUtils.generateCRC32(message);
+		if(crc32 != null) {
+			setCrc(crc32);
+		}
+	}
+
+	public void computeFileHeaders() {
+		setMethod(ZipEntry.STORED);
```
tnleeuw, line 73: Why STORED, and not DEFLATED?

tnleeuw (approved):

PR #4904, opened 2023-05-17
Fix unwanted elements related to MessageStoreSender

mhdirkse:


mhdirkse:
Goes together with https://github.com/ibissource/frank-doc/pull/158.

mhdirkse:
[MessageStoreSender.zip](https://github.com/ibissource/iaf/files/11497004/MessageStoreSender.zip)

This is the comparison between the Frank!Doc JSON and XSDs before these two pull requests and after. The elements `MessageStoreSenderErrorStorage` and `MessageStoreSenderMessageLog` are removed as should be. In addition, `ErrorStorageElementGroup` and `MessageLogElementGroup` loose child element `Param`. These are the groups for using Syntax 1 elements `errorStorage` and `messageLog` in which you add a `className` attribute. These are the only effects of the PRs.

nielsm5 (approved):

PR #4951, opened 2023-06-06
Add SftpFileSystem (pipe/sender/listener)

nielsm5:
Cleanup unreachable SFTP code
Split SFTP and FTPs classes.
Add JUnit tests


tnleeuw (approved):
Just a few very minor nitpicks that can be picked up later as well.

Issue #4883, opened 2023-05-11
Add storageClass option to S3 filesystem


Laurens-makel:
**❗ please do not add sensitive information in issues, you can provide extra information via email using issue number as reference ❗**

**Describe the issue**
AWS S3 offers different classes of storage: https://aws.amazon.com/s3/storage-classes/

Currently the framework does not offer any options to the Frank developer to set the desired storage class and defaults to the standard class. 

We are having a use case in which we would like to move files when a certain process is completed, whilst moving we would like to change the storage class from 'Standard' to 'Glacier Deep Archive'

Please introduce support to set the desired storage class for the following actions:

- write
- copy
- move

PR #4977, opened 2023-06-09
Implement storage class attribute for s3 actions write, move, copy and rename

Laurens-makel:
implements #4883


tnleeuw:
I'm not sure if this PR will fix your problem? Did you test this with the actual intended storage classes on source and destination files?

When we were reviewing this ticket earlier, and checked the documentation for the archival storage classes, I got the impression that not all storage classes can be mixed in a single bucket and you might need to move files into a different bucket to allow setting the deep-archival storage class you need.

Moving between buckets does not seem to be possible, from a glance at the code.

Laurens-makel:
> I'm not sure if this PR will fix your problem? Did you test this with the actual intended storage classes on source and destination files?
> 
> When we were reviewing this ticket earlier, and checked the documentation for the archival storage classes, I got the impression that not all storage classes can be mixed in a single bucket and you might need to move files into a different bucket to allow setting the deep-archival storage class you need.
> 
> Moving between buckets does not seem to be possible, from a glance at the code.

This is incorrect, I have tested this myself with custom code and is working perfectly fine on TST/ACC environments.....

nielsm5:
Grappig dat dat ook gewoon binnen 1 bucket kan... Vage shit dat AWS

aws/src/main/java/nl/nn/adapterframework/filesystem/AmazonS3FileSystem.java
```diff
@@ -265,7 +271,12 @@ public void close() throws IOException {
 						ObjectMetadata metaData = new ObjectMetadata();
 						metaData.setContentLength(file.length());
 
-						s3Client.putObject(bucketName, f.getKey(), fis, metaData);
+						PutObjectRequest por = new PutObjectRequest(bucketName, f.getKey(), fis, metaData);
+						if(StringUtils.isNotEmpty(getStorageClass())){
+							por.setStorageClass(getStorageClass());
+						}
+
+						s3Client.putObject(por);
```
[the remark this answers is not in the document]
nielsm5: Seintje 🔔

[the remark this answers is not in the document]
Laurens-makel: Seintje terug dat ik het bijgewerkt heb

Laurens-makel:
> Grappig dat dat ook gewoon binnen 1 bucket kan... Vage shit dat AWS

Op deze manier heb je slechts 1 bucket nodig om een proces te faciliteren ipv meerdere..  handig feature! Anders zou iets moeten opzetten om bestanden van de ene bucket naar de andere bucket te moven ( en bijbehorende kosten betalen.. ).

Met de Deep Glacier storage kunnen we bestanden na verwerking nog 6 maanden bewaren voor met name incident maar ook andere onderzoeks doeleinden, voor ongeveer een tiende van kosten. De enige twee nadelen zijn dat je het minimaal 6 maanden moet bewaren en dat retrieval niet instant is maar tot 12 uur kan duren.

aws/src/main/java/nl/nn/adapterframework/filesystem/AmazonS3FileSystem.java
```diff
@@ -657,6 +675,11 @@ public void setProxyPort(Integer proxyPort) {
 		this.proxyPort = proxyPort;
 	}
 
+	/** S3 Storage Class, for more information see https://aws.amazon.com/s3/storage-classes/ */
+	public void setStorageClass(String storageClass) {
```
tnleeuw, line 679: There are many different storage classes; to get some kind of validation during the configuration I would like to see here a translation from the string to the enum value `StorageClass`.

aws/src/main/java/nl/nn/adapterframework/filesystem/AmazonS3FileSystem.java
```diff
@@ -30,6 +30,10 @@
 import java.util.List;
 import java.util.Map;
 
+import com.amazonaws.services.s3.model.CopyObjectRequest;
```
tnleeuw, line 33: Please sort the imports better.

aws/src/main/java/nl/nn/adapterframework/filesystem/AmazonS3FileSystem.java
```diff
@@ -92,6 +96,8 @@ public class AmazonS3FileSystem extends FileSystemBase<S3Object> implements IWri
 	private @Getter Integer proxyPort = null;
 	private @Getter int maxConnections = 50;
 
+	private @Getter String storageClass = null;
```
tnleeuw, line 99: If the storageClass applies to every operation on every file, it would make more sense to put it on the bucket?

If you want to apply a specific storage class to files after specific operations, should we then not look at how to add storage class as an option / parameter to an operation (such as the "move file" operation)?

Then storage classes can be defined on a more granular level.

Laurens-makel: > If the storageClass applies to every operation on every file, it would make more sense to put it on the bucket?

This is true for this specific instance of the AmazonS3FileSystem that sets the storage class, there could be other instances (in other adapters for example) where the storage class is not set but operate on the same bucket. 

For example in my use case, I'd upload the files under the default storage class but move them to glacier storage once a certain process has been completed. 

I am also not aware on how to configure another storage class as the default on a bucket. A quick google suggests to use lifecycle rules to transition objects between storage classes, but why would you pay money to AWS to do this for you if you could set the desired storage class during creation of the object :)

Besides that, I totally agree that being able to set the desired storage class trough a parameter would enable the Frank developer to use the same instance with different storage classes, which definitely will have its related use-cases.

tnleeuw (approved):

tnleeuw (approved):

PR #4978, opened 2023-06-09
Limit SFTP/FTP/S3 FileSystem handles

nielsm5:


aws/src/main/java/nl/nn/adapterframework/filesystem/AmazonS3FileSystem.java
```diff
@@ -293,21 +298,16 @@ public Message readFile(S3Object f, String charset) throws FileSystemException,
...
 	 */
-	private S3Object resolve(S3Object f) {
-		if(f instanceof LocalS3Object) {
-			return s3Client.getObject(bucketName, f.getKey());
+	private S3Object updateFileAttributes(S3Object f) {
+		if(f.getObjectMetadata().getRawMetadataValue(Headers.CONTENT_LENGTH) == null) {
+			ObjectMetadata omd = s3Client.getObjectMetadata(bucketName, f.getKey());
+			f.setObjectMetadata(omd);
 		}
 		return f;
```
tnleeuw, line 308: Do you actually still need the return value of these methods in these file systems?

tnleeuw (approved):

aws/src/main/java/nl/nn/adapterframework/filesystem/AmazonS3FileSystem.java
nielsm5: Not really ":)

PR #4979, opened 2023-06-09
Make Message return a MessageContext object

nielsm5:
Ik zag in #4978 wat context dingen voorkomen en dacht laat ik maar gelijk een eerste stap maken..

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/stream/Message.java
+++ b/core/src/main/java/nl/nn/adapterframework/stream/Message.java
@@ -38,7 +38,6 @@ import java.nio.charset.Charset;
 import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.LinkedHashSet;
-import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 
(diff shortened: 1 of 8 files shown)


core/src/main/java/nl/nn/adapterframework/stream/Message.java
```diff
@@ -101,11 +100,11 @@ private Message(Map<String, Object> context, Object request, Class<?> requestCla
 		this.requestClass = requestClass != null ? ClassUtils.nameOf(requestClass) : ClassUtils.nameOf(request);
 	}
 
-	private Message(Map<String, Object> context, Object request) {
+	private Message(MessageContext context, Object request) {
 		this(context, request, request != null ? request.getClass() : null);
 	}
 
-	public Message(String request, Map<String, Object> context) {
+	public Message(String request, MessageContext context) {
```
tnleeuw, line 107: Can we add `@Nonnullable` annotations to all these constructors as well, for the context at least?
(Perhaps also for the request, although there are some private constructors that do allow NULL request).

I would like at least context to never be NULL though and that is not guaranteed now.

core/src/main/java/nl/nn/adapterframework/stream/Message.java
tnleeuw, line 1: There's a null-check on messageContext in method `toStringPrefix()` which can now be removed as well.

tnleeuw (changes requested):

core/src/main/java/nl/nn/adapterframework/stream/Message.java
nielsm5: My silly Eclipse doesn't let me know about these things :(

tnleeuw (approved):

PR #5006, opened 2023-06-14
Fix NPE when unloading a config and calling getConfigWarnings

nielsm5:


core/src/main/java/nl/nn/adapterframework/configuration/ApplicationWarningsBase.java
```diff
@@ -53,15 +53,15 @@ public void destroy() throws Exception {
 	}
 
 	public int size() {
-		return warnings.size();
+		return getWarnings().size();
```
tnleeuw, line 56: Perhaps instead of fixing the issue this way, perhaps it would be better to change the implementation of the `destroy()` method to `warnings.clear()` instead of `warnings = null`?

Then there should be no NPE anymore.

nielsm5: Yea but then it can still be referenced elsewhere preventing it from being destroyed properly.

tnleeuw: > Ehh, I'm not sure anymore. Instances can be created through Spring or manually, it's been a while...
> We can clear the list, sure!

Clearing the list just makes the rest of the code simpler and consistent again. Because there is other code which references the instance directly, so using the getter-method in these places makes once wonder "why?", and then the getter returns an unmodifiable instance when it's null so addWarnings() cannot use the getter, and then it gets confusing when someone looks at it a year from now, etc... 

That's why I'm not happy with it.

tnleeuw (approved):

PR #5013, opened 2023-06-16
Limit amount of BTM DiskJournal retries

nielsm5:


core/src/main/java/nl/nn/adapterframework/jta/btm/BtmDiskJournal.java
```diff
@@ -89,9 +93,15 @@ public Map<Uid, JournalRecord> collectDanglingRecords() throws IOException {
 	}
 
 	private void recover(IOException e) throws IOException {
-		log.error("FileChannel exception, attempting to recover DiskJournal", e);
+		int errorCount = ERROR_COUNT.incrementAndGet();
 
 		close();
+		if(errorCount > MAX_ERROR_COUNT) {
```
tnleeuw, line 99: Should there also be any tracking of how long ago was the last error? Clearing this back to 1 if last error was older than some threshold?

tnleeuw (commented):
Before approving I have 1 question, see comments.

tnleeuw (approved):

tnleeuw (approved):

PR #5015, opened 2023-06-16
Copy session into new session created for the sub-adapter call

tnleeuw:
Copy message-id from input-session into the RawMessageWrapper of the sub-adapter.

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/receivers/JavaListener.java
+++ b/core/src/main/java/nl/nn/adapterframework/receivers/JavaListener.java
@@ -68,7 +68,7 @@ public class JavaListener<M> implements IPushingListener<M>, RequestProcessor, H
 
 	private final @Getter(onMethod = @__(@Override)) String domain = "JVM";
 	protected Logger log = LogUtil.getLogger(this);
-	private @Getter ClassLoader configurationClassLoader = Thread.currentThread().getContextClassLoader();
+	private final @Getter ClassLoader configurationClassLoader = Thread.currentThread().getContextClassLoader();
 	private @Getter @Setter ApplicationContext applicationContext;
 
 	private @Getter String name;
@@ -79,7 +79,7 @@ public class JavaListener<M> implements IPushingListener<M>, RequestProcessor, H
 	private @Getter boolean httpWsdl = false;
 
 	private @Getter boolean open=false;
-	private static Map<String, JavaListener> registeredListeners;
+	private static Map<String, JavaListener<?>> registeredListeners;
 	private @Getter @Setter IMessageHandler<M> handler;
 
 	@Override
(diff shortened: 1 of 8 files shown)


core/src/main/java/nl/nn/adapterframework/senders/IbisLocalSender.java
```diff
@@ -227,12 +227,12 @@ private String getActualServiceName(PipeLineSession session) throws SenderExcept
 	@Override
 	public SenderResult sendMessage(Message message, PipeLineSession session) throws SenderException, TimeoutException {
 		SenderResult result;
-		try (PipeLineSession context = new PipeLineSession()) {
+		try (PipeLineSession nestedSession = new PipeLineSession(session)) {
```
tnleeuw, line 230: Copying original session into the nested session is the actual fix for the issue.

core/src/main/java/nl/nn/adapterframework/receivers/JavaListener.java
```diff
@@ -131,11 +131,18 @@ public RawMessageWrapper<M> wrapRawMessage(M rawMessage, PipeLineSession session
...
+			String messageId;
 			if (context != null) {
 				context.put(PipeLineSession.CORRELATION_ID_KEY, correlationId);
+				messageId = (String) context.get(PipeLineSession.MESSAGE_ID_KEY);
+			} else {
+				messageId = null;
+			}
+			RawMessageWrapper<M> rawMessageWrapper = new RawMessageWrapper<>((M)rawMessage, messageId, correlationId);
+			Message result;
+			try (PipeLineSession session = new PipeLineSession(context != null ? context : Collections.emptyMap())) {
```
tnleeuw, line 143: Create session here so we do not have to do it in the method we call, b/c that method in most cases already gets a session specific for this call.

core/src/main/java/nl/nn/adapterframework/receivers/JavaListener.java
```diff
@@ -144,45 +151,40 @@ public String processRequest(String correlationId, String rawMessage, HashMap co
...
+		Object object = session.get("httpRequest");
+		if (object != null) {
+			if (object instanceof HttpServletRequest) {
+				ISecurityHandler securityHandler = new HttpSecurityHandler((HttpServletRequest)object);
+				session.put(PipeLineSession.SECURITY_HANDLER_KEY, securityHandler);
+			} else {
+				log.warn("No securityHandler added for httpRequest [{}]", object::getClass);
 			}
 		}
-		try (PipeLineSession session = new PipeLineSession(context)) {
```
tnleeuw, line 170: We do not need to create a session here since the session we get is already good to use, from all callers.

core/src/main/java/nl/nn/adapterframework/senders/IbisLocalSender.java
```diff
@@ -254,39 +257,39 @@ public SenderResult sendMessage(Message message, PipeLineSession session) throws
...
 						message.preserve();
 						log.debug("{} calling {} in asynchronously", this::getLogPrefix, () -> serviceIndication);
-						isolatedServiceCaller.callServiceAsynchronous(serviceClient, message, context, threadLifeCycleEventListener);
+						isolatedServiceCaller.callServiceAsynchronous(serviceClient, message, subAdapterSession, threadLifeCycleEventListener);
 						result = new SenderResult(message);
 					}
 				} else {
 					log.debug("{} calling {} in same Thread", this::getLogPrefix, () -> serviceIndication);
-					result = new SenderResult(serviceClient.processRequest(message, context));
+					result = new SenderResult(serviceClient.processRequest(message, subAdapterSession));
```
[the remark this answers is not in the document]
tnleeuw: @sonatype-ignore False positive, closed via session later.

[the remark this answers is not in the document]
tnleeuw: @sonatype-lift ignore false positive, closed via session.

nielsm5 (approved):

PR #5024, opened 2023-06-19
Limit amount of BTM DiskJournal retries

nielsm5:
Backport of  #5013


core/src/main/java/nl/nn/adapterframework/jta/btm/BtmDiskJournal.java
```diff
@@ -42,6 +46,9 @@ public class BtmDiskJournal extends DiskJournal {
 	private static final String LOG_ERR_MSG = "cannot write log, disk logger is not open";
 	private static final String FORCE_ERR_MSG = "cannot force log writing, disk logger is not open";
 	private static final String COLLECT_ERR_MSG = "cannot collect dangling records, disk logger is not open";
+	private static final AtomicInteger ERROR_COUNT = new AtomicInteger(0);
+	private static final int MAX_ERROR_COUNT = AppConstants.getInstance().getInt("transactionmanager.btm.journal.maxRetries", 500);
```
tnleeuw, line 50: Is 500 a good nr? Seems rather high to me. How about 50, or 100?

(Especially since the recover-method is now synchronised so there shouldn't be a ridiculous amount of simultaneous recovery attempts).

tnleeuw (approved):

tnleeuw (approved):

PR #5126, opened 2023-07-14
Fix Json2XmlValidator response type parsing

nielsm5:
Feature introduced in #4314 not taking into effect the accept header can be `*/*` or contain multiple values.

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/pipes/Json2XmlValidator.java
+++ b/core/src/main/java/nl/nn/adapterframework/pipes/Json2XmlValidator.java
@@ -20,6 +20,7 @@ import java.io.StringReader;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 import javax.xml.validation.ValidatorHandler;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/nl/nn/adapterframework/pipes/Json2XmlValidator.java
```diff
@@ -137,6 +141,18 @@ protected void storeInputFormat(DocumentFormat format, Message input, PipeLineSe
...
+	/**
+	 * Default format has precedence over the accept header, accept header may be invalid or * slash *, in which case it should be ignored. First accept value wins.
+	 */
+	private String parseAcceptHeader(DocumentFormat detectedFormat, String acceptHeaderValue) {
+		if(StringUtils.isEmpty(acceptHeaderValue) || "*/*".equals(acceptHeaderValue)) {
+			return detectedFormat.name();
+		}
+
+		Optional<String> value = StringUtil.splitToStream(acceptHeaderValue).filter(t -> !t.equals("*/*")).findFirst();
+		return value.orElse(detectedFormat.name());
```
tnleeuw, line 153: It's a personal thing perhaps, but I find that code using the Streams API is more readable when all calls are chained (so get rid of the intermediate "value" variable), and each step is on a new line so you can clearly read what the steps are in the stream-processing.

So that would look like:
```
return StringUtil.splitToStream(acceptHeaderValue)
    .filter(t -> !t.equals("*/*"))
    .findFirst()
    .orElse(detectedFormat.name());
```

Up to you if you agree or not, of course!

tnleeuw (approved):

core/src/main/java/nl/nn/adapterframework/pipes/Json2XmlValidator.java
nielsm5: Got it! Will remember this 😄 Especially the new lines, because that's also something I struggle with!

Issue #4390, opened 2023-01-20
Refactor/cleanup JmsFacade/MessagingSource/MessagingSourceFactory


nielsm5:
**Describe the issue**
Right now the JMS Facade is a bit messy with both the MessagingSource, MessagingSourceFactory and the Facade doing JMS related things. The code should be cleaned up and tested. A check should be performed if the QCF is wrapped in a TransactionAwareConnectionFactoryProxy and like the `TransactionalDbmsSupportAwareDataSourceProxy` the metadata should be cached.
Also both `MessagingSource#getPhysicalName()` and `JmsMessagingSourceFactory#getConnectionFactoryInfo(...)` have methods to retrieve info about the QCF.

**Reporter**
Niels

PR #5142, opened 2023-07-18
Add more JDBC and JMS pooling

nielsm5:
Also cleaned up some code as stated in #4390
Supersedes #4572

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/jta/narayana/NarayanaDataSource.java
+++ b/core/src/main/java/nl/nn/adapterframework/jta/narayana/NarayanaDataSource.java
@@ -47,7 +47,7 @@ import nl.nn.adapterframework.util.LogUtil;
  *
  */
 public class NarayanaDataSource implements DataSource {
-	private static final Logger LOG = LogUtil.getLogger(NarayanaDataSource.class);
+	private final Logger log = LogUtil.getLogger(NarayanaDataSource.class);
 
 	private final @Getter XADataSource targetDataSource;
 	private final String name;
@@ -67,7 +67,7 @@ public class NarayanaDataSource implements DataSource {
 	}
 
 	/** In order to allow transactions to run over multiple databases, see {@link ModifierFactory}. */
-	public static void checkModifiers(DataSource dataSource) {
+	private void checkModifiers(DataSource dataSource) {
 		try (Connection connection = dataSource.getConnection()) {
 			DatabaseMetaData metadata = connection.getMetaData();
 			String driverName = metadata.getDriverName();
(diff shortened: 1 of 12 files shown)


core/src/main/java/nl/nn/adapterframework/jta/narayana/NarayanaDataSourceFactory.java
```diff
@@ -16,61 +16,96 @@
...
+		PoolableConnectionFactory poolableConnectionFactory = new PoolableManagedConnectionFactory(cf, null);
+
+		ObjectPool<PoolableConnection> connectionPool = createConnectionPool(poolableConnectionFactory);
+
+		PoolingDataSource<PoolableConnection> ds = new ManagedDataSource<>(connectionPool, cf.getTransactionRegistry());
+		log.info("created XA-enabled PoolingDataSource [{}]", ds);
+		return ds;
 	}
 
+	private ObjectPool<PoolableConnection> createConnectionPool(PoolableConnectionFactory poolableConnectionFactory) {
```
tnleeuw, line 99: This method is duplicated with the code to create pool for BTM -- can this code perhaps be shared? To ease maintenance if something should still be changed ?

Or do you think that soon enough we'll be able to drop BTM and then better to have all the relevant code here in this class?

tnleeuw (approved):
Minor comments only after doing a thorough check!

(Assuming all tests pass ;-) )

core/src/main/java/nl/nn/adapterframework/jta/narayana/NarayanaDataSourceFactory.java
nielsm5: No, that's all BTM native, and thus it cannot be shared.

core/src/main/java/nl/nn/adapterframework/jta/narayana/NarayanaDataSource.java
```diff
@@ -17,43 +17,75 @@
...
+		if(!(dataSource instanceof XADataSource)) {
+			throw new IllegalStateException("Only XA DataSources can be registerd with a TransactionManager");
+		}
+
+		this.targetDataSource = (XADataSource) dataSource;
+		this.name = name;
+
+		//If the DataSource implements both DataSource and XADataSource, check for modifiers.
+		if(dataSource instanceof DataSource) {
+			checkModifiers((DataSource) dataSource);
```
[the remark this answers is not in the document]
nielsm5: Because it's not possible to get the connection metadata on an XAConnection 😭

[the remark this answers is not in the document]
tnleeuw: An XAConnection is not yet a Connection if I read it correctly, it looks like you still have to call `getConnection()` on the XAConnection to get an actual connection... Don't ask me why!

core/src/main/java/nl/nn/adapterframework/jta/narayana/NarayanaDataSourceFactory.java
nielsm5: Yea I know it's confusing, but now we know where BTM has it's pooling mechanism from 😄

PR #5259, opened 2023-08-16
Fix and add more logging to MessageUtils

nielsm5:


core/src/main/java/nl/nn/adapterframework/util/MessageUtils.java
```diff
@@ -56,7 +56,10 @@ public static MessageContext getContext(HttpServletRequest request) {
 		result.withCharset(request.getCharacterEncoding());
 		int contentLength = request.getContentLength();
 		result.withSize(contentLength);
-		result.withMimeType(request.getContentType());
+		String contentType = request.getContentType();
+		if(StringUtils.isNotEmpty(contentType)) {
+			result.withMimeType(request.getContentType());
```
philipsens, line 61: Maybe this instead, to save some calls?
```suggestion
			result.withMimeType(contentType);
```

Matthbo: I agree, you store the exact call above already so no need to call it again

nielsm5: In this case it's a pointer to a field and not a value that needs to be calculated every time, but I like your way of thinking.

philipsens (approved):

PR #5315, opened 2023-08-31
ASPOSE: Strip MimeType params for converters

nielsm5:


tnleeuw (approved):

tnleeuw (changes requested):
In principle the changes look good but the unit test fails on a URI that's invalid (on Unix at least) so I'll re-approve once that is fixed.

tnleeuw (approved):

PR #5318, opened 2023-08-31
Allow global setting for the use of eTags and disable by default (7.8)

nielsm5:


tnleeuw (approved):

core/src/main/java/nl/nn/adapterframework/http/rest/ApiListener.java
```diff
@@ -253,9 +253,9 @@ public void setCharacterEncoding(String charset) {
 
 	/**
 	 * Automatically generate and validate etags
-	 * @ff.default <code>true</code> for repeatable responses
+	 * @ff.default <code>true</code>, can be changed by setting the property <code>api.etag.enabled</code>.
```
nielsm5, line 256:
```suggestion
	 * @ff.default <code>false</code>, can be changed by setting the property <code>api.etag.enabled</code>.
```

PR #5353, opened 2023-09-08
Improve robustness of CrlPipe by supporting Messages

gvanbrakel:


core/src/test/java/nl/nn/adapterframework/pipes/CrlPipeTest.java
```diff
@@ -0,0 +1,85 @@
+package nl.nn.adapterframework.pipes;
+
+import static nl.nn.adapterframework.testutil.MatchUtils.assertTestFileEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.io.InputStream;
+
+import org.junit.Test;
```
jkosternl, line 9: Can you please use JUnit 5 (imports)? We're trying to move to JUnit 5.

gvanbrakel: I am afraid that that is not reasonable in the context of this simple pull request. The base class PipeTestBase uses junit4. Using the annotation 'import org.junit.jupiter.api.Test;'in CrlPipeTest causes NPEs, as the '@Before' is not executed.

jkosternl: Ah yes, that's right! I didn't check that. We should probably create a new `ConfiguredTestBase` class, which uses JUnit 5 and we can gradually move forward. Anyways, that's another talk indeed.

nielsm5 (approved):

jkosternl (approved):

Issue #5287, opened 2023-08-24
SessionKey empty on 2nd interation when retrieved by HttpSender and not when retrieved by FixedResultPipe


Laurens-makel:
**❗ please do not add sensitive information in issues, you can provide extra information via email using issue number as reference ❗**

**Describe the issue**
We have an adapter which retrieves a large XML file using the `HttpSender` and iterates over the results using a `ForEachChildElementPipe` to call a sub adapter.

After upgrading to the 7.8.1 release, this adapter started giving strange errors:

`Caused by: nl.nn.adapterframework.core.SenderException: IbisLocalSender [UpdateArchivesAndSchemas] exception calling JavaListener [UpdateArchivesAndSchemas]: Pipe [Determine_Archive_Type] msgId [testmessage0ace0a6f--3b97492a_18a279551b5_afa] got null input`

After some investigation, it turned out that this occurred on the second iteration.  

I've decided to swap-out the `HttpSender` for a `FixedResultPipe`, which returns the exact same XML and suddenly the error does not occur anymore.

Main adapter:
![image](https://github.com/ibissource/iaf/assets/37303445/0d4baa01-9fcf-4bda-9721-40b71aea9a91)

Sub adapter, highlighted pipe throws exception on second iteration when XML is retrieved by http sender, not when retrieved by fixedresultpipe
![image](https://github.com/ibissource/iaf/assets/37303445/ecc61159-5299-4b79-8243-7b095754b223)

MLenterman:
We have a similar issue with ZaakBrug. For us it only happens with sessionkeys above ~500kb in size and only when overwriting the sessionkey. I have 2 Ladybug reports of the same adapter, one where the sessionkey is under 500kb and it succeeds and one where the sessionkey is above 500kb and it fails. Unfortunatly the report is too big to share on GitHub.

Laurens-makel:
> We have a similar issue with ZaakBrug. For us it only happens with sessionkeys above ~500kb in size and only when overwriting the sessionkey. I have 2 Ladybug reports of the same adapter, one where the sessionkey is under 500kb and it succeeds and one where the sessionkey is above 500kb and it fails. Unfortunatly the report is too big to share on GitHub.

Interesting, some differences:

In my case size does not seem to be of an issue, or haven't really tested it so to say, it sounds feasible that the XML I'm working with exceeds 500kb since it's quite big. Will check later to confirm. 

Another noticable difference is that in my case I set the sessionKey only once and never overwrite it.

[...]

PR #5357, opened 2023-09-08
Fix for Session Key close unexpectedly (7.8) (#5287)

tnleeuw:


jkosternl (approved):
I didn't read everything, since I expect it is the same code as into master.

tnleeuw:
> I didn't read everything, since I expect it is the same code as into master.

It is virtually the same code as in master indeed, although there are some minor differences to account for differences between 7.8 and master code bases.

I put the PR back into "DRAFT" however because in master the issue is still not entirely solved and I believe I now know what the remaining problem is, and we might as well fix it fully in 7.8 in a single PR then.

tnleeuw:
Oops, I forgot to run IAF-Test before marking the PR as "Ready for Review" again and it turns out we have a number of failures in there still that I need to investigate now.

tnleeuw:
> Oops, I forgot to run IAF-Test before marking the PR as "Ready for Review" again and it turns out we have a number of failures in there still that I need to investigate now.

I checked the results in Jenkins and it appears to be just Websphere flakiness, and locally on my laptop it appears to be a problem with Docker / ActiveMQ? Not able to connect to ActiveMQ running in the Docker image, which causes a number of tests to fail.

So the failing tests do not appear related to my actual changes.

core/src/test/java/nl/nn/adapterframework/http/cxf/SoapProviderStub.java
```diff
@@ -1,19 +1,34 @@
...
-
+			session.getCloseables().clear();
+		}
 		session = pipelineSession;
+		for (String key : session.keySet()) {
+			if (!(session.get(key) instanceof InputStream)) {
+				// Do not copy input-streams b/c then they are no longer readable
+				sessionCopy.put(key, session.getMessage(key).copyMessage());
+			}
+		}
```
nielsm5, line 31: Ik snap dit niet helemaal maar het is een (oude) test..

nielsm5 (approved):

core/src/test/java/nl/nn/adapterframework/http/cxf/SoapProviderStub.java
tnleeuw: > Ik snap dit niet helemaal maar het is een (oude) test..

De test probeert messages te lezen die in de session zitten, maar omdat de session gesloten is zijn die messages nu NULL.
Dus kopieer ik de session inhoud en maak een kopie van iedere session value.

Alleen niet als het een input stream is want dan lees je die stream en ben je hem kwijt. (Er is 1 test die een input stream in de session doet).

Ik had misschien ook alleen de instance-of-message values kunnen kopiëren.

PR #5362, opened 2023-09-11
Do not close any session keys in sub-adapter that are also in calling adapter

tnleeuw:
* Better fix for the issue of closed session keys after calling a sub-adapter, plus unit-tests that catch and demonstrate the problem.
* More complete fixes for closeables that appear in multiple sessions, and more tests added to catch different situations.

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/core/PipeLineSession.java
+++ b/core/src/main/java/nl/nn/adapterframework/core/PipeLineSession.java
@@ -109,10 +109,8 @@ public class PipeLineSession extends HashMap<String,Object> implements AutoClose
 	 *                   If an empty string then no keys will be copied.
 	 * @param from Child {@link PipeLineSession} from which keys are copied.
 	 * @param to Parent {@link PipeLineSession} or {@link Map}.
-	 * @param requester Tag of where the request to copy comes from so this can be logged when
-	 *                  closing any messages.
 	 */
-	public static void mergeToParentSession(String keysToCopy, PipeLineSession from, Map<String,Object> to, INamedObject requester) {
+	public static void mergeToParentSession(String keysToCopy, PipeLineSession from, Map<String,Object> to) {
 		if (to == null) {
 			return;
 		}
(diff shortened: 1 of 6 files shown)


jkosternl (approved):

tnleeuw:
There are failures in IAF-Test integration tests so this change probably needs a bit more work and investigation. Put to DRAFT for now.

jkosternl (approved):

core/src/main/java/nl/nn/adapterframework/core/PipeLineSession.java
```diff
@@ -172,6 +151,13 @@ public Object put(String key, Object value) {
 		return super.put(key, value);
 	}
 
+	@Override
+	public void putAll(Map<? extends String, ?> m) {
+		for (Entry<? extends String, ?> entry : m.entrySet()) {
+			put(entry.getKey(), entry.getValue());
+		}
+	}
```
nielsm5, line 159: Waarom doe je dit? De native hashmap doet het zelfde

core/src/main/java/nl/nn/adapterframework/core/PipeLineSession.java
```diff
@@ -107,10 +109,8 @@ public PipeLineSession(@Nonnull Map<String, Object> t) {
 	 *                   If an empty string then no keys will be copied.
 	 * @param from Child {@link PipeLineSession} from which keys are copied.
 	 * @param to Parent {@link PipeLineSession} or {@link Map}.
-	 * @param requester Tag of where the request to copy comes from so this can be logged when
-	 *                  closing any messages.
 	 */
-	public static void mergeToParentSession(String keysToCopy, PipeLineSession from, Map<String,Object> to, INamedObject requester) {
+	public static void mergeToParentSession(String keysToCopy, PipeLineSession from, Map<String,Object> to) {
```
nielsm5, line 113: Moet deze nog static zijn? from kan ook this zijn.

tnleeuw: Aangepast!

nielsm5 (approved):

PR #5388, opened 2023-09-14
Fix Locker numRetries not working

jacodg:


tnleeuw (approved):
The changes do look correct but it would be really nice to have some sort of test for this.
Either in a unit-test or in IAF-Test with the LockerPipelineProcessor.

nielsm5 (approved):

PR #5424, opened 2023-09-21
Bump commons-beanutils to 1.9.4 for every module/project

jkosternl:
Only the core bumped beanutils, while other modules where using old vulnerable versions still.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -968,11 +968,6 @@
 				<artifactId>jdom2</artifactId>
 				<version>2.0.6.1</version>
 			</dependency>
-			<dependency>
-				<groupId>commons-beanutils</groupId>
-				<artifactId>commons-beanutils</artifactId>
-				<version>1.9.4</version>
-			</dependency>
 		</dependencies>
 	</dependencyManagement>
 </project>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -955,19 +955,4 @@
 		</profile>
 	</profiles>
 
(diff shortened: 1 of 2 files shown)


tnleeuw (approved):

jkosternl:
> Why does the core project actually have a `<dependencyManagement>` section? Why is that not only in the parent POM?

The only reason I can think of, if there are specific module depedencies which you'd like to enforce only in that module. For all other situations, the parent pom is likely a better location.

nielsm5:
> Why does the core project actually have a `<dependencyManagement>` section? Why is that not only in the parent POM?

Staat in de code er boven:
`<!-- Lock transient dependency versions to fix CVE's -->`

jkosternl:
> > Why does the core project actually have a `<dependencyManagement>` section? Why is that not only in the parent POM?
> 
> Staat in de code er boven: `<!-- Lock transient dependency versions to fix CVE's -->`

Nee, @tnleeuw heeft gelijk. Dit is voor alle entries ook een probleem in andere modules. Ik ga ze verplaatsen.

nielsm5 (approved):

tnleeuw:
> > Why does the core project actually have a `<dependencyManagement>` section? Why is that not only in the parent POM?
> 
> Staat in de code er boven: `<!-- Lock transient dependency versions to fix CVE's -->`

Ja, maar waarom in core, niet in parent?

tnleeuw (approved):

PR #5430, opened 2023-09-21
Ability to suppress integrityCheck warnings globally

RayetSena:


nielsm5 (approved):
Oei, seems like in #1041 it was meant to be a global setting but looks like it wasn't tested properly...

tnleeuw (approved):

PR #5449, opened 2023-09-22
Restore the 'log.HideRegex' functionality

tnleeuw:


core/src/main/java/nl/nn/adapterframework/util/AppConstants.java
```diff
@@ -169,6 +169,7 @@ public void put(String key, String value) {
 	 */
 	private synchronized Object setProperty(String key, String value, boolean local) {
 		if(local) {
+			//noinspection deprecation
```
jkosternl, line 172: I don't understand this comment. Can you improve it?

jkosternl (approved):

tnleeuw:
Unexpected test failure:
```Error:  Failures: 
Error:    TestLogMessages.testHideRegexMatchInMessage:178 expected: <DEBUG - my beautiful log with <password>************</password> hidden value> but was: <DEBUG - my beautiful log with <password>TO BE HIDDEN</password> hidden value>```

The test worked for me locally when running the tests in IntelliJ. Need to check why it is now failing in Maven.

File: core/src/main/java/nl/nn/adapterframework/util/AppConstants.java (hunk 1 above)
Comment (tnleeuw) [reply core/src/main/java/nl/nn/adapterframework/util/AppConstants.java:172]: It's a comment for IntelliJ / Sonarlint that they should not highlight the use of a deprecated method.

File: core/src/test/java/nl/nn/adapterframework/logging/TestLogMessages.java [hunk 2]
```diff
@@ -165,6 +165,24 @@ public void testUnicodeCharactersInMessage() {
...
+	public void testHideRegexMatchInMessage() {
+		TestAppender appender = TestAppender.newBuilder().useIbisPatternLayout(PATTERN).build();
+		TestAppender.addToRootLogger(appender);
+		try {
+			log.debug("my beautiful log with <password>TO BE HIDDEN</password> hidden value");
+
+			List<String> logEvents = appender.getLogLines();
+			assertEquals(1, logEvents.size(), "found messages "+logEvents);
+			String message = logEvents.get(0);
+			assertEquals("DEBUG - my beautiful log with <password>************</password> hidden value", message);
```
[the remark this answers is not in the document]
tnleeuw: Log something like what?

Anyway the ticket was to restore some functionality that got accidentally lost and this is how it used to work. I have not changed the workings of the "hide" function.

tnleeuw:
> Unexpected test failure:
> 
> ```
> Error:  Failures: 
> Error:    TestLogMessages.testHideRegexMatchInMessage:178 expected: <DEBUG - my beautiful log with <password>************</password> hidden value> but was: <DEBUG - my beautiful log with <password>TO BE HIDDEN</password> hidden value>
> ```
> 
> The test worked for me locally when running the tests in IntelliJ. Need to check why it is now failing in Maven.

Test interdependencies should now be fixed.

core/src/main/java/nl/nn/adapterframework/util/AppConstants.java
jkosternl: Ah, didn't know this syntax.. Usually `@SuppresWarnings("xxxx")` is used for that, afaik.

core/src/test/java/nl/nn/adapterframework/logging/TestLogMessages.java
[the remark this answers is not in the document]
jkosternl: Ok, I see. Ah... I now see my comment was corrupted or it was hiding my real suggestion, between brackets :-) 
I meant: ...something like `<hidden>` instead?

core/src/main/java/nl/nn/adapterframework/util/AppConstants.java
tnleeuw: This is what IntelliJ inserted for me. I think this is what you get when you want to ignore the warning for just a single statement, instead of a method or class.

jkosternl: Ok fine.

core/src/test/java/nl/nn/adapterframework/logging/TestLogMessages.java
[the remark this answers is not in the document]
tnleeuw: Well, apart from it being functionality-as-is being restored, I think using `*` to hide your password, credit-card details, etc is common.
It also doesn't need to be machine-parseable. It's just the default regex was made to check for passwords in XML tags, presumably because at some point in the past, some XML messages were being logged that contained username/pwd tags.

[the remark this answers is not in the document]
jkosternl: Yeah, we can leave it as is. If we would design it as new, I'd to for another pattern in the logs instead.

core/src/test/resources/log4j4ibis.properties
nielsm5, line 1: Is deze file ook nodig? Hij pakt toch 'gewoon' de versie uit src/main/resources?

tnleeuw: In src/main/properties is dit property standaard uitgecomment en dat wou ik zo laten, omdat ik niet denk dat we standaard altijd built-in deze regex willen configureren.

nielsm5: Ok, dan moeten we de log.thread.rejectRegex ook standaard uit gaan zetten. Die doet nu (sinds het verwijderen van de WebControl configuration) niets meer.

nielsm5 (changes requested):

core/src/test/resources/log4j4ibis.properties
tnleeuw: Ik ga ernaar kijken.

nielsm5 (approved):

PR #5460, opened 2023-09-26
Allow unauthenticated OPTIONS requests

nielsm5:


tnleeuw (approved):
To be honest I have a hard time properly evaluating these changes for correctness from point of view of security. I do not know the APIs well enough for that and will need to study them.

I also feel we do not have proper tests for the security part, to actually verify that endpoints are secured correctly.

So I'm approving basically "on good faith".

jkosternl:
@nielsm5 Question: do authenticated OPTIONS request also (still?) work now?

nielsm5:
> @nielsm5 Question: do authenticated OPTIONS request also (still?) work now?

Yes, that behaviour hasn't changed. If you don't set the property you will have to authenticate regardless of the HTTP method used.

jkosternl (approved):

PR #5461, opened 2023-09-26
Optimize CompactSaxHandler class for performance and memory usage

jkosternl:


tnleeuw (changes requested):

core/src/main/java/nl/nn/adapterframework/util/CompactSaxHandler.java
```diff
@@ -94,64 +94,100 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
...
 		String lastElement = elements.get(lastIndex);
 
-		if (context != null
+		// Detection of MOVE_END token, while in sessionKeyFound state
+		int length = charDataBuilder.length();
+		if (state == MoveState.SESSION_KEY_FOUND && charDataBuilder.substring(length - 1, length).equals(VALUE_MOVE_END)) {
+			state = MoveState.STOPPED_REPLACING;
+		}
+
+		if (context != null && !startElement && state != MoveState.STOPPED_REPLACING
```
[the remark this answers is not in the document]
jkosternl: It is an optimization. This logic inside the if {} part, never needs to be performed if we're just starting with an element and no character data is received yet.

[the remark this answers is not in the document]
tnleeuw: The important part here being: no character data being received.

start/end element is not relevant here, in my opinion.

At startElement, we're calling this method to process character-data collected between that startElement, and the previous startElement.

core/src/test/java/nl/nn/adapterframework/util/CompactSaxHandlerTest.java
```diff
@@ -106,75 +110,112 @@ void testElementToMoveFeatureSessionKey() throws IOException, SAXException {
...
+	@ParameterizedTest
+	@ValueSource(strings = {
+			"/Util/CompactSaxHandler/input-invalid-moved-ref-1.xml",
+			"/Util/CompactSaxHandler/input-invalid-moved-ref-2.xml",
+			"/Util/CompactSaxHandler/input-invalid-moved-ref-3.xml",
+			"/Util/CompactSaxHandler/input-invalid-moved-ref-4.xml",
+			"/Util/CompactSaxHandler/input-invalid-moved-ref-5.xml",
+			"/Util/CompactSaxHandler/input-invalid-moved-ref-6.xml"
+	})
+	void testElementToMoveEdgeCases(String inputUri) throws IOException, SAXException {
```
nielsm5, line 222: Ik snap nog niet helemaal wat je hier tests? De input is overal verkeerd? Test je of hij de inhoud van het `identificatie` element zet (al staan er rare characters in?) ?

core/src/main/java/nl/nn/adapterframework/util/CompactSaxHandler.java
```diff
@@ -15,74 +15,74 @@
...
 	@Getter @Setter private boolean removeCompactMsgNamespaces = true;
 
-	private final StringBuilder messageBuilder = new StringBuilder();
 	private final StringBuilder charDataBuilder = new StringBuilder();
-	private final StringBuilder namespaceBuilder = new StringBuilder();
 	private final List<String> elements = new ArrayList<>();
 	@Setter private Map<String, Object> context = null;
+	private MoveState state = MoveState.DEFAULT;
+
+	private enum MoveState {
```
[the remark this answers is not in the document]
jkosternl: Adjusted, but needed to introduce a new boolean for some state tracking.

core/src/main/java/nl/nn/adapterframework/util/CompactSaxHandler.java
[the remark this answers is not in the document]
jkosternl: As discussed, kept it, because we need it.

jkosternl:
FYI @tnleeuw: the new code, as just committed, causes 3% performance loss on 100KB XML files and 0,8% loss on 1MB files, all with moving elements turned on.

core/src/main/java/nl/nn/adapterframework/util/CompactSaxHandler.java
```diff
@@ -165,13 +184,13 @@ private String elementsToString() {
 		return chain;
 	}
 
-	public String getXmlString() {
-		return messageBuilder.toString();
+	@Override
+	public String toString() {
+		return getContentHandler().toString();
 	}
```
nielsm5, line 190: Dit is niet zo handig denk ik?

nielsm5 (changes requested):

core/src/test/java/nl/nn/adapterframework/util/CompactSaxHandlerTest.java
tnleeuw: Tests die ik had toegevoegd voor de refactoring, om er zeker van te zijn dat "verkeerde" input niet leid tot nog verkeerdere output.

nielsm5: Dan is het misschien handig een check te doen op de session, dat daar niets bij komt bijvoorbeeld.

core/src/main/java/nl/nn/adapterframework/util/CompactSaxHandler.java
```diff
@@ -94,64 +91,86 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
...
-		if (chompLength >= 0 && charDataBuilder.length() > chompLength) {
-			before = "*** character data size [" + charDataBuilder.length() + "] exceeds [" + getChompCharSize() + "] and is chomped ***";
-			after = "...(" + (charDataBuilder.length() - chompLength) + " characters more)";
-			charDataBuilder.setLength(chompLength);
+		// Detection of MOVE_START and MOVE_END session key fragments, on end-Element phase
+		int length = charDataBuilder.length();
+		if (!startElement
+				&& length > VALUE_MOVE_START.length()
+				&& charDataBuilder.substring(length - 1, length).equals(VALUE_MOVE_END)
+				&& charDataBuilder.substring(0, VALUE_MOVE_START.length()).equals(VALUE_MOVE_START)) {
```
tnleeuw, line 118: These checks are quite expensive and only need to be done if `moveElementFound == true && session != null`, can be changed to a direct assignment:
```
moveElementFound = moveElementFound && session != null && !startElement && length > VALUE_MOVE_START.length()
 				&& charDataBuilder.substring(length - 1, length).equals(VALUE_MOVE_END)
 				&& charDataBuilder.substring(0, VALUE_MOVE_START.length()).equals(VALUE_MOVE_START)
```

core/src/main/java/nl/nn/adapterframework/util/CompactSaxHandler.java
```diff
@@ -15,74 +15,71 @@
...
-			if (isRemoveCompactMsgNamespaces()) {
-				attributeBuffer.append(attributes.getLocalName(i));
-			} else {
-				attributeBuffer.append(attributes.getQName(i));
-			}
-			attributeBuffer.append("=\"");
-			attributeBuffer.append(attributes.getValue(i));
-			attributeBuffer.append("\"");
-		}
+		moveElementFound = getElementToMove() != null && localName.equals(getElementToMove()) ||
```
tnleeuw, line 62: Also depends on `session != null`.

core/src/main/java/nl/nn/adapterframework/util/CompactSaxHandler.java
```diff
@@ -94,64 +91,86 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
...
+		if (!startElement
+				&& length > VALUE_MOVE_START.length()
+				&& charDataBuilder.substring(length - 1, length).equals(VALUE_MOVE_END)
+				&& charDataBuilder.substring(0, VALUE_MOVE_START.length()).equals(VALUE_MOVE_START)) {
+			moveElementFound = false;
 		}
 
-		int lastIndex = elements.size() - 1;
-		String lastElement = elements.get(lastIndex);
+		if (context != null && !startElement && moveElementFound) {
```
tnleeuw, line 122: With the above changes, can be simplified to `if (moveElementFound)`

tnleeuw (changes requested):

core/src/main/java/nl/nn/adapterframework/processors/InputOutputPipeProcessor.java
```diff
@@ -162,7 +163,7 @@ protected PipeRunResult processPipe(PipeLine pipeLine, IPipe pipe, Message messa
 				}
 
 				if (StringUtils.isNotEmpty(pe.getStoreResultInSessionKey())) {
-					if (log.isDebugEnabled()) log.debug("Pipeline of adapter ["+owner.getName()+"] storing result for pipe ["+pe.getName()+"] under sessionKey ["+pe.getStoreResultInSessionKey()+"]");
+					if (log.isDebugEnabled()) log.debug("Pipeline of adapter [{}] storing result for pipe [{}] under sessionKey [{}]", owner.getName(), pe.getName(), pe.getStoreResultInSessionKey());
```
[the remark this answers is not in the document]
jkosternl: Euh ja, inderdaad. Effe niet aan gedacht. Thanks.

core/src/main/java/nl/nn/adapterframework/util/CompactSaxHandler.java
jkosternl: Ok, added.

core/src/main/java/nl/nn/adapterframework/util/CompactSaxHandler.java
jkosternl: Ik had dat van `PrettyPrintFilter` overgenomen en dacht dat jullie dat een goed idee vonden. Ik maak er wel weer wat anders van.

core/src/test/java/nl/nn/adapterframework/util/CompactSaxHandlerTest.java
```diff
@@ -7,25 +7,31 @@
...
 import nl.nn.adapterframework.testutil.MessageTestUtils;
 import nl.nn.adapterframework.testutil.TestFileUtils;
+import nl.nn.adapterframework.xml.XmlWriter;
 
 class CompactSaxHandlerTest {
 
-	private static final String DEFAULT_BODY = "<root xmlns=\"urn:fakenamespace\" xmlns:soapenv=\"" + SOAPConstants.URI_NS_SOAP_1_1_ENVELOPE + "\"><attrib>1</attrib><attrib>2</attrib></root>";
-	private static final String DEFAULT_HEADER = "<soapenv:Header><info>234</info></soapenv:Header>";
+	private static final String DEFAULT_BODY = "<root xmlns=\"urn:fakenamespace\"><attrib>1</attrib><attrib>2</attrib></root>\n";
+	private static final String DEFAULT_HEADER = "<soapenv:Header><info>234</info></soapenv:Header>\n";
```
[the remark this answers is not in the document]
jkosternl: Ja, was nodig. De dubbele namespace is niet nodig en wordt er automatisch uitgefilterd, wat correct is. Deze statics geven een correct XML bericht weer, niet een experiment met duplicate informatie.

core/src/test/java/nl/nn/adapterframework/util/CompactSaxHandlerTest.java
jkosternl: Prima hoor. Toegevoegd.

core/src/main/java/nl/nn/adapterframework/util/CompactSaxHandler.java
jkosternl: I've changed the logic as you suggested, but with an important negation added.

I did also do a benchmark between the previous version (before applying these suggestions). 
The results are not as positive as what you've hoped for: 
100KB XML: 2,2% decrease in the first test, 0,25% decrease in the second test.
1MB XML: 0,8% decrease in the first test, 0,8% increase in the second test. 
Perhaps statistically not relevant, since it is lower then the error margin.

core/src/main/java/nl/nn/adapterframework/util/CompactSaxHandler.java
```diff
@@ -166,12 +186,11 @@ private String elementsToString() {
 	}
 
 	public String getXmlString() {
```
nielsm5, line 188: Deze mag nog weg, dan is hij goed!

tnleeuw: > Deze mag nog weg, dan is hij goed!

en fix van de larva tests.

jkosternl: Beide gedaan 👍

core/src/test/java/nl/nn/adapterframework/util/CompactSaxHandlerTest.java
```diff
@@ -174,6 +175,22 @@ void testElementToMoveChain() throws IOException, SAXException {
 		assertEquals("DC2022-012345", session.getString("ref_identificatie2"));
 	}
 
+	@Test
+	void testElementToMoveChainChomped() throws IOException, SAXException {
```
tnleeuw, line 179: It seems to me that this behaviour is wrong. We could move the element, but instead we chomp it and truncate data.

If that was old behaviour, I'm not sure if it really was behaviour that anyone wanted.

core/src/main/java/nl/nn/adapterframework/util/CompactSaxHandler.java
```diff
@@ -116,6 +116,7 @@ private void printCharData(boolean startElement) throws SAXException {
 		moveElementFound = moveElementFound
 				&& context != null
 				&& !startElement
+				&& (chompLength < 0 || length <= chompLength)
```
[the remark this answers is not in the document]
nielsm5: I agree, it would be a bit silly to chomp and move it... 😮

core/src/test/java/nl/nn/adapterframework/util/CompactSaxHandlerTest.java
nielsm5: Yup, that's why it's handy and important to test the class first before refactoring.

tnleeuw: I believe the original class did chomp the elements that were moved, and that changing this is what made the larva tests fail.

core/src/main/java/nl/nn/adapterframework/util/CompactSaxHandler.java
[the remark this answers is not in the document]
jkosternl: Please blame the persons who invented this. I just restored the way it worked before and spend some time figuring that out. 
I think it makes sense to define a max character size and use that for all character data. No exceptions. That's how it worked and how I now made it again. It is in line with the descriptions of Receiver e.g. 
So now, we don't chomp and move it @nielsm5. It just checks the length and if it is too long, it is not moved, but cut off. I think it actually is consistent and makes sense.

core/src/test/java/nl/nn/adapterframework/util/CompactSaxHandlerTest.java
jkosternl: The current implementation does not move elements that are chomped. And we did add tests before refactoring and covered all logic lines. But this is an odd combination/scenario, which happened to came out of the Larva tests. I don't like late notifications, so I've added this tests to unit tests.

core/src/main/java/nl/nn/adapterframework/util/CompactSaxHandler.java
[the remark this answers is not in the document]
jkosternl: Secondly, I think you both read the logic wrong. 
`&& (chompLength < 0 || length <= chompLength)`
So that means that it is fine if chompLength is not set OR character data size is <= chompLenght (if set).

core/src/test/java/nl/nn/adapterframework/util/CompactSaxHandlerTest.java
tnleeuw: I do indeed prefer this to be tested by unit tests, but it's behaviour that does not make sense to me.

If you want to move an element, you do so (in my view) to preserve its data and use it outside the XML. So an element-to-move is important to you, even if it's large.

Other large elements might not be so relevant to your process, so you might be happy to chomp them.

That's why I personally prefer the behaviour as we had it before the weekend, even if that broke Larva tests -- but the Larva tests tell is we are making a chance in behaviour so we have to make it consciously and explicitly, and also update the Larva tests. (Or perhaps we can remove them if they fully overlap with new unit tests? @nielsm5 would you agree with removing some Larva tests if the behaviour is now fully covered by unit tests? And the tests are not testing any kind of system integration?

core/src/main/java/nl/nn/adapterframework/util/CompactSaxHandler.java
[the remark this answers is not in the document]
tnleeuw: I'm not blaming you Jakobjob :-) I understand you are restoring previous behaviour, I just happen to think the previous behaviour does not make sense. :-)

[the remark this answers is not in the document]
jkosternl: Ah yes 😉 But if I read the above statements "then we effectively never move" and from Niels "it would be a bit silly to chomp and move it", I thought otherwise, because both are not true. It is just not moved if it is too long, which can be a valid safety measure. But yes, we're just thinking here about what the intentions of the functionality should be. I'll continue below.

core/src/test/java/nl/nn/adapterframework/util/CompactSaxHandlerTest.java
jkosternl: Ok, to wrap up, I think @tnleeuw 's idea sounds good. I suggest the following, up for both @nielsm5 and @tnleeuw approval:

1. Do NOT do a size check (chomping) when we're processing the 'element to move' (same for move chain). So, I'll revert the single added line to the CompactSaxHandler class, I've added to 'fix' it.
2. Make unit tests compatible with the above statement.
3. Remove the Larva tests who failed, with the wrong chomping check. Still keep a few Larva tests, to test the basic behavior, because it is used by the Receiver etc, which is a good idea to keep (integrated) tested. 

Both agree? I will wait for both approvals before doing anything (and revert it later).

tnleeuw: Sounds good to me.

jkosternl: As discussed and agreed today at the office, we make it like described in the 3 points above.

tnleeuw (changes requested):
Bijna, maar er moet volgens mij nog 1 Larva test weg.

test/src/main/configurations/MainConfig/ConfigurationCompactMessage.xml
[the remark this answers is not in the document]
jkosternl: Thanks. out.xml aangepast.

tnleeuw (approved):

nielsm5 (approved):

Issue #5487, opened 2023-10-03
Remove BTM - an old Transaction Manager


jkosternl:
Since FF! is going to Java 11, cleanups are taking place as well. BTM is leaving the boat. Dep was dated December 2016. Narayana TM is still there. 

Merge after #5485 is brought to master.

PR #5488, opened 2023-10-03
Remove BTM, an old Transaction Manager (#5487)

jkosternl:


core/src/test/java/nl/nn/adapterframework/receivers/ReceiverTest.java
```diff
@@ -228,19 +224,10 @@ public ITransactionalStorage<Serializable> setupErrorStorage() {
 
 	public static Stream<Arguments> transactionManagers() {
 		return Stream.of(
-			Arguments.of(supplier(ReceiverTest::buildNarayanaTransactionManagerConfiguration)),
-			Arguments.of(supplier(ReceiverTest::buildBtmTransactionManagerConfiguration))
+			Arguments.of(supplier(ReceiverTest::buildNarayanaTransactionManagerConfiguration))
```
tnleeuw, line 227: Since there's still a chance that we might add new transaction manager implementations in future I think we should keep these tests parameterised.

test/src/test/testtool/TransactionHandling/scenario02.properties
```diff
@@ -17,6 +17,6 @@ ignoreContentBetweenKeys.interruptedException.key2=</message>
 replaceRegularExpressionKeys.msgWas.key1=<message>null</message>
 replaceRegularExpressionKeys.msgWas.key2=<message>REPLACED</message>
 
-#TOMCATBTM/TOMCAT6
+#TOMCAT6
```
tnleeuw, line 20: Still `TOMCAT6`? Shall we update this comment to `TOMCAT9` for now? 😉 

(And soon perhaps 10 or 11)

tnleeuw (commented):
This PR is going to break the CI after it is merged, because the CI still tries to run tests with BTM.

So we'll need to set up separate CI for 7.9 and master pretty soon.

test/src/test/testtool/TransactionHandling/scenario02.properties
jkosternl: A version number does not make sense in this property file. Removing it.

core/src/main/java/nl/nn/adapterframework/lifecycle/IbisApplicationInitializer.java
```diff
@@ -114,21 +113,6 @@ public WebApplicationContext initWebApplicationContext(ServletContext servletCon
 		}
 	}
 
-	//TODO: remove this in 8.0
```
tnleeuw, line 117: Perhaps a bit tricky to remove this in the PR for removing BTM although I agree that this code should be removed.
Perhaps better to do that in another PR, there's a lot more WAS related code that needs to be removed.

tnleeuw (changes requested):

core/src/main/java/nl/nn/adapterframework/lifecycle/IbisApplicationInitializer.java
jkosternl: I agree. Done! #5525

tnleeuw (approved):
In principle this PR looks OK to me now.

nielsm5 (approved):

PR #5490, opened 2023-10-03
Remove Rhino and Nashorn Javascript Engines

jkosternl:


core/src/main/java/nl/nn/adapterframework/senders/JavascriptSender.java
```diff
@@ -68,17 +66,17 @@ public class JavascriptSender extends SenderSeries {
 	private String fileInput;
 
 	public enum JavaScriptEngines {
-		J2V8(J2V8.class), NASHORN(Nashorn.class), RHINO(Rhino.class);
+		J2V8(J2V8.class);
```
tnleeuw, line 69: Does it make sense to maintain support for multiple JS engines?
Or is the effort to take that out too great?

Should we do anything to facilitate loading old configurations with values of "Rhino" or "Nashorn" in here, which are now no longer valid enum values?

I think that unless we foresee adding new JS engines in future we should at least have a separate PR for 7.9 to make the `engine` config property of the JavascriptSender `@Deprecated` and with a `@ConfigurationWarning` attached.

test/src/main/configurations/MainConfig/ConfigurationJavascript.xml
```diff
@@ -15,7 +15,7 @@
 					<sender className="nl.nn.adapterframework.senders.JavascriptSender"
 						jsFileName="JavaScript/JavascriptTest.js"
 						jsFunctionName="f2"
-						engineName="Rhino">
+						engineName="J2V8">
```
tnleeuw, line 18: I think this change should be in 7.9 proper already, and any other changes to the test-JS file needed to make it work (along with change to deprecate this configuration property).

tnleeuw (changes requested):

core/src/test/java/nl/nn/adapterframework/extensions/javascript/JavascriptSenderCallbackTest.java
```diff
@@ -36,7 +36,7 @@ public class JavascriptSenderCallbackTest extends SenderTestBase<JavascriptSende
 
 	@Parameters(name = "{0}")
 	public static Collection<Object[]> data() {
-		return Arrays.asList(new Object[][] {{JavaScriptEngines.J2V8}, {JavaScriptEngines.NASHORN}});
+		return Arrays.asList(new Object[][] {{JavaScriptEngines.J2V8}});
```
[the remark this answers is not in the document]
jkosternl: As discussed today at the office, we keep it, because for #5496, we're going to add a new Javascript engine, likely.

core/src/test/java/nl/nn/adapterframework/extensions/javascript/JavascriptSenderTest.java
```diff
@@ -28,7 +27,7 @@
 @RunWith(Parameterized.class)
 public class JavascriptSenderTest extends SenderTestBase<JavascriptSender> {
 
-	@Parameterized.Parameter(0)
+	@Parameterized.Parameter()
```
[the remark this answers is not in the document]
jkosternl: As discussed today at the office, we keep it, because for #5496, we're going to add a new Javascript engine, likely.

core/src/main/java/nl/nn/adapterframework/senders/JavascriptSender.java
jkosternl: Done inside #5517

tnleeuw (approved):

nielsm5 (approved):

PR #5508, opened 2023-10-04
Fix default-values in log configuration properties

tnleeuw:
Fix the regex for log configuration properties so that it handles the defaults in logfiles.


jkosternl (approved):

nielsm5 (approved):
Super!

PR #5513, opened 2023-10-05
Handle console servlet exceptions

nielsm5:


console/frontend/src/main/java/nl/nn/adapterframework/console/ConsoleFrontend.java
```diff
@@ -117,13 +121,12 @@ private void doGetSafely(HttpServletRequest req, HttpServletResponse resp) throw
...
 		try(InputStream in = resource.openStream()) {
 			IOUtils.copy(in, resp.getOutputStream());
+
+			resp.flushBuffer();
 		} catch (IOException e) {
-			log.warn("error reading resource [{}]", resource, e);
-			resp.sendError(500);
-			return;
+			// Either something has gone wrong, or the request has been cancelled
+			log.debug("error reading resource [{}]", resource, e);
```
tnleeuw, line 128: I understand that sending an error gave itself an error in some conditions but should the logging perhaps still be at level WARN?

And should we perhaps still try to send an error but with a try-catch in case part of the response was already written? (I don't know if this can be detected in a clean way).

nielsm5: I have no clue how we could implement that. Technically we only care about the IOException thrown by the inputStream but the most likely thing to cause this exception is the outputstream because the request has been cancelled.

tnleeuw (approved):

tnleeuw (approved):

PR #5516, opened 2023-10-05
add unregister for all rootscope $on handlers

Matthbo:
Also fixed various small issue found during testing
* inlinestore giving an error because appService wasnt being injected
* added $onInit to configurations-upload, this might not have been an actual issue but stability and consistency is important

Fixes issue found by @nielsm5 where moving from 'monitors' to 'manage configurations' would trigger monitors' rootscope$on which caused the router to transition back to 'monitors' page


nielsm5 (approved):
To be honest I have a hard time properly evaluating these changes for correctness. I do not know the new APIs well enough for that and will need to study them.

philipsens (approved):

nielsm5 (approved):
Timing ;)

Issue #5515, opened 2023-10-05
Deprecate JavaScript Engines: Nashorn and Rhino


jkosternl:
In FF! version 8.0, both Rhino and Nashorn JavaScript engines are deleted, see #5490. 
In v7.9 they should be marked as deprecated and tests should use J2V8 as default instead.

PR #5517, opened 2023-10-05
Deprecate Rhino and Nashorn JavaScript engines (#5515)

jkosternl:


core/src/main/java/nl/nn/adapterframework/senders/JavascriptSender.java
```diff
@@ -68,7 +69,7 @@ public class JavascriptSender extends SenderSeries {
 	private String fileInput;
 
 	public enum JavaScriptEngines {
-		J2V8(J2V8.class), NASHORN(Nashorn.class), RHINO(Rhino.class);
+		J2V8(J2V8.class), @Deprecated NASHORN(Nashorn.class), @Deprecated RHINO(Rhino.class);
```
tnleeuw, line 72: Can you add `@ConfigurationWarning` on the deprecated elements too? Or will that not be displayed in the ConfigurationWarnings anyway?

core/src/main/java/nl/nn/adapterframework/senders/JavascriptSender.java
```diff
@@ -207,9 +208,12 @@ public void setJsFunctionName(String jsFunctionName) {
 	}
 
 	/**
-	 * the name of the javascript engine to be used
+	 * the name of the JavaScript engine to be used.
+	 * @deprecated Both Nashorn and Rhino are deprecated. Use J2V8 instead.
```
tnleeuw, line 212: Please add that Nashorn and Rhine will be removed in 8.0.

core/src/main/java/nl/nn/adapterframework/senders/JavascriptSender.java
```diff
@@ -207,9 +208,12 @@ public void setJsFunctionName(String jsFunctionName) {
...
 	 * @ff.default J2V8
 	 */
+	@Deprecated
+	@ConfigurationWarning("JavaScript engines Nashorn and Rhino deprecated. Use \"J2V8\" instead")
```
tnleeuw, line 216: Please also add here that Nashorn and Rhine will be deleted in 8.0

tnleeuw (changes requested):

tnleeuw:
Can you please also add on `RhinoPipe` the `@ConfigurationWarning` element that it will be deleted in 8.0 and instead the `JavascriptSender` should be used?
That way anyone still using it will get a more direct warning about this.

core/src/main/java/nl/nn/adapterframework/senders/JavascriptSender.java
tnleeuw: As discussed in standup, not needed.

core/src/main/java/nl/nn/adapterframework/senders/JavascriptSender.java
tnleeuw: As discussed in standup, not needed.

jkosternl: Thanks

core/src/main/java/nl/nn/adapterframework/senders/JavascriptSender.java
jkosternl: I'll add it anyways.. is faster than checking if it is being used. We'll remove it in 8.0 anyways.

tnleeuw (approved):

core/src/main/java/nl/nn/adapterframework/senders/JavascriptSender.java
```diff
@@ -67,10 +68,18 @@ public class JavascriptSender extends SenderSeries {
...
+
 	public enum JavaScriptEngines {
-		J2V8(J2V8.class), NASHORN(Nashorn.class), RHINO(Rhino.class);
+		J2V8(J2V8.class),
+
+		@ConfigurationWarning("Engine Nashorn is deprecated. Use J2V8 instead.")
+		@Deprecated NASHORN(Nashorn.class),
+
+		@ConfigurationWarning("Engine Rhino is deprecated. Use J2V8 instead.")
+		@Deprecated RHINO(Rhino.class);
```
nielsm5, line 79:
```suggestion
		@Deprecated NASHORN(Nashorn.class),

		@Deprecated RHINO(Rhino.class);
```
Just the deprecated tag is enough here to 'hide' the element, the ConfigurationWarning should be on the setter element instead.

PR #5539, opened 2023-10-10
Show error while parallel attribute is true in ForEachChildElementPipe

mericakgul:
- We were reported that error message is not reported even though there is an error in one of the threads that are created thanks to the " parallel = 'true' " attribute in ForEachChildElementPipe.
- In this PR, we throw an exception in case an error occurs during parallel execution in order to be able to report it back.
- This is the issue reported in zaakbrug repo: https://github.com/wearefrank/zaakbrug/issues/144

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/pipes/IteratingPipe.java
+++ b/core/src/main/java/nl/nn/adapterframework/pipes/IteratingPipe.java
@@ -417,7 +417,7 @@ public abstract class IteratingPipe<I> extends MessageSendingPipe {
 					List<Throwable> exceptions = new ArrayList<>();
 					for (ParallelSenderExecutor pse : executorList) {
 						count++;
-						String itemResult="";
+						String itemResult;
 						if (pse.getThrowable() == null) {
 							SenderResult senderResult = pse.getReply();
 							if (senderResult.isSuccess()) {
--- a/core/src/main/java/nl/nn/adapterframework/pipes/IteratingPipe.java
+++ b/core/src/main/java/nl/nn/adapterframework/pipes/IteratingPipe.java
@@ -414,18 +414,34 @@ public abstract class IteratingPipe<I> extends MessageSendingPipe {
 				try {
 					guard.waitForAllResources();
 					int count = 0;
+					List<Throwable> exceptions = new ArrayList<>();
 					for (ParallelSenderExecutor pse : executorList) {
 						count++;
-						String itemResult;
+						String itemResult="";
 						if (pse.getThrowable() == null) {
 							SenderResult senderResult = pse.getReply();
 							if (senderResult.isSuccess()) {
 								itemResult = senderResult.getResult().asString();
-							} else {
-								throw new SenderException("Something went wrong during parallel execution of iterating pipe: " + senderResult.getResult().asString());
+							} 
+							else {
+								if(isIgnoreExceptions()){
+									itemResult = "<exception>"+ XmlEncodingUtils.encodeChars(senderResult.getResult().asString())+"</exception>";
+								} else {
+									SenderException se = new SenderException(senderResult.getResult().asString());
+									exceptions.add(se);
+								}
 							}
 						} else {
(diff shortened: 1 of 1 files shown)


core/src/main/java/nl/nn/adapterframework/pipes/IteratingPipe.java
```diff
@@ -422,10 +422,10 @@ public void waitForResults() throws SenderException, IOException {
 							if (senderResult.isSuccess()) {
 								itemResult = senderResult.getResult().asString();
 							} else {
-								itemResult = "<exception>"+ XmlEncodingUtils.encodeChars(senderResult.getResult().asString())+"</exception>";
```
nielsm5, line 425: You can't just change this behavior. There are many applications using this which will break if we merge this as is.

Please make this behavior configurable, with an attribute for example `throwExceptions="true"`.

Also don't forget to make a few tests for this. To test both the old and new behavior 😃.

nielsm5 (changes requested):

core/src/main/java/nl/nn/adapterframework/pipes/IteratingPipe.java
```diff
@@ -414,18 +414,34 @@ public void waitForResults() throws SenderException, IOException {
...
 								itemResult = senderResult.getResult().asString();
-							} else {
-								itemResult = "<exception>"+ XmlEncodingUtils.encodeChars(senderResult.getResult().asString())+"</exception>";
+							} 
+							else {
+								if(isIgnoreExceptions()){
+									itemResult = "<exception>"+ XmlEncodingUtils.encodeChars(senderResult.getResult().asString())+"</exception>";
+								} else {
+									SenderException se = new SenderException(senderResult.getResult().asString());
+									exceptions.add(se);
```
nielsm5, line 431: Although it's quite minor, this could be on a single line.
```suggestion
									exceptions.add(new SenderException(senderResult.getResult().asString()));
```

core/src/main/java/nl/nn/adapterframework/pipes/IteratingPipe.java
```diff
@@ -414,18 +414,34 @@ public void waitForResults() throws SenderException, IOException {
...
 						} else {
-							itemResult = "<exception>"+ XmlEncodingUtils.encodeChars(pse.getThrowable().getMessage())+"</exception>";
+							if(isIgnoreExceptions()){
+								itemResult = "<exception>"+ XmlEncodingUtils.encodeChars(pse.getThrowable().getMessage())+"</exception>";
+							} else {
+								exceptions.add(pse.getThrowable());
+							}
+						}
+						if(!exceptions.isEmpty()){
+							SenderException se = new SenderException("Something went wrong in one of the parallel running pipes: ");
```
nielsm5, line 442: It's not always a pipe that's being executed, and you typically never want the exception message to be left on a cliffhanger (using `: ` without the actual reason).

```suggestion
							SenderException se = new SenderException("an error occurred during parallel execution");
```

core/src/main/java/nl/nn/adapterframework/pipes/IteratingPipe.java
```diff
@@ -414,18 +414,34 @@ public void waitForResults() throws SenderException, IOException {
 				try {
 					guard.waitForAllResources();
 					int count = 0;
+					List<Throwable> exceptions = new ArrayList<>();
 					for (ParallelSenderExecutor pse : executorList) {
 						count++;
-						String itemResult;
+						String itemResult="";
```
nielsm5, line 420: Why are you assigning this variable? It shouldn't be necessarily?
```suggestion
						String itemResult;
```

nielsm5 (changes requested):
Althought the solution is not wrong, I'm missing at least the non-backwards compatible notification *(`releases.md`) to warn other developers that this pipe now suddenly can throw exceptions which it did not do in previous versions.

nielsm5:
I've added more tests (#5954), I'll try and see how we can implement this.

nielsm5:
Superseded by #5961

PR #5563, opened 2023-10-13
Jari webservices to typescript

Jarivanoort:


console/frontend/src/main/frontend/src/angularjs/app/views/webservices/webservices.component.ts
```diff
@@ -0,0 +1,29 @@
...
+  $onInit() {
+    this.rootURL = this.Misc.getServerPath();
+
+    this.Api.Get("webservices", (data) => {
+      Object.assign(this, data);
+    });
+  };
+
+  compileURL(apiListener: any) {
+    return this.rootURL + "iaf/api/webservices/openapi.json?uri=" + encodeURI(apiListener.uriPattern);
```
philipsens, line 22: I would love to see this as an environment variable

console/frontend/src/main/frontend/src/angularjs/app/views/webservices/webservices.component.ts
```diff
@@ -0,0 +1,29 @@
...
+  };
+};
+
+appModule.component('webservices', {
+  controller: ['Api', 'Misc', WebservicesController],
+  templateUrl: 'js/app/views/webservices/webservices.component.html'
```
philipsens, line 28: Same for this one

console/frontend/src/main/frontend/src/app/views/webservices/webservices.component.ts
```diff
@@ -0,0 +1,49 @@
...
+  ) { };
+
+  ngOnInit() {
+    this.apiService.Get("webservices", (data) => {
+      Object.assign(this, data);
+    });
+  };
+
+  compileURL(apiListener: any) {
+    return this.rootURL + "iaf/api/webservices/openapi.json?uri=" + encodeURI(apiListener.uriPattern);
```
philipsens, line 47: Same for this one

console/frontend/src/main/frontend/src/angularjs/app/views/webservices/webservices.component.ts
Matthbo: This is the angularjs file that is now converted to ts and later copied over to the angular version of this component, also this is a template url thats not dynamic in any way shape or form, this wont ever change anymore since its only necessary to load the angularjs version of this component if needed to and will be removed once everything has moved over to angular

console/frontend/src/main/frontend/src/angularjs/app/views/webservices/webservices.component.ts
Matthbo: Great idea for after we've upgraded everything to angular

philipsens: It's so little effort, tho.

Procrastinating on these kinds of things will only add to the refactoring process.

console/frontend/src/main/frontend/src/angularjs/app/views/webservices/webservices.component.ts
philipsens: Alright, thanks for clarification.

console/frontend/src/main/frontend/src/angularjs/app/views/webservices/webservices.component.ts
Matthbo: It's only used twice, once in the component logic and once in the template
I do agree that it should be in the environment variables however I dont really see it as part of the upgrade process we're doing right now

Matthbo (commented):
Please make sure all files use the right editorconfig settings
```
indent_style = space
indent_size = 2
```
webservices.component.html uses tabs with size 4 in this pr

console/frontend/src/main/frontend/src/angularjs/app/views/webservices/webservices.component.ts
philipsens: In my opinion, small things like this are next to no work, but can be overlooked in the future. You make a good point, stating that this is not part of the upgrade.

Matthbo (approved):

PR #5577, opened 2023-10-16
Fix 'default' legacy properties overwriting new properties

nielsm5:


security/src/main/java/nl/nn/adapterframework/lifecycle/servlets/ActiveDirectoryAuthenticator.java
```diff
@@ -59,15 +61,19 @@ private void configure() throws FileNotFoundException {
...
 	}
 
-	/** Set default values for legacy properties to ease application upgrades, these values will be overwritten when the appropiate setter is called. */
+	/** Set default values for legacy properties to ease application upgrades, these values can be overwritten when the appropriate setter is called. */
 	private void setDefaultValues() {
 		Environment env = getApplicationContext().getEnvironment();
-		url = env.getProperty("ldap.auth.url");
-		baseDn = env.getProperty("ldap.auth.user.base");
+		String legacyURL = env.getProperty("ldap.auth.url");
+		if(StringUtils.isEmpty(url) && StringUtils.isNotBlank(legacyURL)) {
```
tnleeuw, line 70: Why `isEmpty()` in one check, and `isNotBlank()` in the other?

security/src/main/java/nl/nn/adapterframework/lifecycle/servlets/ActiveDirectoryAuthenticator.java
```diff
@@ -59,15 +61,19 @@ private void configure() throws FileNotFoundException {
...
+			this.url = legacyURL;
+		}
+		String legacyBaseDn = env.getProperty("ldap.auth.user.base");
+		if(StringUtils.isEmpty(baseDn) && StringUtils.isNotBlank(legacyBaseDn)) {
```
tnleeuw, line 74: Idem

security/src/main/java/nl/nn/adapterframework/lifecycle/servlets/ActiveDirectoryAuthenticator.java
nielsm5: To prevent the legacyUrl overwriting the new url property by existing but set as an empty/whitespace value.

tnleeuw: Is it then OK to set an empty/whitespace value in the URL property in the configuration, to ignore legacy properties?
I was wondering about that side more than the legacy-properties.

nielsm5: My intent is for users which are already using the legacy property's that it should work as-is. And if they add the new properties, it should ignore the legacy ones.
Even in the case of 'attempting' to set the new property but failing to properly do so should have preference over the default.

tnleeuw (approved):

PR #5592, opened 2023-10-18
Use unbounded threadpool executor

tnleeuw:
Instead of a simple task executor that creates a new thread for every task, allow recycling threads using the threadpool executor with a configuration that does not limit the number of threads.

### DIFF
--- a/core/src/main/resources/SpringApplicationContext.xml
+++ b/core/src/main/resources/SpringApplicationContext.xml
@@ -257,10 +257,11 @@
 
 	<bean
 		name="taskExecutor"
-		class="org.springframework.core.task.SimpleAsyncTaskExecutor"
+		class="org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor"
 		scope="singleton"
 	>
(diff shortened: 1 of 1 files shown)


jkosternl:
Sounds like a good idea to change. But what about the other defaults? E.g. what is written here: https://www.quinbay.com/blog/working-with-threadpooltaskexecutor-of-spring they also configure corePoolSize, which defaults to 1! 

See https://javadoc.io/doc/org.springframework/spring-context/5.3.30/org/springframework/scheduling/concurrent/ThreadPoolTaskExecutor.html

tnleeuw:
> Sounds like a good idea to change. But what about the other defaults? E.g. what is written here: https://www.quinbay.com/blog/working-with-threadpooltaskexecutor-of-spring they also configure corePoolSize, which defaults to 1!
> 
> See https://javadoc.io/doc/org.springframework/spring-context/5.3.30/org/springframework/scheduling/concurrent/ThreadPoolTaskExecutor.html

In my understanding the core pool size just configured the minimum size of the pool. It doesn't limit the number of threads in the pool.

PR #5593, opened 2023-10-19
Allow public endpoints

nielsm5:


jkosternl (approved):
Do we need to document in configuration files that if an URL starts with ! it is a public endpoint?

nielsm5:
Yes, we need to document all authentication/authorization options. For now this will be added to the Frank!Manual, as I don't think people using the framework will go through the source code.

tnleeuw (approved):

jkosternl:
> Yes, we need to document all authentication/authorization options. For now this will be added to the Frank!Manual, as I don't think people using the framework will go through the source code.

Yes, I was thinking about example configuration, where you also add an example of a public endpoint.

tnleeuw (approved):

PR #5610, opened 2023-10-24
Fix deadlock with stopping receiver when it is still starting

tnleeuw:


tnleeuw:
This PR probably also fixes the problems with 2 of the tests which timed out creating Mockito Mocks after updates.

core/src/test/java/nl/nn/adapterframework/receivers/SlowListenerWithPollGuard.java
```diff
@@ -25,21 +27,26 @@ public class SlowListenerWithPollGuard extends SlowPushingListener implements IP
...
+		mockConnector = mock(SpringJmsConnector.class);
+		when(mockConnector.getLastPollFinishedTime()).thenAnswer(invocationOnMock -> System.currentTimeMillis() - mockLastPollDelayMs);
+		when(mockConnector.getThreadsProcessing()).thenReturn(new Counter(0));
+		when(mockConnector.getReceiver()).thenReturn(receiver);
+		when(mockConnector.getListener()).thenReturn(this);
+		when(mockConnector.getJmsContainer()).thenReturn(mockContainer);
+		when(mockConnector.getLogPrefix()).thenReturn("MockJmsConnector ");
+	}
+
+	@Override
```
tnleeuw, line 44: These changes should fix the 2 tests that caused trouble after updating Mockito.

1. The mocks were created in a different thread
2. The mocks were created while opening the Receiver, which was done under a TimeoutGuard.
3. After updating Mockito it took forever to create the mocks from this thread. Probably a Mockito bug.
4. This triggered the TimeoutGuard and thus it failed the test.

Now the timeout guard should not be triggered anymore because `configure()` doesn't run under the TimeoutGuard, and the mocks are now created in the main JUnit thread again so this should not trigger the Mockito issue.

nielsm5 (approved):

PR #5613, opened 2023-10-24
Fix "replyTo" and "bounceAddress" options not working in MailSender

tnleeuw:


core/src/main/java/nl/nn/adapterframework/senders/MailSender.java
```diff
@@ -257,6 +254,20 @@ private MimeMessage createMessage(Session session, MailSession mailSession, Stri
 			throw new SenderException("Error occurred while setting sender email", e);
 		}
 
+		if (mailSession.getReplyTo() != null) {
+			try {
+				msg.setReplyTo(new jakarta.mail.internet.InternetAddress[]{mailSession.getReplyTo().getInternetAddress()});
+			} catch (Exception e) {
+				throw new SenderException("Error occurred while setting replyTo email", e);
+			}
+		} else if (StringUtils.isNotBlank(mailSession.getBounceAddress())) {
```
tnleeuw, line 263: NB: I decided here to give priority to `replyTo` if set in the message, over the `bounceAddress` which may have been configured in the MailSender but I'm not sure if that is the right priority.

PR #5639, opened 2023-10-27
Raise message.size.warn.default to 30M

nielsm5:


tnleeuw (changes requested):
Please also update JavaDoc in Pipeline.java:774 to match new default.

tnleeuw (approved):

PR #5665, opened 2023-10-31
Update servicemix SAAJ

nielsm5:
We need to figure out if we want to continue with SHA1 digests, newer versions seem to no longer support this.


tnleeuw (changes requested):
Some tests now broken -- this needs to be fixed before we can proceed.


As discussed, we need to find another SAAJ implementation.

tnleeuw:
By updating also
```
		<dependency><!-- javax based dependency -->
			<groupId>jakarta.xml.soap</groupId>
			<artifactId>jakarta.xml.soap-api</artifactId>
			<version>3.0.1</version>
			<exclusions>
				<exclusion>
					<groupId>jakarta.activation</groupId>
					<artifactId>jakarta.activation-api</artifactId>
				</exclusion>
			</exclusions>
		</dependency>
```
from 1.4.2 to 3.0.1 in core/pom.xml the SAAJ lib will work in both JDK11 and JDK17, resolving almost all immediate JDK17 compatibility problems.

nielsm5:
> By updating also
> 
> ```
> 		<dependency>
> 			<groupId>jakarta.xml.soap</groupId>
> 			<artifactId>jakarta.xml.soap-api</artifactId>
> 			<version>3.0.1</version>
> 			<exclusions>
> 				<exclusion>
> 					<groupId>jakarta.activation</groupId>
> 					<artifactId>jakarta.activation-api</artifactId>
> 				</exclusion>
> 			</exclusions>
> 		</dependency>
> ```
> 
> from 1.4.2 to 3.0.1 in core/pom.xml the SAAJ lib will work in both JDK11 and JDK17, resolving almost all immediate JDK17 compatibility problems.

Not sure why we have this dependency but I rather fix the Java 17 issues in #5677.

tnleeuw (approved):

PR #5668, opened 2023-11-01
Remove 2nd copy of FilenameUtils.java.

tnleeuw:


jkosternl (approved):
Approved, if it is nowhere used? How strange...

PR #5682, opened 2023-11-03
More Dependency updates for JDK17

tnleeuw:
I believe that with these extra dependency-updates at least the Maven build works with JDK17.


nielsm5:
Perhaps you can also enable the jdk17 build in the github actions script?

nielsm5 (approved):

PR #5692, opened 2023-11-06
Update dependabot yml

nielsm5:


nielsm5:
Is only required in the default branch.

PR #5700, opened 2023-11-07
Upgrade the IAF-Test JBoss Docker test image to 7.4.0

philipsens:
I couldn't upgrade further because the elytron tool didn't work in the newer versions that I tried. We need that for creating a credential store.

In the future, we could do this in an extra build step or by using the tool in the CI.


nielsm5 (approved):
Er van uit gaande dat de CI straks groen is 😄 .

tnleeuw (approved):

PR #5725, opened 2023-11-09
Remove XML logging functionality

nielsm5:
I've kept the change relatively simple, without changing the functionality too much.
The FileViewerServlet needs to be replaced with an endpoint to transfer data over the SpringBus. For now this will remain to work with the console, even without the frontend changes.


tnleeuw (approved):
0% test coverage... Dat is ook knap. 😬

PR #5726, opened 2023-11-09
Fix zip with empty files added by sub-adapters

tnleeuw:
Tested that without this fix, the included IAF-Test scenario produces a ZIP file with only 0-byte files, like in the original issue. The scenario produces only `<result></result>` as output, manually checking the ZIP file shows 0 byte files.

With the fix the ZIP file contains the actual data it is supposed to.

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/compression/ZipWriter.java
+++ b/core/src/main/java/nl/nn/adapterframework/compression/ZipWriter.java
@@ -98,6 +98,7 @@ public class ZipWriter implements ICollector<MessageZipEntry> {
 			if (StringUtils.isEmpty(filename) && contents != input) {
 				filename = input.asString();
 			}
+			session.unscheduleCloseOnSessionExit(contents);
 			MessageZipEntry entry = new MessageZipEntry(contents, filename);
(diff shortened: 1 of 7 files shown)


core/src/main/java/nl/nn/adapterframework/compression/ZipWriter.java
```diff
@@ -98,6 +98,7 @@ public MessageZipEntry createPart(Message input, PipeLineSession session, Parame
 			if (StringUtils.isEmpty(filename) && contents != input) {
 				filename = input.asString();
 			}
+			session.unscheduleCloseOnSessionExit(contents);
```
tnleeuw, line 101: This line is the actual fix.

It could also be called in the Collection class when adding to the collection, or in the CollectorPipeBase.

Not sure what really is best, I decided to add this here because it's the ZipWriter which actually uses the files and does not consume the contents until the time of building. Another Collector implementation might consume the files early and then this is not needed.

tnleeuw:
The issues that Sonar sees are in files that are not part of this PR.

nielsm5 (approved):

PR #5730, opened 2023-11-10
Move JS Engine classes

nielsm5:


nielsm5:
Sonar seems to not like the FrankJ2V8LibraryLoader which is a direct copy of the J2V8LibraryLoader with some minor bugfixes.

tnleeuw (approved):

PR #5753, opened 2023-11-14
Fixes for the MultipartUtils headers parser

tnleeuw:


jkosternl (approved):

core/src/main/java/nl/nn/adapterframework/http/mime/MultipartUtils.java
```diff
@@ -42,7 +41,7 @@ public static boolean isMultipart(HttpServletRequest request) {
 	public static String getFieldName(BodyPart part) {
 		try {
 			String[] id = part.getHeader("Content-ID"); //MTOM requests
-			if(id != null) {
+			if(id != null && !StringUtils.isBlank(id[0])) {
```
nielsm5, line 44: Why not use isNotBlank ?

tnleeuw: I have no idea why I didn't use that.

jkosternl (approved):

Issue #5752, opened 2023-11-14
Index out of bounds when downloading image


dirkjanp:
**Describe the issue**
When downloading certain images through an apilistener adapter an StringIndexOutOfBoundsException is returned.
It concerns .png and/or .jpg file types. But there are also .png files handled succesfully.
The ibis adapter itself returns a 200 but the StringIndexOutOfBoundsException is found in the ibis.logs.

From the ladybug I took the base64 representation of the file response. Put it through https://base64.guru/converter/decode/file and the you can download the file properly.

**This is the base64 string of the file:**
[...]

PR #5757, opened 2023-11-15
Fix for StringIndexOutOfBoundsException (#5752)

tnleeuw:
Fix issue #5752 for version 7.8 (against which this was reported as production issue).

Forward-ports for 7.9 and master will come once the fix is merged.

### DIFF
--- a/core/src/test/java/nl/nn/adapterframework/receivers/ReceiverTest.java
+++ b/core/src/test/java/nl/nn/adapterframework/receivers/ReceiverTest.java
@@ -45,6 +45,7 @@ import nl.nn.adapterframework.jdbc.MessageStoreListener;
 import nl.nn.adapterframework.jta.narayana.NarayanaJtaTransactionManager;
 import nl.nn.adapterframework.pipes.EchoPipe;
 import nl.nn.adapterframework.stream.Message;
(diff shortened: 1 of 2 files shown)


nielsm5 (approved):

core/src/test/java/nl/nn/adapterframework/receivers/ReceiverTest.java
```diff
@@ -563,4 +564,42 @@ public void startReceiver() throws Exception {
 		assertEquals(RunState.STOPPED, receiver.getRunState());
 		assertEquals(RunState.STOPPED, adapter.getRunState());
 	}
+
+	@Test
+	public void processLargeResult() throws Exception {
```
jkosternl, line 569: I'd suggest a name which reflects the special situation you're testing. Such as: `processTooLargeMessageSize`, which is a bit exaggerating, but ok.

jkosternl (approved):

core/src/test/java/nl/nn/adapterframework/receivers/ReceiverTest.java
tnleeuw: `testResultLargerThanMaxCommentSize`?

jkosternl (approved):

PR #5782, opened 2023-11-16
Update GitHub actions/checkout to v4

jkosternl:


tnleeuw:
Two points I want to raise before approving:
1. None of the workflows that have been updated, has actually been triggered. So while nothing _should_ break I'm still a bit hesitant to approve.
2. Once approved this change should be backported to all branches we are still maintaining. (I'm not entirely certain if that should include 7.7-release or not at this point b/c I'm not sure what the contracts say).

jkosternl:
> Two points I want to raise before approving:

Thanks @tnleeuw, I didn't know about that. I used the checkout action on the Frank-Doc project already. At least, I thought it is used to checkout the project and IAF. 
I will backport it too, when needed. Wasn't aware of that.

tnleeuw (approved):

tnleeuw:
> > Two points I want to raise before approving:
> 
> Thanks @tnleeuw, I didn't know about that. I used the checkout action on the Frank-Doc project already. At least, I thought it is used to checkout the project and IAF. I will backport it too, when needed. Wasn't aware of that.

Thanks, knowing you already use this version in the Frank!Doc project without need to change anything else makes me feel more confident to approve it here!

nielsm5 (approved):

nielsm5:
> Two points I want to raise before approving:
> 
> 1. None of the workflows that have been updated, has actually been triggered. So while nothing _should_ break I'm still a bit hesitant to approve.
> 2. Once approved this change should be backported to all branches we are still maintaining. (I'm not entirely certain if that should include 7.7-release or not at this point b/c I'm not sure what the contracts say).

I doubt it will have a significant impact. So I don't think a backport to 7.7/7.8 is required. The only change so far seems to be [partial checkout filters](https://github.com/actions/checkout/pull/1396).

PR #5788, opened 2023-11-16
Chore/compare enums with equality operator

jkosternl:
Improving code quality with OpenRewrite:
https://docs.openrewrite.org/recipes/staticanalysis/booleanchecksnotinverted -> 1 change
https://docs.openrewrite.org/recipes/staticanalysis/compareenumswithequalityoperator -> all the rest

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/scheduler/JobDefFunctions.java
+++ b/core/src/main/java/nl/nn/adapterframework/scheduler/JobDefFunctions.java
@@ -60,7 +60,7 @@ public enum JobDefFunctions implements DocumentedEnum {
 	public boolean isEqualToAtLeastOneOf(JobDefFunctions... functions) {
 		boolean equals = false;
 		for (JobDefFunctions func : functions) {
-			if(super.equals(func))
+			if(super == func)
 				equals = true;
 		}
 		return equals;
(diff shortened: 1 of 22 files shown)


core/src/main/java/nl/nn/adapterframework/scheduler/JobDefFunctions.java
```diff
@@ -60,7 +60,7 @@ public boolean isNotEqualToAtLeastOneOf(JobDefFunctions... functions) {
 	public boolean isEqualToAtLeastOneOf(JobDefFunctions... functions) {
 		boolean equals = false;
 		for (JobDefFunctions func : functions) {
-			if(super.equals(func))
+			if(super == func)
```
tnleeuw, line 63: Using `super` on its own is not legal indeed. `super` should be replaced with `this`.

And the function can also be optimised to `return true` instead of assigning `equals = true`

tnleeuw (changes requested):

core/src/main/java/nl/nn/adapterframework/scheduler/JobDefFunctions.java
jkosternl: Yes, thanks @tnleeuw. Just fixed it and improved the code around it.

tnleeuw (approved):

nielsm5 (approved):

PR #5797, opened 2023-11-17
Cleanup web.xml

nielsm5:
Step one, start the framework via servlet SPI (ServletContainerInitializer).


bundle-minimal/pom.xml
```diff
@@ -53,4 +53,18 @@
...
+	<build>
+		<pluginManagement>
+			<plugins>
+				<plugin>
+					<groupId>org.apache.maven.plugins</groupId>
+					<artifactId>maven-war-plugin</artifactId>
+					<configuration>
+						<failOnMissingWebXml>false</failOnMissingWebXml>
+					</configuration>
+				</plugin>
```
jkosternl, line 66: Is this just to configure the property or to create the war output file? Because this minimal bundle exists to run as a monolithic application, I didn't expect this here.

jkosternl (approved):

bundle-minimal/pom.xml
nielsm5: It's in pluginManagement, thus to set this as default for other projects that inherit from the bundle-minimal pom.

jkosternl: Ah yes, I missed the 'pluginManagement' tag. Clear.

tnleeuw (approved):

PR #5819, opened 2023-11-21
Add security-constraint to web-fragment

nielsm5:


security/src/main/java/nl/nn/adapterframework/lifecycle/servlets/ServletConfiguration.java
```diff
@@ -53,7 +54,7 @@ public class ServletConfiguration implements InitializingBean, EnvironmentAware
 
 	private static final char SLASH = '/';
 	private @Getter String name;
-	private @Getter List<String> securityRoles = Collections.emptyList();
+	private List<String> securityRoles = Collections.emptyList();
```
jkosternl, line 57: This list is immutable.. does that work? I expect security roles might be added later.

jkosternl: Oh, never mind.. it is overwritten by `setSecurityRoles`

jkosternl (approved):

tnleeuw (approved):

PR #5828, opened 2023-11-21
Refactor custom-view component to more TypeScript compliant solution

Laurens-makel:
- double if not required, simplified to one
- use const when not reassigning variables
- object spread: `{name}` is the same as `{name: name}` in typescript
- prefer to define object based types as a generic functional re-usable type
- prefer to use single quotes or use the ` character when embedding variables in strings

I'm trying to understand why this logic has been placed in a subscription in `ngOnInit`, whilst the appConstants value being used in this block of logic is set in the constructor and not taken from the observable emission?

Could the subscription be removed and the logic placed directly in the constructor, since this is the only assignment of `this.appConstants`..?


Matthbo:
Im going to close this pr as this branch is outdated, the new frontend has already been merged into 7.9 and master, and changes should be discussed according to #5934

PR #5829, opened 2023-11-21
Refactor pages-topnavbar.component.ts to more TypeScript compliant solution

Laurens-makel:
not required to explicitly define the type for a variable when its clear from its initialization value what type is.


Matthbo:
Im going to close this pr as this branch is outdated, the new frontend has already been merged into 7.9 and master, and changes should be discussed according to #5934

PR #5885, opened 2023-11-29
Remove file name utils pt2

tnleeuw:
I had this still in an open branch locally...


tnleeuw:
I had this open locally but it seems that in the meantime, from master this file had already been deleted. So this PR brings no changes actually but I didn't realise that until after creating it!

Issue #5466, opened 2023-09-27
XsltPipe resolves escaped line endings after processing an xml


alisihab:
I have an XsltPipe which tries to replace the conversationId in the soap header. After processing an xml that has `&#xd;` chars in the body with an XsltPipe, in the output:
 - Ladybug shows that the chars are resolved line feed
 - Test a pipeline shows everything in one line without the char and lf

When I do the same transformation with the same xml and xsl in XMLSpy I see that the body content is not changed. I'd like to see the same behavior in the framework.

**Reporter**
Ali

Pipe:
`<pipe className="nl.nn.adapterframework.pipes.XsltPipe" name="InjectId" styleSheetName="InjectId.xsl">
	<param name="customId" value="asd"/>
</pipe>`

[InjectConvId.xsl.txt](https://github.com/ibissource/iaf/files/12735322/InjectConvId.xsl.txt)
[input.xml.txt](https://github.com/ibissource/iaf/files/12735323/input.xml.txt)

Test a pipeline screen:
![image](https://github.com/ibissource/iaf/assets/10929856/147fdcc5-c9ba-4abc-a1cd-fde4e9ea1d58)

Ladybug screen:
![image](https://github.com/ibissource/iaf/assets/10929856/8f1e725d-f59a-4125-986b-520d60b18a97)

When I write the output of XsltPipe to a file:
[asd.txt](https://github.com/ibissource/iaf/files/12735370/asd.txt)

FF: 7.9-20230811.190015

jkosternl:
@alisihab, the current implementation of the XsltPipe always seems to work like that. Meaning that input `asd&#xd;` turns into 'asd' and a new line added. 
See `nl.nn.adapterframework.xslt.XsltPipeTest#testSessionKey` since 2019 or 2020. 
I'll discuss if we need to change the displaying of the results in some screens in the UI now.

PR #5895, opened 2023-11-30
Make Test a Pipeline show correct line feeds in the output (#5466)

jkosternl:


tnleeuw:
Have you tested that this resolves the display-issue? I'd like to see some visual evidence of that, with and without the change, if possible.
If not then perhaps an explanation of why this fixes the issue?

Otherwise I find it hard from the code-change to verify the fix is correct.

tnleeuw (approved):

nielsm5 (approved):

PR #5896, opened 2023-11-30
Remove @SupportsOutputStreaming annotation and usage

jkosternl:
Forgot to cleanup while removing the StreamingPipe etc stuff. Now, it is not used anymore. Hence, it can be removed.


tnleeuw (approved):
Is there also a PR for the frank-doc plugin to remove the same tag there?

nielsm5 (approved):

PR #5907, opened 2023-12-01
Upgrade activation to 1.2.2

nielsm5:


core/pom.xml
```diff
@@ -101,10 +101,9 @@
 			<version>${micrometer.version}</version>
 		</dependency>
 
-		<!-- explicit dependency to force version 1.1.1 -->
-		<dependency><!-- this dependency handles javax.activation.* imports, compatible with the IBM JRE -->
-			<groupId>javax.activation</groupId>
-			<artifactId>activation</artifactId>
+		<dependency><!-- explicit dependency to force version 1.2.2 which handles javax.activation.* imports, compatible with the IBM JRE -->
```
tnleeuw, line 104: For master, why do we still need to mention compatibility with IBM JRE?

nielsm5: Yea, I was meant to change this in the 7.9 branch.. After making the pr I thought why not also in here 😞

tnleeuw (approved):

PR #5911, opened 2023-12-01
Upgrade activation to 1.2.2

nielsm5:
Originally it was thought that using version 1.1.1 would solve issues, however this was not the case...
See #3408 for the original dependency lock


tnleeuw (approved):

jkosternl (approved):
Good finding to use that version, which works in both situations.

PR #6021, opened 2023-12-20
Add ObfuscatePipe to tibco module

alisihab:


jkosternl:
Hi @alisihab, did you also add this PR to the master already? Usually we change master first and after approval, we create a patch for versions < master.

alisihab:
I was asked to share the file only

PR #6062, opened 2023-12-29
Implement method HEAD in ApiListener

Yale96:


jkosternl:
@Yale96, can you check (or perhaps add a check) that no real content data is received if the HEAD method is used? Maybe we need to add something to prevent that (or that it is used to by pass security etc).

tnleeuw:
> @Yale96, can you check (or perhaps add a check) that no real content data is received if the HEAD method is used? Maybe we need to add something to prevent that (or that it is used to by pass security etc).

Indeed please add a test for this. I see nothing in code now to prevent data from being returned.

core/src/main/java/org/frankframework/http/rest/ApiListener.java
```diff
@@ -169,6 +169,11 @@ public void close() {
 
 	@Override
 	public Message processRequest(Message message, PipeLineSession session) throws ListenerException {
+		if (getMethod() == HttpMethod.HEAD) {
+			// If it's a HEAD request, return an empty message
+			return new Message("");
```
jkosternl, line 174: I guess it should be returning no data, instead of an empty message

jkosternl (changes requested):

tnleeuw (approved):

jkosternl (approved):

jkosternl (changes requested):
As discussed in daily.

core/src/test/java/org/frankframework/http/rest/ApiListenerServletTest.java
```diff
@@ -382,6 +398,20 @@ public void apiListenerThatProducesJSON() throws ServletException, IOException,
 		assertNull(result.getErrorMessage());
 	}
 
+	@Test
+	public void apiListenerThatProducesJSONForMethodHead() throws ServletException, IOException, ListenerException, ConfigurationException {
+		String uri = "/ApiListenerThatProducesJSON/";
+		new ApiListenerBuilder(uri, Methods.HEAD, null, MediaTypes.JSON).build();
+
+		Response result = service(createRequest(uri, Methods.HEAD, "{}"));
```
jkosternl, line 406: Does the 3rd parameter (content) should be null or an empty String?

core/src/test/java/org/frankframework/http/rest/ApiListenerTest.java
```diff
@@ -258,4 +261,10 @@ public void testGetPhysicalDestinationNameWith2Endpoints() throws Exception {
 		listener.configure();
 		assertEquals("uriPattern: [/aap, /noot]/dummy; method: PUT; consumes: JSON; produces: XML", listener.getPhysicalDestinationName());
 	}
+
+	@Test
+	void testGetNoMessageWithHeadMethod() throws ListenerException {
```
jkosternl, line 266:
```suggestion
	void testGetEmptyBodyMessageWithHeadMethod() throws ListenerException {
```

jkosternl (approved):
Looking good. Just a few minor remarks to consider.

core/src/test/java/org/frankframework/http/rest/ApiListenerServletTest.java
Yale96: I do think an empty body, providing null gives both warnings/errors and looks a bit odd to me.

jkosternl: Ah yes, this is input, instead of output.. it is fine 😉

jkosternl (approved):

core/src/test/java/org/frankframework/http/rest/ApiListenerServletTest.java
```diff
@@ -220,6 +220,10 @@ private MockHttpServletRequest doCreateRequest(String uriPattern, Methods method
 			request.setContent(content);
 		}
 
+		if (method == Methods.HEAD) {
+			request.setContent(null);
+		}
+
```
tnleeuw, line 226: This is a strange bit of code. We have first an `if` to check that method != GET then we set content. Then we check method == HEAD and then oops we remove the content again.

That can be combined.

core/src/test/java/org/frankframework/http/rest/ApiListenerServletTest.java
tnleeuw: It's the request content, which is not even set when method = HEAD. Right?
So you should pass NULL I think.

I wonder why that would give errors?

core/src/test/java/org/frankframework/http/rest/ApiListenerTest.java
```diff
@@ -258,4 +261,10 @@ public void testGetPhysicalDestinationNameWith2Endpoints() throws Exception {
 		listener.configure();
 		assertEquals("uriPattern: [/aap, /noot]/dummy; method: PUT; consumes: JSON; produces: XML", listener.getPhysicalDestinationName());
 	}
+
+	@Test
+	void testGetEmptyBodyMessageWithHeadMethod() throws ListenerException {
+		listener.setMethod(ApiListener.HttpMethod.HEAD);
+		assertTrue(listener.processRequest(new Message("Mocked request"), new PipeLineSession()).toString().contains("[]"));
```
tnleeuw, line 268: What does this really test?

And should the ApiListener really return an empty message always? Why?
Important is that the ApiListenerServlet doesn't return data.

core/src/main/java/org/frankframework/http/rest/ApiListenerServlet.java
```diff
@@ -524,6 +522,11 @@ else if(segment.startsWith("{") && segment.endsWith("}")) {
 				 */
 				Message result = listener.processRequest(body, messageContext);
 
+				if (listener.getMethod().equals(ApiListener.HttpMethod.HEAD)) {
+					response.setContentType(listener.getContentType().toString());
+					result.isNull();
```
tnleeuw, line 527: What does `result.isNull()` here do?

core/src/main/java/org/frankframework/http/rest/ApiListenerServlet.java
```diff
@@ -524,6 +522,11 @@ else if(segment.startsWith("{") && segment.endsWith("}")) {
 				 */
 				Message result = listener.processRequest(body, messageContext);
 
+				if (listener.getMethod().equals(ApiListener.HttpMethod.HEAD)) {
```
tnleeuw, line 525: Laurens has a PR to allow the ApiListener to handle multiple HTTP methods. Then this will no longer work.
To be better prepared for that better to check the method of the actual HTTP Request itself, not the Listener configuration.

But in addition I'm also not sure that this is the right thing to do. We have code to set content-type further down. I would keep all code for setting content-type together.

core/src/main/java/org/frankframework/http/rest/ApiListener.java
tnleeuw: I think it is wrong that we return an empty message instead of NULL, because below we say that we explicitly want NULL returned instead of empty messages.

In addition, I think it is wrong that we skip execution here altogether. I mean, if we do not execute code in this method, which is called from ApiListenerServlet, why even invoke it altogether?

I would rather have the listener executed normally so the adapter / pipeline has a change to process the HEAD request in a way it sees fit. For instance, to correctly determine the actual content-type header for the response (which could be different from the default configured on the ApiListener).

@jjansenvr @alisihab What do you think of this? Should HTTP method HEAD invoke the pipeline like a normal call in order to potentially compute content-type (and perhaps content-length?) header and let the pipeline itself figure out what to do with method HEAD?

Or should pipeline execution be skipped, and default content-type be returned?

(I'm thinking of the case of a pipeline that downloads something like an image. The content-type will be the particular image-type: JPEG, PNG, etc. So content-type depends on the result).

tnleeuw (changes requested):
In addition to the comments on the PR also on all affected files change the copyright-year to include the year 2024! :-)

core/src/test/java/org/frankframework/http/rest/ApiListenerTest.java
Yale96: What I wanted to test here is that we return an empty body when we call the HEAD method.
Just an empty body, no null values or so

core/src/main/java/org/frankframework/http/rest/ApiListenerServlet.java
Yale96: This is indeed wrong, do not really know why I did this, but I will change it

core/src/test/java/org/frankframework/http/rest/ApiListenerTest.java
tnleeuw: I'm not sure how this test actually accomplishes that, or if it's the right way to accomplish that.

core/src/main/java/org/frankframework/http/rest/ApiListener.java
tnleeuw: I discussed my questions on this with Jeroen and we concluded that the full pipeline should be executed as normal. So we should not skip execution and just return an empty message.

The ApiListenerServlet should then apply all headers including content-length and content-type as normal for any request, but should not return the actual body.

Jeroen has updated the ticket with the functional requirements regarding this.

core/src/main/java/org/frankframework/http/rest/ApiListenerServlet.java
```diff
@@ -522,7 +520,9 @@ else if(segment.startsWith("{") && segment.endsWith("}")) {
 				/*
 				 * Do the actual request processing by the ApiListener
 				 */
-				Message result = listener.processRequest(body, messageContext);
+				//@SuppressWarnings("java:S2583") // Suppress Sonar warning for setting result to null
+				Message result = listener.getMethod().equals(ApiListener.HttpMethod.HEAD) ? null : listener.processRequest(body, messageContext);
+
```
tnleeuw, line 525: No longer needed.

tnleeuw (changes requested):
I discussed my functional questions on this PR with Jeroen and we concluded that the full pipeline should be executed as normal. So we should not skip execution and just return an empty message.

The ApiListenerServlet should then apply all headers including content-length and content-type as normal for any request, but should not return the actual body.

Jeroen has updated the ticket with the functional requirements regarding this.

tnleeuw (changes requested):

core/src/test/java/org/frankframework/http/rest/ApiListenerServletTest.java
```diff
@@ -1198,8 +1225,11 @@ public void testRequestWithAccept(Methods method) throws Exception {
 
 		// Assert
 		assertEquals(200, result.getStatus());
-		Message input = requestMessage;
-		assertEquals("application/xml", input.getContext().get("Header.accept"));
+		if (method != Methods.HEAD) {
```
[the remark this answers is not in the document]
Yale96: That's not needed anymore indeed. I will remove it

core/src/test/java/org/frankframework/http/rest/ApiListenerServletTest.java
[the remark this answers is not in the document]
Yale96: I have added a test now for the HEAD method

core/src/main/java/org/frankframework/http/rest/ApiListener.java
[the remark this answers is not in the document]
Yale96: Is this still needed with the new changes provided?

core/src/main/java/org/frankframework/http/rest/ApiListenerServlet.java
```diff
@@ -579,9 +579,11 @@ else if(segment.startsWith("{") && segment.endsWith("}")) {
 				 */
 				response.addHeader("Allow", (String) messageContext.get("allowedMethods"));
 
-				if (!Message.isEmpty(result)) {
+				if (!Message.isEmpty(result) && result.size() != -1) {
```
tnleeuw, line 582: This is not the correct change because you now make setting content-type dependent on result-size being known.
What I had suggested earlier was to combine `(!Message.isEmpty(result) || method == ApiListener.HttpMethod.HEAD)`

Checking for size can be done in a nested `if`. And when setting content-length on the response, you should not need to also add a header `content-length` because the HttpResponse implementation should do this for you.

core/src/main/java/org/frankframework/http/rest/ApiListenerServlet.java
```diff
@@ -600,16 +602,20 @@ else if(segment.startsWith("{") && segment.endsWith("}")) {
...
+					 */
+					final boolean outputWritten = writeToResponseStream(response, result);
+					if (!outputWritten) {
+						LOG.debug("No output written, set content-type header to null");
+						response.resetBuffer();
+						response.setContentType(null);
+					}
+				} else {
+					response.setContentLength(0);
+					response.addHeader("content-length", "0");
```
tnleeuw, line 617: Now you always set content-length to 0 when method = HEAD?
I thought that that goes against the intention?

core/src/test/java/org/frankframework/http/rest/ApiListenerServletTest.java
```diff
@@ -913,6 +926,33 @@ public void apiListenerWithRepeatableMessageAndGloballyDisabled() throws Excepti
...
+
+		Map<String, String> headers = new HashMap<>();
+		headers.put("Accept", "application/json");
+		headers.put("content-type", "application/json");
+
+		// Act
+		Response result = service(createRequest(uri, Methods.HEAD, null, headers));
+
+		// Assert
+		assertEquals(200, result.getStatus());
```
tnleeuw, line 946: I do not see a check here that the content of the response is empty. That check is perhaps the most important check to validate if HEAD did what it is supposed to do.

core/src/test/java/org/frankframework/http/rest/ApiListenerServletTest.java
```diff
@@ -913,6 +926,33 @@ public void apiListenerWithRepeatableMessageAndGloballyDisabled() throws Excepti
...
+		assertEquals("OPTIONS, HEAD", result.getHeader("Allow"));
+		assertNull(result.getErrorMessage());
+		assertFalse(result.containsHeader("etag"));
+		assertEquals("0", result.getHeader("content-length"));
```
tnleeuw, line 950: The length of the response should not have been 0 as the response-content that would have been returned by GET would have been `{"tralalalallala"}`.

core/src/test/java/org/frankframework/http/rest/ApiListenerServletTest.java
```diff
@@ -913,6 +926,33 @@ public void apiListenerWithRepeatableMessageAndGloballyDisabled() throws Excepti
...
+		assertEquals("*/*", result.getHeader("content-type"));
```
tnleeuw, line 951: I'm surprised that the content-type header is `*/*` and not JSON. Can you please set the mimetype on the `repeatableMessage` to JSON? Or otherwise the `produces` on the ApiListenerBuilder?

core/src/test/java/org/frankframework/http/rest/ApiListenerServletTest.java
```diff
@@ -913,6 +926,33 @@ public void apiListenerWithRepeatableMessageAndGloballyDisabled() throws Excepti
 		assertTrue(result.containsHeader("pragma"));
 	}
 
+	@Test
+	public void apiListenerWithHeadMethodCall() throws Exception {
+		// Arrange
+		String uri = "/apiListenerWithHeadMethodCall";
+		Message repeatableMessage = Message.asMessage(new Message("{\"tralalalallala\":true}").asByteArray());
```
tnleeuw, line 933: After applying other comments to this test, I would like a copy of this test where this `repeatableMessage` is constructed with an empty string, but the size set to something like 20 by setting the corresponding key in the MessageContext. Then the content-length returned by HEAD should also be 20.

tnleeuw (changes requested):

tnleeuw (approved):

core/src/main/java/org/frankframework/http/rest/ApiListenerServlet.java
```diff
@@ -600,16 +603,17 @@ else if(segment.startsWith("{") && segment.endsWith("}")) {
...
-				final boolean outputWritten = writeToResponseStream(response, result);
-				if (!outputWritten) {
-					LOG.debug("No output written, set content-type header to null");
-					response.resetBuffer();
-					response.setContentType(null);
+				if (method != ApiListener.HttpMethod.HEAD) {
+					/*
+					 * Finalize the pipeline and write the result to the response
+					 */
+					final boolean outputWritten = writeToResponseStream(response, result);
```
nielsm5, line 610: Sluiten we deze wel?

nielsm5 (commented):
I've made some changes in #6148 so one of us needs to merge those :)

core/src/main/java/org/frankframework/http/rest/ApiListener.java
```diff
@@ -240,6 +226,7 @@ public boolean accepts(@Nullable String acceptHeader) {
 	 * HTTP method to listen to
 	 * @ff.default GET
 	 */
+	@Deprecated
```
nielsm5, line 229:
```suggestion
```

nielsm5 (approved):

tnleeuw (approved):

PR #6068, opened 2024-01-03
Support Ladybug database storage

jacodg:


ladybug/pom.xml
```diff
@@ -11,7 +11,7 @@
 	<name>Frank!Framework Ladybug</name>
 
 	<properties>
-		<ladybug.version>2.3-20230911.183804</ladybug.version>
+		<ladybug.version>2.3-20240103.172229</ladybug.version>
```
jkosternl, line 14: Does this version bump also fixes vulnerabilities of the current (older) Ladybug version?

jkosternl (approved):
Looks fine. Where can I find exactly what is stored in the database by Ladybug or how to use it Jaco? I just miss some background knowledge, but perhaps that's already documented somewhere?

ladybug/pom.xml
jacodg: Yes, it is the latest version which includes:

https://github.com/wearefrank/ladybug/blob/master/RELEASES.md

We still need to improve/automate the release mechanism for Ladybug.

jacodg:
See the database changelog files at:

https://github.com/wearefrank/ladybug/tree/master/src/main/resources/ladybug

Basically the Frank!Developer doesn't have to do anything. When jdbc.migrator.active is true (which is generally already used by Franks that have the rights to make database changes) Ladybug will create it's tables and use the database storage classes. If not, it will fall back to using the file storage classes storing data in log.dir (as it currently does).

ladybug/pom.xml
jkosternl: Ok great!

ladybug/src/main/java/org/frankframework/ibistesttool/DeploymentSpecificsBeanPostProcessor.java
```diff
@@ -92,6 +93,14 @@ public Object postProcessBeforeInitialization(Object bean, String beanName) thro
 				}
 			}
 		}
+		if (bean instanceof SpringLiquibase) {
+			boolean active = APP_CONSTANTS.getBoolean("ladybug.jdbc.migrator.active",
+					APP_CONSTANTS.getBoolean("jdbc.migrator.active", false));
+			if (!active) {
```
tnleeuw, line 99: Waarom niet gewoon altijd de waarde van `active` in de bean zetten?

nielsm5: Als private field (bovenaan in de classe)

nielsm5 (approved):

PR #6070, opened 2024-01-04
Add JUnit tests for JwtSecurityHandler

Laurens-makel:


tnleeuw (changes requested):
I like the improvements in coverage and code quality but I see 1 minor complaint from Codacy. Could you please fix that?

jkosternl:
@Laurens-makel, you've fixed the Codacy issue, but the builds are failing on your unit tests. Can you fix that (and merge master), so this PR can be merged and finished too?

core/src/main/java/org/frankframework/jwt/JwtSecurityHandler.java
```diff
@@ -68,40 +68,52 @@ public Principal getPrincipal() {
...
+				.findFirst();
+
+		if(nonMatchingClaim.isPresent()){
+			String key = nonMatchingClaim.get().getKey();
+			String expectedValue = nonMatchingClaim.get().getValue();
+			throw new AuthorizationException("JWT "+key+" claim has value ["+claimsSet.get(key)+"], must be ["+expectedValue+"]");
+		}
+	}
+
+	public void validateAnyMatchClaims(@Nonnull String anyMatchClaims) throws AuthorizationException {
```
jkosternl, line 107: Looking good. I just wonder if these 3 public methods validate* also can be made private or at least package scoped.

jkosternl (approved):

core/src/main/java/org/frankframework/jwt/JwtSecurityHandler.java
Laurens-makel: Yes, that should be possible but I'm wondering why? 

Is it because these methods are not publicly used currently?

jkosternl: Because it belongs to the internals of this class. Also because it is a good idea to make it as open as needed. So, private if possible, or public if it need to be exposes to other packages explicitly. 
This article describes it even better: https://methodpoet.com/public-methods/

Laurens-makel: > Because it belongs to the internals of this class. Also because it is a good idea to make it as open as needed. So, private if possible, or public if it need to be exposes to other packages explicitly. This article describes it even better: https://methodpoet.com/public-methods/

Thanks for the article, let's make these methods package scoped because it makes testing easier.

tnleeuw (approved):

PR #6073, opened 2024-01-05
Add unit tests & notFoundForwardName attribute to GetPrincipalPipe

Laurens-makel:
Changes:

- New forward added to `GetPrincipalPipe` which should be taken when the Principal object is null or the principal 's name is either an empty string or null.
- Fully backwards compatible, only if Frank developer configures the `notFoundForwardName;` attribute it will be effective.
- 100% test coverage added.


Laurens-makel:
Moet herzien worden na het mergen van #6080

tnleeuw:
> Moet herzien worden na het mergen van #6080

Dat PR is nu gemerged dus je kan hieraan werken wanneer je tijd hebt! 😃

tnleeuw:
PS @Laurens-makel : Als het PR klaar is voor review zou je dan Jacobjob, Yannick & mijzelf als reviewers willen toevoegen? Dan krijgen we een seintje om ernaar te kijken.

core/src/main/java/org/frankframework/pipes/GetPrincipalPipe.java
```diff
@@ -33,13 +38,29 @@
  */
 @ElementType(ElementTypes.SESSION)
 public class GetPrincipalPipe extends FixedForwardPipe {
+	private @Getter String notFoundForwardName;
+	protected PipeForward notFoundForward;
+
+	@Override
+	public void configure() throws ConfigurationException {
+		super.configure();
+		if (StringUtils.isNotEmpty(getNotFoundForwardName())) {
```
tnleeuw, line 47: I often wonder wether it makes more sense to check `isEmpty()` or `isBlank()`.
What do you think is the better check in this case?

tnleeuw (approved):

core/src/main/java/org/frankframework/pipes/GetPrincipalPipe.java
Laurens-makel: As long as I am able to register a forward with a blank name, for example `<Forward name=" " path="doSomething" />`,  I should be able to specify the same value for the `notFoundForwardName` attribute.

Laurens-makel:
@tnleeuw 
Sometimes I review my own work and find some remarks, I've noticed that when `principalName = principal.getName();` returns `null`, this is also returned in the `PipeRunResult`.

I believe, correct me if wrong, this will cause the pipe to output `null` and this behaviour is not really desired? Would it be better to return an empty string?

tnleeuw:
@Laurens-makel Do you want to make more changes to this PR based on your considerations above or shall I merge it as it is now?

core/src/test/java/org/frankframework/pipes/GetPrincipalPipeTest.java
```diff
@@ -0,0 +1,230 @@
...
+		String result=prr.getResult().asString();
+		assertEquals("success", prr.getPipeForward().getName());
+		assertEquals(PRINCIPAL_NAME, result);
+	}
+
+	@Test
+	void getNullPrincipalNameDoPipe() throws Exception {
+		// Given
+		when(principal.getName()).thenReturn(null);
+				when(securityHandler.getPrincipal()).thenReturn(principal);
```
jkosternl, line 103: Please fix your indentation 😉  except if you confess that you're using Notepad.

Laurens-makel:
> @Laurens-makel Do you want to make more changes to this PR based on your considerations above or shall I merge it as it is now?

I will solve the indentation issues first but no more further changes.

core/src/test/java/org/frankframework/pipes/GetPrincipalPipeTest.java
Laurens-makel: I'm a fan of notepad++ for anything except writing code.

Not sure when and how this happened, will solve this.

PR #6077, opened 2024-01-08
Add EscapePipe iterative escaping when using substringStart

Yale96:


core/src/main/java/org/frankframework/pipes/EscapePipe.java
```diff
@@ -72,26 +72,31 @@ public PipeRunResult doPipe(Message message, PipeLineSession session) throws Pip
...
-				}
+		while ((i = input.indexOf(substringStart, i + 1)) != -1) {
+			int j = input.indexOf(substringEnd, i);
+			if (j != -1) {
+				// Pak de content tussen de substrings
+				String content = input.substring(i + substringStart.length(), j);
+
+				String processedContent = handle(content);
+
+				// Replace the geparste content in de originele string
```
jkosternl, line 93: Please don't use Dutch in the code, also for line 88.

core/src/main/java/org/frankframework/pipes/EscapePipe.java
```diff
@@ -72,26 +72,31 @@ public PipeRunResult doPipe(Message message, PipeLineSession session) throws Pip
...
-		if (substringStart != null && substringEnd != null) {
-			i = input.indexOf(substringStart);
-			if (i != -1) {
-				j = input.indexOf(substringEnd, i);
-				if (j != -1) {
-					substring = input.substring(i + substringStart.length(), j);
-					substring = handle(substring);
-					result = input.substring(0, i + substringStart.length()) + substring + input.substring(j);
-				}
+		while ((i = input.indexOf(substringStart, i + 1)) != -1) {
```
jkosternl, line 85: With this new solution, the while-loop is cycling through every character of the input string, even though perhaps there are no hits at all. That is not needed right? 

Perhaps an approach could work that works like this: search for first occurrence -> handle it. Check of second occurrence -> handle it (or if not found, stop). Etc. without looping over every character of the String.

core/src/test/java/org/frankframework/pipes/EscapePipeTest.java
```diff
@@ -51,6 +51,18 @@ public void testPartialEscapedMultiLineXmlInput() throws Exception { //double es
...
+		pipe.setSubstringStart("<message>");
+		pipe.setSubstringEnd("</message>");
+		configureAndStartPipe();
+
+		PipeRunResult result = doPipe(getResource("multi-line-partial-escaped-multi-occurences.xml"));
+		String expected = getResource("multi-line-partial-escaped-multi-occurences.result").asString();
+		System.out.println(result.getResult().asString());
+		TestAssertions.assertEqualsIgnoreCRLF(expected, result.getResult().asString());
+	}
+
```
jkosternl, line 65: Can you also add a test case (or change the current one), to proof it does not change the parts outside the start/end Substrings? At least, that is how I think it should work.

jkosternl (changes requested):

core/src/main/java/org/frankframework/pipes/EscapePipe.java
tnleeuw: This loop is indeed rather sub-optimal.
1. You replace the string while working on it. Potentially multiple times.
2. You continue searching from the last found position of the start-string, instead of last found position of the end-string. (Although by changing the string in-place, that position is now invalid).

Much better to collect new string in a StringBuilder, and also the method `handle(content)` can append to that buffer.

tnleeuw (changes requested):

core/src/test/java/org/frankframework/pipes/EscapePipeTest.java
Yale96: It does do that right?
It compares a new xml file where only the content between the message tags is being replaced

jkosternl: Yes, but perhaps you can also proof by (the same?) a test that outside the start/end substrings it does not replace chars.

core/src/main/java/org/frankframework/pipes/EscapePipe.java
```diff
@@ -72,26 +72,34 @@ public PipeRunResult doPipe(Message message, PipeLineSession session) throws Pip
...
+			int j = result.indexOf(substringEnd, i);
+			if (j != -1) {
+				try {
+					// Get the content between the substrings
+					String content = result.substring(i + substringStart.length(), j);
+
+					// Replace the new content in the original string
+					result.replace(i + substringStart.length(), j, handle(content));
+				} catch (NullPointerException nullPointerException) {
+					return substringStart + "null" + substringEnd;
```
jkosternl, line 96: As discussed in Teams: catching a NPE should not be needed, but prevented. Please figure out where in the logic you can make an improvement so it won't happen.

jkosternl (changes requested):

jkosternl (approved):

core/src/main/java/org/frankframework/pipes/EscapePipe.java
```diff
@@ -72,42 +72,52 @@ public PipeRunResult doPipe(Message message, PipeLineSession session) throws Pip
...
+		String result = handleSubstrings(input);
+		return new PipeRunResult(getSuccessForward(), result);
+	}
+
+	private String handleSubstrings(String input) {
+		if (substringStart == null || substringEnd == null) {
+			return handle(input);
+		}
+
+		StringBuilder result = new StringBuilder(input);
```
tnleeuw, line 84: I suggested using a StringBuilder not to do in-place changes in the StringBuilder but to gradually build the final result there, appending each piece as you process it.
The current code is very inefficient, causing a lot of overhead as strings are moved around over and over again in memory.

You can look at the implementation of `XmlEncodingUtils.encodeChars()` for an example of what I mean.

core/src/test/java/org/frankframework/pipes/EscapePipeTest.java
tnleeuw, line 1: I would like to see a test added that has, in the input message, a start-substring, then an end-substring, and then a start-substring without end-substring.

tnleeuw (changes requested):

core/src/main/java/org/frankframework/pipes/EscapePipe.java
jkosternl: Oops, I've missed that part. I didn't check/see how you've handled the prior feedback from Tim. Sorry for that.

jkosternl (changes requested):
Please update me when you've fixed the things Tim said, and I'll review them.

core/src/main/java/org/frankframework/pipes/EscapePipe.java
tnleeuw: Ideally, for best performance, you can also refactor the code in XmlEncodingUtils encodeChars() / decodeChars() and the cdata handling code a bit to add overloaded methods that have a StringBuilder parameter to which they append. That will save a lot of additional copying too.
But if you will find that too confusing I can pick up that final optimisation.

core/src/main/java/org/frankframework/pipes/EscapePipe.java
```diff
@@ -72,42 +72,71 @@ public PipeRunResult doPipe(Message message, PipeLineSession session) throws Pip
...
+				result.append(input, startIndex, input.length());
+				break;
+			}
+
+			// Append the part from the last startIndex to the current substringStart
+			result.append(input, startIndex, start + substringStart.length());
+
+			String content = input.substring(start + substringStart.length(), end);
+
+			String handledContent = handle(content);
```
tnleeuw, line 115: I think this cannot be null unless the input is null?

core/src/test/java/org/frankframework/pipes/EscapePipeTest.java
```diff
@@ -51,6 +51,39 @@ public void testPartialEscapedMultiLineXmlInput() throws Exception { //double es
...
+	@Test
+	public void nothingGetsEscapedWhenNoSubstringEndConfigured() throws Exception {
+		pipe.setSubstringStart("<message>");
+		pipe.setSubstringEnd("");
+		configureAndStartPipe();
+
+		PipeRunResult result = doPipe(getResource("no-substring-configured.xml"));
+		String expected = getResource("no-substring-configured.result").asString();
```
tnleeuw, line 72: Input and output seem identical?

In general in such a situation I think you can just reuse the other input file and assert that input and output are identical, but however I think this test is actually invalid -- see other comment above.

core/src/test/java/org/frankframework/pipes/EscapePipeTest.java
tnleeuw: I'd still like to see a test for this case where the input message is like

```xml
<reasons>
	<message>Validation using WsdlXmlValidator with 'test.wsdl' failed</message>
	<test>Validation using WsdlXmlValidator with 'test.wsdl' failed</test>
	<reason type="ERROR">
		<xpath>/Envelope/Body/path/to/code</xpath>
		<location>at (13,3)</location>
		<message>The element type "ns4:code" must be terminated by the matching end-tag "</ns4:code>".
	</reason>
</reasons>
```

(Not the missing closing tag `</message>` for the 2nd opening-tag).

I also want to see some tests for the `DECODE` direction and for `CDATA2TEXT`, I don't think that I see any tests at all for that?

core/src/main/java/org/frankframework/pipes/EscapePipe.java
```diff
@@ -72,42 +72,71 @@ public PipeRunResult doPipe(Message message, PipeLineSession session) throws Pip
...
+		return new PipeRunResult(getSuccessForward(), result);
+	}
+
+	private String handleSubstrings(String input) {
+		if (substringStart == null || substringEnd == null) {
+			return handle(input);
 		}
 
-		return new PipeRunResult(getSuccessForward(), result);
+		if (substringStart.isEmpty() || substringEnd.isEmpty()) {
```
tnleeuw, line 84: The `configure()` method already checks for one of these to be `null` but not the other.

I think it is not a sensible configuration to have one set but the other set to empty string.

So instead of doing this check here and returning input unchanged, I think it is better to extend the checks done in `configure()`. Instead of checking for `null`, better to check for `StringUtils.isBlank()` both in `configure()` and in the checks now on line 80. And then remove this case that doesn't change input at all which seems wrong.

tnleeuw (changes requested):

core/src/main/java/org/frankframework/pipes/EscapePipe.java
Yale96: It could return null, when the imput XML is not correct this happens

core/src/test/java/org/frankframework/pipes/EscapePipeTest.java
Yale96: I thought this is exactly what we want.
When no substringEnd is configured we don't know between what we would like to escape so we escape nothing

core/src/test/java/org/frankframework/pipes/EscapePipeTest.java
Yale96: We have discussed this in the chat right?
That there where already existing tests for DECODE and CDATA2TEXT

tnleeuw: Good to mention it in the PR as well though. Thanks for getting back on it.

Yale96: I have added the extra test case

jkosternl (approved):

core/src/test/java/org/frankframework/pipes/EscapePipeTest.java
```diff
@@ -51,6 +51,44 @@ public void testPartialEscapedMultiLineXmlInput() throws Exception { //double es
...
+	}
+
+	@Test
+	public void testPartialEscapedMultiLineXmlInputMultiOccurencesMissingClosingTag() throws Exception {
+		pipe.setSubstringStart("<message>");
+		pipe.setSubstringEnd("</message>");
+		configureAndStartPipe();
+
+		PipeRunResult result = doPipe(getResource("no-closing-tag-configured.xml"));
+		String expected = getResource("no-closing-tag-configured.result").asString();
```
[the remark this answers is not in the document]
Yale96: The second <message> opening is missing the </message> closing tag here right?

Yale96:
> I would like to see one more test-case added.
> 
> We now have a test-case for an input looking like "data.... begin-string....xml-tags... end-string.... begin-string.... end of data" but I would also like to have a test for the case:
> 
> "data ... begin-string... begin-string.... xml-tags... end-string... end of data".
> 
> I think we now handle this correctly but we could easily make a mistake, do it incorrectly, and double-escape part of the input so that's why I want this tested.

I've added the extra test case

jkosternl (approved):

core/src/test/java/org/frankframework/pipes/EscapePipeTest.java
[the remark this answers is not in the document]
tnleeuw: The filename is still confusing because closing tag is configured in the pipe, but not supplied in the input.
The filename implies the closing tag is not configured. Configuration applies to the pipe.

tnleeuw (approved):

nielsm5 (approved):

Issue #6032, opened 2023-12-21
(BLOCKING issue in 7.9.0) soap version in wrapper (of sender) is not taken into account when auto is used


jjansenvr:
We have an inbound call `soapVersion=auto`
in our outgoing call we specify `soapVersion=1.1` but it still uses the 1.2 version of the input message. 

**Reporter**
jjansenvr

**To Reproduce**
see email with report 

**Screenshots**

**Configuration**

**Input**

**Environment**
```
7.9.0
```

VBPham:
Hi @jkosternl I tried your suggestion of adding `soapNamespaceSessionKey` to the wrappers but I am still getting this stacktrace when the sender tries to unwrap the response. This is with `7.9-RC1-hotfix10`

````txt
2024-01-10 07:40:42,317 INFO [Thread-145] {cid=Test Tool correlation id(5), adapter=GetAccounts_1} processors.ExceptionHandlingPipeProcessor - exception occured, forwarding to exception-forward [exception], exception:

nl.nn.adapterframework.core.PipeRunException: Pipe [- CallBE_SalesforceLogin_UTL_CreateSession_3: wrap output] Unexpected exception during (un)wrapping : Pipe [- CallBE_SalesforceLogin_UTL_CreateSession_3: wrap output] SOAP Body is empty or message is not a SOAP Message
      at nl.nn.adapterframework.soap.SoapWrapperPipe.doPipe(SoapWrapperPipe.java:252) ~[ibis-adapterframework-core-7.9-RC1-hotfix10.jar:7.9-RC1-hotfix10]
      ... suppressed 65 lines
      at nl.nn.adapterframework.pipes.MessageSendingPipe.postProcessOutput(MessageSendingPipe.java:762) ~[ibis-adapterframework-core-7.9-RC1-hotfix10.jar:7.9-RC1-hotfix10]
      at nl.nn.adapterframework.pipes.MessageSendingPipe.doPipe(MessageSendingPipe.java:513) ~[ibis-adapterframework-core-7.9-RC1-hotfix10.jar:7.9-RC1-hotfix10]
      at nl.nn.adapterframework.pipes.MessageSendingPipe$$FastClassBySpringCGLIB$$b131b2cb.invoke(<generated>) ~[?:?]
      ... suppressed 8 lines
      at nl.nn.adapterframework.pipes.SenderPipe$$EnhancerBySpringCGLIB$$80b1bb02.doPipe(<generated>) ~[?:?]
[...]

PR #6092, opened 2024-01-10
SoapWrapper should try unwrapping with soap version 1.1 too (#6032)

jkosternl:
* Prevents bug of reusing stored SoapVersion from previous SoapWrappers

### DIFF
--- a/core/src/main/java/org/frankframework/soap/SoapWrapper.java
+++ b/core/src/main/java/org/frankframework/soap/SoapWrapper.java
@@ -167,9 +167,11 @@ public class SoapWrapper {
 		String extractedMessage;
 		if (soapVersion == SoapVersion.SOAP12) {
 			extractedMessage = transformerS12.transform(message.asSource());
-			// If session had the wrong SOAP version stored (e.g. multiple SoapWrappers), try SOAP 1.1 too.
+			// If session had the wrong SOAP version stored (e.g. multiple SoapWrappers), try SOAP 1.1 too. (#6032)
 			if (StringUtils.isEmpty(extractedMessage)) {
 				extractedMessage = transformerS11.transform(message.asSource());
+				// TODO: previous SoapWrapper configurations can write the wrong SOAP version to the session (using the same name).
+				// Consider a solution to match the right saved SOAP version with the right SoapWrapper: e.g. cache SoapVersion inside Message.context
 			}
 		} else if (soapVersion == SoapVersion.NONE) {
 			return null;
--- a/core/src/main/java/org/frankframework/soap/SoapWrapper.java
+++ b/core/src/main/java/org/frankframework/soap/SoapWrapper.java
@@ -39,8 +39,13 @@ import org.apache.wss4j.dom.message.WSSecSignature;
 import org.apache.wss4j.dom.message.WSSecTimestamp;
 import org.apache.wss4j.dom.message.WSSecUsernameToken;
 import org.apache.xml.security.algorithms.JCEMapper;
+import org.frankframework.configuration.ConfigurationException;
+import org.frankframework.core.PipeLineSession;
+import org.frankframework.core.SenderException;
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/frankframework/pipes/SoapWrapperPipeTest.java
```diff
@@ -516,13 +504,80 @@ void shouldUseConfiguredPipelineSoapVersionAuto() throws Exception {
...
+		// Assert: but the output is without SOAP namespaces, as it is unwrapped.
+		String result = pipeRunResult3.getResult().asString();
+		assertNotNull(result);
+		assertFalse(result.contains(SoapVersion.SOAP11.namespace), "result should be unwrapped for sure");
+		assertFalse(result.contains(SoapVersion.SOAP12.namespace), "result should be unwrapped for sure");
+		assertTrue(result.contains("FindDocuments_Response"), "result should contain original message, without namespaces");
+		outputWrapper.stop();
+	}
+
+	private P createWrapperSoapPipe(final SoapVersion soapVersion, boolean started) throws ConfigurationException, PipeStartException {
```
Yale96, line 557: I only see started being set to true so is it really needed to add a specific param for this?
What happens if it's not started?

core/src/test/java/org/frankframework/pipes/SoapWrapperPipeTest.java
```diff
@@ -516,13 +504,80 @@ void shouldUseConfiguredPipelineSoapVersionAuto() throws Exception {
...
+		wrapPipeSoap.setSoapVersion(soapVersion);
+		pipeline.addPipe(wrapPipeSoap);
+		if (started) {
+			wrapPipeSoap.configure();
+			wrapPipeSoap.start();
+		}
+		return wrapPipeSoap;
+	}
+
+	private void setupUnwrapSoapPipe(final SoapVersion soapVersion, boolean started) throws ConfigurationException, PipeStartException {
```
Yale96, line 572: The same counts for this

core/src/test/java/org/frankframework/pipes/SoapWrapperPipeTest.java
jkosternl: It is set to false in my new test: `shouldUnwrapEvenThoughSessionHasDifferentSoapVersionStored`
It is postponed in my test, so I can configure more variables/settings and then run the configure&start inside my unit test. Changing settings after a configure is not a good idea (and mostly not taken into account).

core/src/test/java/org/frankframework/pipes/SoapWrapperPipeTest.java
jkosternl: It is set to false in my new test: `shouldUnwrapEvenThoughSessionHasDifferentSoapVersionStored`
It is postponed in my test, so I can configure more variables/settings and then run the configure&start inside my unit test. Changing settings after a configure is not a good idea (and mostly not taken into account).

Yale96 (approved):

core/src/main/java/org/frankframework/soap/SoapWrapper.java
```diff
@@ -168,6 +167,10 @@ private String extractMessageWithTransformers(TransformerPool transformerS11, Tr
 		String extractedMessage;
 		if (soapVersion == SoapVersion.SOAP12) {
 			extractedMessage = transformerS12.transform(message.asSource());
+			// If session had the wrong SOAP version stored (e.g. multiple SoapWrappers), try SOAP 1.1 too.
```
tnleeuw, line 170: This works for now but for future I think we should find a better solution.
Can you please add a TODO comment for that in the code, at least in master?
(we don't need such TODO in the bugfix branches).

tnleeuw (approved):

tnleeuw (approved):

PR #6112, opened 2024-01-15
Add StatusCode to BusException

Yale96:


management-gateway/src/test/java/org/frankframework/management/bus/ResourceNotFoundExceptionTest.java
```diff
@@ -0,0 +1,31 @@
...
+		return Arrays.asList(new Object[][]{
+				{"", Level.WARN, null},
+				{": cannot configure", Level.ERROR, new IbisException("cannot configure")},
+				{": cannot configure: (IllegalStateException) something is wrong", Level.ERROR, new IbisException("cannot configure", new IllegalStateException("something is wrong"))},
+		});
+	}
+
+	@ParameterizedTest
+	@MethodSource("data")
+	public void testExceptionMessage(String expectedLogMessage, Level expectedLogLevel, Exception innerException) {
```
tnleeuw, line 27: Perhaps I'm overlooking something, but where do you use the `expectedLogLevel`?

management-gateway/src/test/java/org/frankframework/management/bus/ResourceNotFoundExceptionTest.java
```diff
@@ -0,0 +1,31 @@
...
+import org.apache.logging.log4j.Level;
+import org.frankframework.core.IbisException;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+public class ResourceNotFoundExceptionTest {
+
+	private static final String RESOURCE_NOT_FOUND_EXCEPTION_MESSAGE = "outer exception";
+
+	public static List<?> data() {
```
tnleeuw, line 17: I know you copied this from another test in the same package, but it is better if this method returns a `Stream<Argument>` instead.
See https://github.com/frankframework/frankframework/blob/2844695d4a4e97514dde6e618f30b3198355c8bc/cmis/src/test/java/org/frankframework/extensions/cmis/TestCreateAction.java#L42-L50 for how to write the method to build that stream.

core/src/main/java/org/frankframework/management/bus/endpoints/BusEndpointBase.java
```diff
@@ -102,7 +102,7 @@ protected Configuration getConfigurationByName(String configurationName) {
 		}
 		Configuration configuration = getIbisManager().getConfiguration(configurationName);
 		if(configuration == null) {
```
tnleeuw, line 104: Please make changes similar to this in all the `getXxxByName()` methods in this class, not just for configuration.

tnleeuw (changes requested):
In addition to the comments made, is it possible to add a unit test covering the changes in the exception handling? That on not finding the configuration/adapter/etc, a status 404 is actually returned?
Current test only seems to test the constructor of the new exception.

In addition I had expected there to have been a change in the class ErrorMessageConverter, to actually return a status 404 to the front-end instead of a 400?

management-gateway/src/main/java/org/frankframework/management/bus/ResourceNotFoundException.java
```diff
@@ -0,0 +1,46 @@
...
+   limitations under the License.
+*/
+package org.frankframework.management.bus;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.frankframework.core.IbisException;
+
+public class ResourceNotFoundException extends RuntimeException {
+	private static final Logger LOG = LogManager.getLogger(ResourceNotFoundException.class);
```
jkosternl, line 23: Feel free to use the `@Log4J` annotation at the class level, to prevent this manual code.

management-gateway/src/main/java/org/frankframework/management/bus/ResourceNotFoundException.java
```diff
@@ -0,0 +1,46 @@
...
+
+	private static final long serialVersionUID = 1L;
+
+	/**
+	 * Seen as WARNING
```
jkosternl, line 28: I don't understand this Javadoc. Not sure if it is helping then. It's just the constructor without an exception.

management-gateway/src/test/java/org/frankframework/management/bus/ResourceNotFoundExceptionTest.java
Yale96: It is somehow needed. I indeed used another class as an example but when removing the parameter, it gives a ParameterResolutionException

tnleeuw: That is because the `data` method, where your test-data comes from, gives some test-data including this value.

You need to remove it from that `data` method in your test-class as well.

core/src/main/java/org/frankframework/management/bus/endpoints/BusEndpointBase.java
jkosternl: It looks like Yannick also applied this comment.

management-gateway/src/main/java/org/frankframework/management/bus/ResourceNotFoundException.java
```diff
@@ -0,0 +1,43 @@
...
+	/**
+	 * Seen as ERROR
+	 * Stacktrace information is logged but not passed to the parent to limit sensitive information being sent over the 'bus'.
+	 */
+	public ResourceNotFoundException(String message, Throwable exception) {
+		super(new IbisException(message, exception).getMessage());
+		if (exception == null) {
+			log.warn(super.getMessage()); // expanded message is logged directly
+		} else {
+			log.error(message, exception); // normal message, expanded by printing the stacktrace
```
jkosternl, line 40: Can you revisit the Javadoc and comments in this class? 
1. L32: What is 'Seen as error'? 
1. L38: Javadoc can be removed. `super.` can be removed too.
1. L40: Javadoc can be removed, I think.

core/src/test/java/org/frankframework/management/bus/BusTestEndpoints.java
```diff
@@ -46,6 +46,8 @@ public Message<String> throwException(Message<?> message) {
 			new ConfigurationException("cannot configure",
 				new IllegalStateException("something is wrong")));
 		switch (type) {
+			case NOT_FOUND:
+				throw new ResourceNotFoundException("Resource not found");
```
jkosternl, line 50: Looks like the formatting is different then the other lines around it.

jkosternl (approved):
Added only small remarks. Looking fine for the rest 👍🏻

core/src/test/java/org/frankframework/management/bus/BusTestEndpoints.java
Yale96: Yeah indeed, my IntelliJ seems to do this automatically so I will look into this

tnleeuw (approved):

jkosternl (approved):

management-gateway/src/main/java/org/frankframework/management/bus/BusException.java
```diff
@@ -19,27 +19,43 @@
...
 import org.frankframework.core.IbisException;
 
+import lombok.Getter;
+
 /**
  * Serialized and send as an ExceptionMessage over the Spring Bus
  */
 public class BusException extends RuntimeException {
-	private static final Logger LOG = LogManager.getLogger(BusException.class);
+	private static final transient Logger LOG = LogManager.getLogger(BusException.class);
```
tnleeuw, line 28: `static` is never serialized so does not need to be `transient` as well, to my knowledge.

(See also https://stackoverflow.com/questions/35083970/what-is-the-effect-of-static-final-transient-in-java and https://codezup.com/transient-keyword-in-java-static-vs-transient-vs-final/ )

tnleeuw (changes requested):
Please remove `transient` from `static final` variable, otherwise good to merge to me!

tnleeuw (approved):

PR #6119, opened 2024-01-16
Add Connection Pool Validation

tnleeuw:
Add some configuration properties for connection pool max-idle connections, and connection validation.


jkosternl:
I am not sure if adding a `maxIdle` connections is the solution. The other existing variable `maxIdleTime` should already be sufficient for a good working mechanism. Because if e.g. 5 connections are idle for 2 seconds, I don't want them to be disconnected and removed directly. The timer should start counting and discard them after e.g. 30 or 60 seconds. But since we are seeing a few issues reported on this topic, I wonder if the maxIdleTime is working überhaupt.

tnleeuw:
> I am not sure if adding a `maxIdle` connections is the solution. The other existing variable `maxIdleTime` should already be sufficient for a good working mechanism. Because if e.g. 5 connections are idle for 2 seconds, I don't want them to be disconnected and removed directly. The timer should start counting and discard them after e.g. 30 or 60 seconds. But since we are seeing a few issues reported on this topic, I wonder if the maxIdleTime is working überhaupt.

maxIdleTime is not available in all pool implementations. It seems to be only available when using BTM's own pool. For non-XA connections we always use DBCP2's pool, and for Narayana we also always use DBCP2's pool implementation (for either XA or non-XA connections).

So it seemed to make sense to me, to add a setting for `maxIdle` as well. But perhaps we can tweak the default.

BTW I made this PR when on a call with Jeroen & Ali to discuss an issue they are having with database connections in one of the systems. I'm not sure if this PR will even help them.

When you are back, we might decide you finish the changes since you are already working on similar sets of changes.

jkosternl:
> When you are back, we might decide you finish the changes since you are already working on similar sets of changes.

Yes, I wasn't aware of the fact that only BTM understands that `maxIdleTime`. I think adding a `maxIdle` make sense. I would not put the setting at 1 because then it directly closes every open idle connection, which basically disables the idea of a connection pool. 

If maxPoolSize = 20, and minPoolSize = 0, I'd consider a `maxIdle` e.g. at 10. Then we assure pooling is effectively used and also enough connections are really closed when not in use.

tnleeuw:
> > When you are back, we might decide you finish the changes since you are already working on similar sets of changes.
> 
> Yes, I wasn't aware of the fact that only BTM understands that `maxIdleTime`. I think adding a `maxIdle` make sense. I would not put the setting at 1 because then it directly closes every open idle connection, which basically disables the idea of a connection pool.
> 
> If maxPoolSize = 20, and minPoolSize = 0, I'd consider a `maxIdle` e.g. at 10. Then we assure pooling is effectively used and also enough connections are really closed when not in use.

Some more settings should be added I think, for validation of connections while idle and also, importantly, for the time in between idle connection checking. Otherwise there's no regular validation of idle connections and if my understanding is correct, also not a cleanup of idle connections at regular intervals.

Those settings are only for DBCP2, not for BTM, but still needed in the BTM configuration properties for non-XA connections pools.

jkosternl:
> Some more settings should be added I think, for validation of connections while idle and also, importantly, for the time in between idle connection checking. Otherwise there's no regular validation of idle connections and if my understanding is correct, also not a cleanup of idle connections at regular intervals.

We're already setting `pooledConnectionFactory.setConnectionCheckInterval(connectionCheckInterval);` at `NarayanaConnectionFactoryFactory.java` file. That starts a connection thread which regularly checks connections already. 
 
> Those settings are only for DBCP2, not for BTM, but still needed in the BTM configuration properties for non-XA connections pools.

Ok, I thought those apply to Narayana already. I'll check for BTM.

core/src/main/resources/AppConstants.properties
```diff
@@ -358,7 +359,6 @@ msg.log.hidden.default=true
 msg.log.keys=conversationId
 # Define xPath for key conversationId
 msg.log.xPath.conversationId=*[local-name()='Envelope']/*[local-name()='Header']/*[local-name()='MessageHeader']/*[local-name()='HeaderFields']/*[local-name()='ConversationId']
-msg.log.xPath.value=/test/value
```
nielsm5, line 361: Oei! Goede...

core/src/main/java/nl/nn/adapterframework/jndi/JndiDataSourceFactory.java
```diff
@@ -15,38 +15,85 @@
...
+		maxIdle = appConstants.getInt("transactionmanager.jdbc.connection.maxIdle", maxIdle);
+		maxLifeTime = appConstants.getInt("transactionmanager.jdbc.connection.maxLifeTime", maxLifeTime);
+		connectionCheckInterval = appConstants.getInt("transactionmanager.jdbc.connection.checkInterval", connectionCheckInterval);
+		testQuery = appConstants.getString("transactionmanager.jdbc.connection.testQuery", testQuery);
 	}
 
 	protected DataSource augmentDatasource(CommonDataSource dataSource, String dataSourceName) {
+		if (dataSource instanceof XADataSource) {
+			return createXAPool((XADataSource) dataSource, dataSourceName);
+		}
```
nielsm5, line 74: Deze moet ook in de maxPoolSize if check?

core/src/main/java/nl/nn/adapterframework/jndi/JndiDataSourceFactory.java
```diff
@@ -15,38 +15,85 @@
...
+
+		log.info("DataSource [{}] is not XA enabled, unable to register with an Transaction Manager", dataSourceName);
```
nielsm5, line 76: Deze log klopt niet helemaal meer 😃

core/src/main/java/nl/nn/adapterframework/jndi/JndiDataSourceFactory.java
```diff
@@ -15,38 +15,85 @@
...
+		if(maxPoolSize > 1) {
+			return createPool((DataSource)dataSource);
+		}
 		return (DataSource) dataSource;
 	}
 
+	protected DataSource createXAPool(XADataSource xaDataSource, String dataSourceName) {
+		throw new UnsupportedOperationException("Creating XA DataSources not created by [" + this.getClass().getName() + "]");
```
nielsm5, line 84:
```suggestion
		throw new UnsupportedOperationException("non-XA DataSourcefactoryFactory [" + this.getClass().getName() + "] cannot create XA-DataSources");
```

nielsm5 (commented):
Ik denk dat je nog wel JJ's wijzigingen moet mergen?

tnleeuw:
> Ik denk dat je nog wel JJ's wijzigingen moet mergen?

Klopt maar dan moeten die eerst naar 7.9 gebackport en gemerged worden.

core/src/main/java/nl/nn/adapterframework/jndi/JndiDataSourceFactory.java
tnleeuw: Ik hernoem de method naar `createXADataSource()` om verwarring te voorkomen. Bij BTM wordt al altijd een pool gecreerd voor XA sessions, ongeacht maxPoolSize. Die zet daarin ook een aantal andere properties, die niet direct gerelateerd zijn aan pooling, dus ik durf daar niet teveel aan te veranderen.

Met een andere naam van de method is het hopelijk minder verwarrend.

nielsm5 (approved):

nielsm5:
Nu helaas conflicten 😞

tnleeuw:
> Nu helaas conflicten 😞

Dat wisten we, dat dat ging gebeuren.

Ik heb gisteren niet meer gekeken naar de mail of je nog dingen had gemerged, sorry.

core/src/main/java/nl/nn/adapterframework/jdbc/datasource/OpenManagedDataSource.java
```diff
@@ -29,14 +28,17 @@
...
-			GenericObjectPoolUtil.addPoolMetadata((GenericObjectPool<C>) objectPool, info);
-		}
+	@Override
+	protected GenericObjectPool<C> getPool() {
+		return (GenericObjectPool<C>) super.getPool();
+	}
+
+	@Override
+	public String toString() {
+		return "ManagedDataSource [], DBCP2 Pool Info: " + super.getPool();
```
nielsm5, line 42: Waarom de `[]`?

tnleeuw: Ik had er iets over de datasource tussen will zetten en toen bleek dat toch niet nodig.

nielsm5 (approved):

jkosternl (approved):

nielsm5 (approved):

PR #6120, opened 2024-01-16
Log file sorting

Matthbo:
The table of logs in the logging view can now be sorted on any of the named headers
* Added ThSortableDirective for any table element to make use of
* Changed `/logging` api endpoint to give back both `size` (long) & `sizeDisplay` (string) as property & removed the queryparam sizeFormat

### DIFF
--- a/console/frontend/src/main/frontend/src/app/components/th-sortable.directive.ts
+++ b/console/frontend/src/main/frontend/src/app/components/th-sortable.directive.ts
@@ -14,7 +14,6 @@ export function updateSortableHeaders(headers: QueryList<ThSortableDirective>, c
   }
 }
 
-
 const compare = (v1: string | number, v2: string | number) => (v1 < v2 ? -1 : v1 > v2 ? 1 : 0);
 export function basicTableSort<T extends Record<string, any>>(arr: T[], headers: QueryList<ThSortableDirective>, { column, direction }: SortEvent): T[] {
   updateSortableHeaders(headers, column);
@@ -31,8 +30,6 @@ export function basicTableSort<T extends Record<string, any>>(arr: T[], headers:
 @Directive({
   selector: 'th[sortable]',
   host: {
-    // '[class.asc]': 'direction === "asc"',
-    // '[class.desc]': 'direction === "desc"',
     '(click)': 'nextSort()',
   }
 })
--- a/console/frontend/src/main/frontend/src/app/components/th-sortable.directive.ts
+++ b/console/frontend/src/main/frontend/src/app/components/th-sortable.directive.ts
@@ -1,0 +1,73 @@
+import { Directive, ElementRef, EventEmitter, Input, OnInit, Output, QueryList } from '@angular/core';
+
+export type SortDirection = 'asc' | 'desc' | null;
+export type SortEvent = {
+  column: string;
+  direction: SortDirection;
+}
+
+export function updateSortableHeaders(headers: QueryList<ThSortableDirective>, column: string) {
+  for (const header of headers) {
+    if (header.sortable !== column) {
+      header.direction = null;
+    }
+  }
+}
+
+
+const compare = (v1: string | number, v2: string | number) => (v1 < v2 ? -1 : v1 > v2 ? 1 : 0);
+export function basicTableSort<T extends Record<string, any>>(arr: T[], headers: QueryList<ThSortableDirective>, { column, direction }: SortEvent): T[] {
+  updateSortableHeaders(headers, column);
+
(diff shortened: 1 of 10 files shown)


console/frontend/src/main/frontend/src/app/components/th-sortable.directive.ts
```diff
@@ -0,0 +1,76 @@
...
+export class ThSortableDirective implements OnInit {
+  @Input() sortable: string = '';
+  @Input() direction: SortDirection = null;
+  @Output() onSort = new EventEmitter<SortEvent>();
+
+  private nextSortOption: Record<string, SortDirection> = {
+    '': 'asc',
+    asc: 'desc',
+    desc: null
+  }
```
philipsens, line 45: Would this be possible? If so, I think it is more logical to have them be the same type. 

```suggestion
  private nextSortOption: Record<SortDirection, SortDirection> = {
    null: 'asc',
    asc: 'desc',
    desc: null
  }
```

console/frontend/src/main/frontend/src/app/views/logging/logging.component.ts
```diff
@@ -84,20 +87,28 @@ export class LoggingComponent implements OnInit {
...
+        this.alert = false;
+        this.originalList = data.list;
+        this.sortedlist = data.list;
+        this.directory = data.directory;
+        this.path = data.directory;
+        if (data.count > data.list.length) {
+          this.alert = "Total number of items [" + data.count + "] exceeded maximum number, only showing first [" + (data.list.length - 1) + "] items!";
+        }
+      }, error: (data) => {
+        this.alert = (data.error) ? data.error.error : "An unknown error occured!";
```
philipsens, line 101:
```suggestion
        this.alert = data.error?.error || "An unknown error occured!";
```
Maybe a bit nicer syntactically

console/frontend/src/main/frontend/src/app/views/logging/logging.component.ts
```diff
@@ -84,20 +87,28 @@ export class LoggingComponent implements OnInit {
...
-        this.alert = "Total number of items [" + data.count + "] exceeded maximum number, only showing first [" + (data.list.length - 1) + "] items!";
+    this.loggingService.getLogging(directory).subscribe({
+      next: (data) => {
+        this.alert = false;
+        this.originalList = data.list;
+        this.sortedlist = data.list;
+        this.directory = data.directory;
+        this.path = data.directory;
+        if (data.count > data.list.length) {
+          this.alert = "Total number of items [" + data.count + "] exceeded maximum number, only showing first [" + (data.list.length - 1) + "] items!";
```
philipsens, line 98: I prefer template strings, if the string consists of more than 2 parts. Has there been any discussion about this in the past?
```suggestion
          this.alert = `Total number of items [${data.count}] exceeded maximum number, only showing first [${data.list.length - 1}] items!`;
```

console/frontend/src/main/frontend/src/app/components/th-sortable.directive.ts
Matthbo: Sadly `null` cant be an object key, thats the only reason I used an empty string and a null check instead

console/frontend/src/main/frontend/src/app/views/logging/logging.component.ts
Matthbo: This is basically still Niels' code so I didnt even look at it all that much, however I do agree with some minor exceptions, but your suggestion definitely looks better here

core/src/main/java/org/frankframework/util/Dir2Map.java
```diff
@@ -93,13 +91,14 @@ private Map<String, Object> FileInfo(File file) {
...
-		Map<String, Object> fileInfo = new HashMap<>(5);
+		Map<String, Object> fileInfo = new HashMap<>(6);
 
 		fileInfo.put("name", displayName);
 		fileInfo.put("path", normalizePath(file));
 		fileInfo.put("lastModified", file.lastModified());
 		fileInfo.put("type", file.isDirectory() ? "directory" : "file");
-		fileInfo.put("size", (sizeFormat) ? Misc.toFileSize(file.length(), true) : file.length());
+		fileInfo.put("size", file.length());
+		fileInfo.put("sizeDisplay", Misc.toFileSize(file.length(), true));
```
tnleeuw, line 101: Instead of fetching file size twice it is slightly faster to fetch it once and put it into a local variable.

philipsens (approved):

Issue #6113, opened 2024-01-15
FR: Aspose multi threaded email attachment conversion


Laurens-makel:
**❗ please do not add sensitive information in issues, you can provide extra information via email using issue number as reference ❗**

**Describe the issue**
In the current `MailConvertor` implementation, the discovered attachments of an e-mail are converted to PDF in a single thread resulting in sub-optimal processing performance.

Ideally this work is divided by a configurable amount of threads.

Keep in mind that attachments can have attachments, threads should be limited to a maximum and share a 'queue' of work

evandongen:
After merging #10619, this should be fixed

PR #6134, opened 2024-01-17
Multi threaded email attachment conversion

Laurens-makel:
#6113

Initial draft, haven't tested this yet, will do this asap.

Open for a first review & ideas to 'proof' this behaviour using unit tests.

### DIFF
--- a/aspose/src/main/java/org/frankframework/extensions/aspose/services/conv/impl/convertors/MailConvertor.java
+++ b/aspose/src/main/java/org/frankframework/extensions/aspose/services/conv/impl/convertors/MailConvertor.java
@@ -21,10 +21,17 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Files;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 
 import org.frankframework.extensions.aspose.services.conv.CisConfiguration;
 import org.frankframework.extensions.aspose.services.conv.CisConversionResult;
@@ -63,6 +70,8 @@ class MailConvertor extends AbstractConvertor {
 	private static final float MAX_IMAGE_HEIGHT_IN_POINTS = PageConvertUtil.convertCmToPoints(PageConvertUtil.PAGE_HEIGTH_IN_CM - 2 * 1.1f);
 	private static final String MAIL_HEADER_DATEFORMAT = "dd-MM-yyyy HH:mm:ss";
 	private final CisConversionService cisConversionService;
+	private final ExecutorService attachmentConversionExecutor;
+
 
 	// contains mapping from MediaType to the LoadOption for the Aspose Word conversion.
 	private static final Map<MediaType, Class<? extends LoadOptions>> MEDIA_TYPE_LOAD_FORMAT_MAPPING;
@@ -90,6 +99,7 @@ class MailConvertor extends AbstractConvertor {
 	protected MailConvertor(CisConversionService cisConversionService, CisConfiguration configuration) {
 		super(configuration, MEDIA_TYPE_LOAD_FORMAT_MAPPING.keySet());
 		this.cisConversionService = cisConversionService;
+		this.attachmentConversionExecutor = Executors.newFixedThreadPool(configuration.getMaxThreads());
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


tnleeuw:
Given the failing tests I've converted the PR to "Draft" for now.

tnleeuw:
Did you now test these changes with multiple threads?

Would it be possible to test this in unit-tests with multiple threads?

Laurens-makel:
> Did you now test these changes with multiple threads?
> 
> Would it be possible to test this in unit-tests with multiple threads?

Haven't found the time today to test this in the desired application, hopefully tomorrow.

Do you have any ideas/suggestions on how to implement such tests?

Laurens-makel:
Added some log statements to the method that should be executed multi threaded:
image

Logs:

```
Thread [pool-4-thread-2] with ID [2648]: Starting conversion for [RE: Constatering in X mail met knipsel erin geplakt is niet te lezen wordt X ] at 11:42:15.927
Thread [pool-4-thread-1] with ID [2647]: Starting conversion for [BP1.PNG] at 11:42:15.927
Thread [pool-4-thread-3] with ID [2652]: Starting conversion for [Test emailbericht met knipsel] at 11:42:18.130
Thread [pool-4-thread-3] with ID [2652]: Finished conversion for [Test emailbericht met knipsel] at 11:42:18.916
Thread [pool-4-thread-2] with ID [2648]: Finished conversion for [RE: Constatering in X mail met knipsel erin geplakt is niet te lezen wordt X ] at 11:42:18.916
Thread [pool-4-thread-1] with ID [2647]: Finished conversion for [BP1.PNG] at 11:42:19.038
```

tnleeuw:
Could you do the unit test with multiple threads configured, let's say max 2 threads? And then have an input that has at least 3 items?

If you log both thread entry / exit it should show better actual simultanuous processing and then the thread safety can perhaps be assessed a bit better.

What do you think? Is that realistic?

Laurens-makel:
> Could you do the unit test with multiple threads configured, let's say max 2 threads? And then have an input that has at least 3 items?
> 
> If you log both thread entry / exit it should show better actual simultanuous processing and then the thread safety can perhaps be assessed a bit better.
> 
> What do you think? Is that realistic?

If I understand your proposal correctly, we should assert the order of what has been logged?

A bit more specific, given your situation with 2 threads & 3 items, we would assert the following order of logs:

- Thread 1 - start item 1
- Thread 2 - start item 2
- Thread 1 - finish item 1
- Thread 1 - start item 3
- Thread 2 - finish item 2
- Thread 1 - finish item 3

tnleeuw:
No I don't think it is possible to assert anything like that in a multi-threaded execution environment.

We can only look at logfiles and try to see if there was some kind of concurrent execution, but I don't know of a way to verify that automatically.

However having multiple threads and more tasks than threads increases chances of concurrent execution and thus revealing any kind concurrency errors. (Such as potential concurrent modification of shared state in the Aspose library that we're not aware of).

What do you think? Any thoughts?

Laurens-makel:
> No I don't think it is possible to assert anything like that in a multi-threaded execution environment.
> 
> We can only look at logfiles and try to see if there was some kind of concurrent execution, but I don't know of a way to verify that automatically.
> 
> However having multiple threads and more tasks than threads increases chances of concurrent execution and thus revealing any kind concurrency errors. (Such as potential concurrent modification of shared state in the Aspose library that we're not aware of).
> 
> What do you think? Any thoughts?

Are you sure this is not possible?

We submit tasks in the same order, if we have two threads and two items, we can assert two log lines stating `starting` before two log lines stating `finished`, that is the proof of multi threaded processing.

Regarding concurrency errors, if they are present, wouldn't they also occur if the adapter is processing messages with multiple threads? We have been running adapters with the `PdfPipe` with multiple threads in the past and didn't experience any issues

aspose/src/main/java/org/frankframework/extensions/aspose/services/conv/impl/convertors/MailConvertor.java
```diff
@@ -90,6 +97,7 @@ class MailConvertor extends AbstractConvertor {
 	protected MailConvertor(CisConversionService cisConversionService, CisConfiguration configuration) {
 		super(configuration, MEDIA_TYPE_LOAD_FORMAT_MAPPING.keySet());
 		this.cisConversionService = cisConversionService;
+		this.attachmentConversionExecutor = Executors.newFixedThreadPool(configuration.getMaxThreads());
```
tnleeuw, line 100: I was wondering if it would be a good idea to use here a https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html and in the `convert()` method use a parallelStream() with this custom ForkJoinPool to do the multi-threaded conversion and collection of results?

See also:
https://www.baeldung.com/java-8-parallel-streams-custom-threadpool

Laurens-makel: The problem is the type of `Attachment` currently does not support `.stream()` methods.

Although this problem could be solved by changing the underlying base type of `Attachment`.

tnleeuw: Ah, I see.

Well, it was just a suggestion. If it won't work, then no problem.

Laurens-makel: I mean `AttachmentCollection`, not `Attachment` itself..

But since these are `Aspose` classes,  it's a bit hard to solve it.

tnleeuw:
To give an example of what can happen with multiple threads, the exeuction order could be:

- T1 start
- T2 start
- T1 end
- T3 start
- T2 end
- T3 end

Or it could be:
- T1 start
- T1 end
- T2 start
- T2 end
- T3 start
- T3 end

That all depends on scheduling, the code executed in each thread (the I/O done for instance), etc.
But even if you can see all thread-starts before any thread-ends, there's no proof of actual concurrent execution on multiple CPUs.

You can look at this in your Process Monitor, but that is hard to automate I think.

tnleeuw:
> Regarding concurrency errors, if they are present, wouldn't they also occur if the adapter is processing messages with multiple threads? We have been running adapters with the `PdfPipe` with multiple threads in the past and didn't experience any issues

I was concerned about any shared state between the various Aspose objects used in each process, such as the email attachments linked to the same e-mail. I don't know what their domain-model is.

Laurens-makel:
> To give an example of what can happen with multiple threads, the exeuction order could be:
> 
> * T1 start
> * T2 start
> * T1 end
> * T3 start
> * T2 end
> * T3 end
> 
> Or it could be:
> 
> * T1 start
> * T1 end
> * T2 start
> * T2 end
> * T3 start
> * T3 end
> 
> That all depends on scheduling, the code executed in each thread (the I/O done for instance), etc. But even if you can see all thread-starts before any thread-ends, there's no proof of actual concurrent execution on multiple CPUs.
> 
> You can look at this in your Process Monitor, but that is hard to automate I think.

I understand the problem with the execution order, however there are a couple factors in play here. The speed of the conversion is depended on the size and type of the file.

1. PDF = near instant, doesnt need conversion.
2. Excel file 50mb = yes a rare case but this appearently could take up to 5 minutes.
3. Word file 1-5mb = depending on the size & content, it would take some time but not too long.

This would create a more predictable workload and execution order, which I think could asserted.

aspose/src/main/java/org/frankframework/extensions/aspose/services/conv/impl/convertors/MailConvertor.java
tnleeuw: Actually I found you could do this:

`StreamSupport.stream(attachments.spliterator(), true)`

tnleeuw:
> I understand the problem with the execution order, however there are a couple factors in play here. The speed of the conversion is depended on the size and type of the file.
> 
>     1. PDF = near instant, doesnt need conversion.
> 
>     2. Excel file 50mb = yes a rare case but this appearently could take up to 5 minutes.
> 
>     3. Word file 1-5mb = depending on the size & content, it would take some time but not too long.
> 
> 
> This would create a more predictable workload and execution order, which I think could asserted.

Hm, but by then it's doubtful you'd want this in your unit-tests. Aspose tests already take quite a long time of the overall build time.

Laurens-makel:
> > Regarding concurrency errors, if they are present, wouldn't they also occur if the adapter is processing messages with multiple threads? We have been running adapters with the `PdfPipe` with multiple threads in the past and didn't experience any issues
> 
> I was concerned about any shared state between the various Aspose objects used in each process, such as the email attachments linked to the same e-mail. I don't know what their domain-model is.

Yes, unfortunately their code is obfuscated but I don't really see any possible dangerous references being passed to the `this.d.addItem()` method in `com.aspose.email.MailMessage`. In this context `this.d` stands for `AttachmentCollection` being retrieved & used.

aspose/src/main/java/org/frankframework/extensions/aspose/services/conv/impl/convertors/MailConvertor.java
Laurens-makel: > Actually I found you could do this:
> 
> `StreamSupport.stream(attachments.spliterator(), true)`

Will investigate in the weeknd into this !

Laurens-makel: Ive tried something but I dont think this is much better

nielsm5:
This has been in draft for 8 months, I'm closing this but will keep the branch.

PR #6143, opened 2024-01-19
Fix reading bytes message data after extracting message properties.

tnleeuw:


Yale96 (approved):

core/src/main/java/nl/nn/adapterframework/jms/BytesMessageInputStream.java
```diff
@@ -32,14 +33,19 @@ public BytesMessageInputStream(BytesMessage bytesMsg) {
 	@Override
 	public int read() throws IOException {
 		try {
-			return bytesMsg.readByte();
+			byte[] data = new byte[1];
+			int len = bytesMsg.readBytes(data);
+			if (len == -1) {
+				return -1;
+			}
+			return data[0] & 0xFF;
```
jkosternl, line 41: Why this interesting bit flipping? Can you add a comment on the purpose of this magic?

core/src/main/java/nl/nn/adapterframework/jms/BytesMessageInputStream.java
```diff
@@ -48,12 +54,17 @@ public int read(byte[] b) throws IOException {
...
-	public int read(byte[] b, int off, int len) throws IOException {
+	public int read(@Nonnull byte[] b, int off, int len) throws IOException {
 		try {
-			byte[] readbuf = new byte[len];
-			int result = bytesMsg.readBytes(readbuf);
+			// If we try filling the whole array from start, then we do not need to read data into
+			// a temp array and copy it.
+			if (off == 0 && len == b.length) {
+				return read(b);
+			}
```
jkosternl, line 63: Just to understand your fix: this if-statement at line 61 is an optimization right, not a fix?

core/src/main/java/nl/nn/adapterframework/extensions/esb/EsbJmsListener.java
```diff
@@ -136,34 +136,37 @@ public Map<String, Object> extractMessageProperties(Message rawMessage) throws L
...
-				TextMessage textMessage = (TextMessage) rawMessage;
-				soapMessage = textMessage.getText();
-			} else if (rawMessage instanceof BytesMessage) {
-				BytesMessage bytesMessage = (BytesMessage) rawMessage;
-				InputStream input = new BytesMessageInputStream(bytesMessage);
-				soapMessage = StreamUtil.streamToString(input);
-			} else {
-				soapMessage = null;
-			}
+		if (!getxPathLogMap().isEmpty()) {
```
jkosternl, line 139:
```suggestion
		if (getxPathLogMap().isEmpty()) {
		     return messageProperties;
		}
```

jkosternl (approved):

core/src/main/java/nl/nn/adapterframework/jms/BytesMessageInputStream.java
tnleeuw: This was suggested by IntelliJ to ensure the return data was unsigned. You are absolutely right this needs a comment with explanation.

core/src/main/java/nl/nn/adapterframework/extensions/esb/EsbJmsListener.java
```diff
@@ -136,34 +136,37 @@ public Map<String, Object> extractMessageProperties(Message rawMessage) throws L
...
+				if (rawMessage instanceof TextMessage) {
+					TextMessage textMessage = (TextMessage) rawMessage;
+					soapMessage = textMessage.getText();
+				} else if (rawMessage instanceof BytesMessage) {
+					BytesMessage bytesMessage = (BytesMessage) rawMessage;
+					InputStream input = new BytesMessageInputStream(bytesMessage);
+					soapMessage = StreamUtil.streamToString(input);
+					bytesMessage.reset();
+				} else {
+					soapMessage = null;
```
Laurens-makel, line 151: Just wondering, is this explicit `null` assignment needed or is it just a code style preference?

Laurens-makel: I thought that a `String` is by default `null` when it's not initialized right?

tnleeuw: It's a local variable, not a field in the class. So when not initialized it has no value. You cannot access it unless it's explicitly assigned -- the compiler will give an error.

Laurens-makel: I understand, although I think it's a bit silly that local variables behave differently.

Thanks for the clarification :)

tnleeuw: Actually I'm glad they behave the way they do, as it allows to catch accidentally unitialized variables.

Laurens-makel: > Actually I'm glad they behave the way they do, as it allows to catch accidentally unitialized variables.

That's a good argument.

For me it's the inconsistency, apparently only null is assigned to class member variables.. why wouldn't the same concept apply here?

tnleeuw: Yes, I agree.

jkosternl:
Looking good now, with the extra changes. Thanks 👍🏻

nielsm5 (approved):

tnleeuw:
Since I'm still adding a test for IAF-Test I put the PR back into "Draft" for now, I expect to be ready with that later today.

tnleeuw:
> Since I'm still adding a test for IAF-Test I put the PR back into "Draft" for now, I expect to be ready with that later today.

I've now added integration tests, verified that the integration test with BytesMessage indeed breaks without the applied fixes.

jkosternl (approved):

nielsm5 (approved):

PR #6151, opened 2024-01-23
Fix configuration versions page & api giving back sql error

Matthbo:
* Fixes configuration details page showing `[Object object]` instead of configuration name
* Fixes sql error caused by `rs.getDate` not being able to convert TIMESTAMP column in `ConfigManagement.getConfigsFromDatabase`


philipsens (approved):
Thank you!

nielsm5 (approved):
Lijkt gewoon te werken! (forwards-compatible)

PR #6155, opened 2024-01-24
Don't close CMIS Session & Bindings before reading the stream

tnleeuw:
No tests yet, need to think about how to make a good test for this.

### DIFF
--- a/cmis/src/main/java/nl/nn/adapterframework/extensions/cmis/CmisSessionWrapper.java
+++ b/cmis/src/main/java/nl/nn/adapterframework/extensions/cmis/CmisSessionWrapper.java
@@ -1,0 +1,47 @@
+/*
+   Copyright 2024 WeAreFrank!
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+ */
+package nl.nn.adapterframework.extensions.cmis;
+
+import org.apache.chemistry.opencmis.client.api.Session;
+import org.apache.chemistry.opencmis.commons.SessionParameter;
+import org.apache.chemistry.opencmis.commons.spi.CmisBinding;
+
+import lombok.Getter;
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
+public class CmisSessionWrapper implements AutoCloseable {
+
+	private final @Getter Session cmisSession;
+
+	public CmisSessionWrapper(Session cmisSession) {
+		this.cmisSession = cmisSession;
+	}
+
+	@Override
+	public void close() {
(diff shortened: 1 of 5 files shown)


jkosternl (approved):

cmis/src/main/java/nl/nn/adapterframework/extensions/cmis/CmisSessionWrapper.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.apache.chemistry.opencmis.client.api.Session;
+import org.apache.chemistry.opencmis.commons.SessionParameter;
+import org.apache.chemistry.opencmis.commons.spi.CmisBinding;
+
+import lombok.Getter;
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
+public class CmisSessionWrapper implements AutoCloseable {
```
nielsm5, line 26: Niet extends `SessionImpl`?

tnleeuw: Dan moeten we ook de `SessionFactory` extenden, want die creeert de instance:

```
		SessionFactory sessionFactory = SessionFactoryImpl.newInstance();
		Session session = sessionFactory.createSession(parameterMap);
		log.debug("connected with repository [{}]", ()->getRepositoryInfo(session));

		return new CmisSessionWrapper(session);
```

nielsm5: Ja klopt, die doet new SessionImpl(parameters, null, null, null, null)

Laurens-makel:
Just wondering and trying to learn a bit here, the initial solution with a private inner class which takes the `Session` as constructor argument, is only used when runtime sessions are used.

What is the advantage of **always** wrapping the `Session` object created instead? It doesn't have to be an `AutoClosable` for global sessions right? Only when we want to schedule the close on pipeline exit..

nielsm5:
> Just wondering and trying to learn a bit here, the initial solution with a private inner class which takes the `Session` as constructor argument, is only used when runtime sessions are used.
> 
> What is the advantage of **_always_** wrapping the `Session` object created instead? It doesn't have to be an `AutoClosable` for global sessions right? Only when we want to schedule the close on pipeline exit..

I think in this instance it's easier and makes the code more readable to always wrap the session.

nielsm5 (approved):

cmis/src/main/java/nl/nn/adapterframework/extensions/cmis/CmisSessionWrapper.java
tnleeuw: Ik zal dat doen wanneer ik een forward-port doe van de fix naar 7.9/8.0/master.
Voor 7.8 wil ik het zo laten als het nu is omdat het nu getest is zo door NN.

tnleeuw:
> > > Just wondering and trying to learn a bit here, the initial solution with a private inner class which takes the `Session` as constructor argument, is only used when runtime sessions are used.
> > > What is the advantage of **_always_** wrapping the `Session` object created instead? It doesn't have to be an `AutoClosable` for global sessions right? Only when we want to schedule the close on pipeline exit..
> > 
> > 
> > I think in this instance it's easier and makes the code more readable to always wrap the session.
> 
> In my opinion, these changes makes the code actually less readable and could be avoided.
> 

I think it makes some parts more readable, and others get a bit an extra call.

For 7.9 and forward I will try to extend the session / sessionfactory instead of wrapping it, as Niels suggested, and that would address all concerns regarding code quality and consistency.

PR #6168, opened 2024-01-26
Make it possible to change the Tomcat tag in the Docker image

philipsens:


docker/Tomcat/webapp/Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM tomcat:9.0.83-jre11
+ARG TOMCAT_TAG=9.0.83-jre11
```
nielsm5, line 1:
```suggestion
ARG TOMCAT_TAG=9.0.85-jre11
```

nielsm5 (approved):

PR #6169, opened 2024-01-26
Do not close CMIS session before PipeLineSession closes

tnleeuw:


jkosternl (approved):
Assumed this is the same code as 7.8

nielsm5 (approved):

tnleeuw:
> Assumed this is (roughly) the same code as 7.8

Not it's not and I see I somehow made tests fail as well -- oops. Committed too hastily again.

jkosternl (approved):

PR #6172, opened 2024-01-26
Make LogContextPipe not throw when it has an error during execution

tnleeuw:


core/src/main/java/nl/nn/adapterframework/pipes/LogContextPipe.java
```diff
@@ -56,12 +57,15 @@ public PipeRunResult doPipe(Message message, PipeLineSession session) throws Pip
...
+					message.preserve();
+				}
 				ParameterValueList pvl = getParameterList().getValues(message, session);
 				for(ParameterValue pv : pvl) {
 					values.put(pv.getName(), pv.asStringValue());
 				}
-			} catch (ParameterException e) {
-				throw new PipeRunException(this, "exception extracting parameters", e);
+			} catch (Exception e) {
+				log.debug("Exception getting parameter values. Ignoring.", e);
```
Laurens-makel, line 68: Could this be changed to a `warn` statement?

If you see this a lot in production, something might be wrongly/sub-optimal configured..

tnleeuw: I was doubting because some messages we did make warnings, turned out to be excessive without pointing to any kind of misconfiguration.
But in this case you're probably right.

core/src/main/java/nl/nn/adapterframework/pipes/LogContextPipe.java
```diff
@@ -56,12 +57,15 @@ public PipeRunResult doPipe(Message message, PipeLineSession session) throws Pip
...
+					message.preserve();
+				}
 				ParameterValueList pvl = getParameterList().getValues(message, session);
 				for(ParameterValue pv : pvl) {
 					values.put(pv.getName(), pv.asStringValue());
 				}
-			} catch (ParameterException e) {
-				throw new PipeRunException(this, "exception extracting parameters", e);
+			} catch (Exception e) {
+				log.warn("Exception getting parameter values. Ignoring.", e);
```
jkosternl, line 68: Just wondering if this is enough info for the end user (integration specialist) to understand/find the configuration which is wrong. @Laurens-makel what is your opinion?

jkosternl (approved):

core/src/main/java/nl/nn/adapterframework/pipes/LogContextPipe.java
```diff
@@ -56,12 +57,15 @@ public PipeRunResult doPipe(Message message, PipeLineSession session) throws Pip
 		if (!getParameterList().isEmpty()) {
 			Map<String,String> values = new LinkedHashMap<>();
 			try {
+				if (!message.isRepeatable()) {
+					message.preserve();
+				}
```
nielsm5, line 62: Dit is onderdeel van de ParameterValueList en hoeft hier niet.

core/src/main/java/nl/nn/adapterframework/pipes/LogContextPipe.java
Laurens-makel: Depends on what is in `e`... Does it mention which parameter caused the exception?

jkosternl: > Depends on what is in `e`... Does it mention which parameter caused the exception?
No, it only mentions the stacktrace, which isn't helping.

jkosternl (approved):

nielsm5 (approved):

PR #6174, opened 2024-01-26
Change compiler source/target to 17

philipsens:
This is to see if the unit and larva tests in the CI will pass.


pom.xml
```diff
@@ -9,8 +9,8 @@
 	<properties>
 		<revision>8.1.0-SNAPSHOT</revision>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		<maven.compiler.target>11</maven.compiler.target>
-		<maven.compiler.source>11</maven.compiler.source>
+		<maven.compiler.target>17</maven.compiler.target>
+		<maven.compiler.source>17</maven.compiler.source>
```
tnleeuw, line 13: Does this mean we will drop support for JDK11 already?

tnleeuw (changes requested):
Changing the source version and target without changing anything else in the project breaks the build in Github and Jenkins.

pom.xml
jkosternl: This is too fast.. `compiler.source` should still be at 11, I think. I think it would be wise to move to 17 language level when we don't need to do often backports to < v8.0 as well.

PR #6175, opened 2024-01-26
Add ConnectionPool information to SecurityItems

jkosternl:


core/src/main/java/nl/nn/adapterframework/jdbc/dbms/IDbmsSupportFactory.java
```diff
@@ -30,5 +30,6 @@ public interface IDbmsSupportFactory {
 	final int DBMS_MYSQL=5;
 
 	IDbmsSupport getDbmsSupport(DataSource datasource);
+	IDbmsSupport getDbmsSupport(String product, String productVersion);
```
jkosternl, line 33: Somehow I needed to change this interface, which wasn't there in 8.0 or higher

tnleeuw: In 8.0 and higher that interface is moved to another module, I believe? I wonder why it needs to be changed to accept strings instead of the actual dataSource, will look at the rest of the changes.

tnleeuw: Checking it, in 8.0 the interface was removed as it was deemed unnecessary. In 8.0 this method with 2 String parameters is called from the JdbcFacade, which has a DataSource and in 8.0 the DbmsSupportFactory class has a method to get an instance taking a Connection, and one taking these Strings.

I think in 7.9 you probably do not need to make this method public? Since the JdbcFacade has a DataSource and there's a method to get the implementation from the DataSource?

core/src/main/java/nl/nn/adapterframework/jdbc/JdbcFacade.java
```diff
@@ -115,43 +117,36 @@ protected DataSource getDatasource() throws JdbcException {
...
 	public IDbmsSupport getDbmsSupport() {
-		if (dbmsSupport==null) {
-			try {
+		if (dbmsSupport != null) {
+			return dbmsSupport;
+		}
+		try {
+			if (datasource instanceof TransactionalDbmsSupportAwareDataSourceProxy) {
+				Map<String, String> md = ((TransactionalDbmsSupportAwareDataSourceProxy) datasource).getMetaData();
+				dbmsSupport = dbmsSupportFactory.getDbmsSupport(md.get("product"), md.get("product-version"));
```
tnleeuw, line 143: I think you can get the instance directly from the DataSource in 7.9 and don't need to extract all this meta-data and make the other method public?

tnleeuw (changes requested):

core/src/main/java/nl/nn/adapterframework/jdbc/JdbcFacade.java
jkosternl: Ok, will do that.

core/src/main/java/nl/nn/adapterframework/jdbc/dbms/IDbmsSupportFactory.java
jkosternl: Ok, will do that.

tnleeuw (approved):

jkosternl:
> Nu wordt de (voorheen) enige reden dat de TransactionalDbmsSupportAwareDataSourceProxy class bestaat niet meer toegepast. De DbmsSupportFactory moet juist uit een gecachede waarde kunnen halen wat het 'product' is. Anders is voor elk database element een connectie nodig.

In de 8.0 en hogere versies werkte deze aanpassing wel. Omdat `DbmsSupportFactory.getDbmsSupport` alleen werd aangeroepen vanuit `JdbcFacade.getDbmsSupport()` en daar zit nu de check op `TransactionalDbmsSupportAwareDataSourceProxy`. Dus het lijkt idd vreemd, dat ik dit zo had gedaan, maar in 8.0 is het prima en in 7.9 idd niet. Daarom heb ik het hier weer teruggezet.

nielsm5 (approved):

nielsm5 (approved):

PR #6193, opened 2024-01-30
Convert MessageStoreListenerTest

nielsm5:
Nog even een handigheidje verzinnen om niet alle transactionmanagers te testen...

### DIFF
--- a/core/src/test/java/org/frankframework/jdbc/MessageStoreListenerTest.java
+++ b/core/src/test/java/org/frankframework/jdbc/MessageStoreListenerTest.java
@@ -21,6 +21,7 @@ import org.frankframework.receivers.MessageWrapper;
 import org.frankframework.receivers.RawMessageWrapper;
 import org.frankframework.testutil.junit.DatabaseTest;
 import org.frankframework.testutil.junit.DatabaseTestEnvironment;
+import org.frankframework.testutil.junit.TxManagerTest;
 import org.frankframework.testutil.junit.WithLiquibase;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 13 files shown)


core/src/test/java/org/frankframework/jdbc/MessageStoreListenerTest.java
```diff
@@ -5,7 +5,7 @@
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assume.assumeThat;
-import static org.mockito.Mockito.doAnswer;
+import static org.junit.Assume.assumeTrue;
```
tnleeuw, line 8: Dit zijn nog allemaal JUnit4 methods.

tnleeuw (changes requested):

core/src/test/java/org/frankframework/jdbc/MessageStoreListenerTest.java
nielsm5: Grrr...

core/src/test/java/org/frankframework/jdbc/MessageStoreListenerTest.java
```diff
@@ -36,7 +34,7 @@ public class MessageStoreListenerTest {
 
 	@BeforeEach
 	public void setup(DatabaseTestEnvironment env) throws Exception {
-		assumeThat(env.getDbmsSupport().getDbms(),equalTo(Dbms.H2)); // tests are based on H2 syntax queries
+		assumeTrue(Dbms.H2 == env.getDbmsSupport().getDbms()); // tests are based on H2 syntax queries
```
tnleeuw, line 37: So the tests are only executed for H2 and for all other databases they are skipped? Is that right actually, that we will never test against other databases when they're running?

jkosternl (approved):

PR #6194, opened 2024-01-30
Improve reliability of ParallelSender and ShadowSender

tnleeuw:
- Fix ParallelSender and ShadowSender when messages are not repeatable
- Fix ShadowSender when not waiting for all senders to finish, and pipeline finishes before all senders have finished.
-

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/senders/ShadowSender.java
+++ b/core/src/main/java/nl/nn/adapterframework/senders/ShadowSender.java
@@ -1,5 +1,5 @@
 /*
-   Copyright 2018 Nationale-Nederlanden, 2020, 2022 WeAreFrank!
+   Copyright 2018 Nationale-Nederlanden, 2020, 2022, 2024 WeAreFrank!
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
@@ -21,6 +21,9 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 import org.apache.commons.lang3.StringUtils;
 import org.xml.sax.SAXException;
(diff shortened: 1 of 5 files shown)


core/src/test/java/nl/nn/adapterframework/senders/ShadowSenderTest.java
```diff
@@ -211,6 +212,60 @@ public void testResultSenderResultWith3ShadowSenders() throws Exception {
...
+
+		// Act
+		inputMessage.closeOnCloseOf(session, sender);
+		String result = sender.sendMessageOrThrow(inputMessage, session).asString();
+		session.close();
+
+		// Assert
+		assertEquals(ORIGINAL_SENDER_RESULT, result);
+
+		Thread.sleep(3000); // wait for results to be collected in the background
```
jkosternl, line 236: Is it easy to convert this to use `await()` method, so this unittest perhaps runs faster when possible?

jkosternl (approved):
Just wondering why it worked before 😉

tnleeuw:
> Just wondering why it worked before 😉

I don't think it ever did, but some things were just never used in that particular way. 😃

core/src/test/java/nl/nn/adapterframework/senders/ShadowSenderTest.java
tnleeuw: Not sure. Will look at it.

core/src/main/java/nl/nn/adapterframework/senders/ShadowSender.java
```diff
@@ -137,30 +152,37 @@ public SenderResult sendMessage(Message message, PipeLineSession session) throws
 		try {
 			primaryGuard.waitForAllResources();
 		} catch (InterruptedException e) {
-			throw new SenderException(getLogPrefix()+"was interupted", e);
+			throw new SenderException(getLogPrefix() + "was interrupted", e);
```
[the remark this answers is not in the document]
tnleeuw: I was unsure. I think we probably should.

nielsm5 (approved):

PR #6198, opened 2024-01-31
Make MermaidFlowGenerator more resilient to class loading errors

tnleeuw:
I might have a test for this too but it requires changing the scanner to also allow scanning classes on the test classpath and I need to verify that there are no negative side-effects on other tests.

### DIFF
--- a/core/src/main/java/org/frankframework/util/flow/MermaidFlowGenerator.java
+++ b/core/src/main/java/org/frankframework/util/flow/MermaidFlowGenerator.java
@@ -93,11 +93,17 @@ public class MermaidFlowGenerator implements IFlowGenerator {
 		transformerPoolConfig = TransformerPool.getInstance(xsltSourceConfig, 2);
 	}
 
-	private String compileFrankElementList() throws SAXException, ClassNotFoundException {
+	private String compileFrankElementList() throws SAXException {
 		try (SaxDocumentBuilder builder = new SaxDocumentBuilder("root")) {
 			List<String> classNames = findAllFrankElements();
 			for(String className : classNames) {
-				Class<?> clazz = Class.forName(className);
+				Class<?> clazz;
+				try {
+					clazz = Class.forName(className);
+				} catch (Exception e) {
+					log.info("Found class [{}] on classpath which cannot be loaded due to exception", className, e);
+					continue;
+				}
 				ElementType type = AnnotationUtils.findAnnotation(clazz, ElementType.class);
 				if(type != null) {
 					try (SaxElementBuilder classElement = builder.startElement(className)) {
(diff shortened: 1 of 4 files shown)


tnleeuw:
@nielsm5 I added some tests in a separate commit. If you do not like those changes, it's best I have to adding a unit test for this case. We can revert that commit but keep the actual fixes. :-)

jacodg (approved):

core/src/main/java/org/frankframework/util/flow/MermaidFlowGenerator.java
```diff
@@ -93,11 +93,17 @@ public void afterPropertiesSet() throws Exception {
...
+	private String compileFrankElementList() throws SAXException {
 		try (SaxDocumentBuilder builder = new SaxDocumentBuilder("root")) {
 			List<String> classNames = findAllFrankElements();
 			for(String className : classNames) {
-				Class<?> clazz = Class.forName(className);
+				Class<?> clazz;
+				try {
+					clazz = Class.forName(className);
+				} catch (ExceptionInInitializerError | NoClassDefFoundError | Exception e) {
+					log.info("Found class [{}] on classpath which cannot be loaded due to exception", className, e);
```
jkosternl, line 104: This is not a serious message that should warn an user / integration specialist?

core/src/test/java/org/frankframework/configuration/digester/TestAnnotationUtils.java
```diff
@@ -219,7 +219,10 @@ protected String buildDefaultBeanName(@Nonnull BeanDefinition definition) {
 		Set<Class<?>> interfazes = new HashSet<>();
 		String[] names = scanner.getRegistry().getBeanDefinitionNames();
 		for (String beanName : names) {
-			if(beanName.contains(this.getClass().getCanonicalName()) || beanName.startsWith("org.frankframework.credentialprovider")) continue; //Ignore this class
+			if(beanName.contains(this.getClass().getCanonicalName())
+					|| beanName.startsWith("org.frankframework.credentialprovider")
+					|| beanName.endsWith(".UnloadableClass")
+			) continue; //Ignore this class
```
jkosternl, line 225:
```suggestion
			) continue; //Ignore this class because credentialProviders don't love class loading???
```
Perhaps a 'why' can be added?

jkosternl (approved):

core/src/main/java/org/frankframework/util/flow/MermaidFlowGenerator.java
tnleeuw: I don't think so. I think that if you know you do not have SAP-CD libraries on your classpath but you have that module loaded because it's just part of the bundle, you do not need to see that as a warning every time you start up the system.

jkosternl: Ok thanks. I just didn't know how this is used in practice.

nielsm5: This is just for the flow visualization, in fact, only for the colors or Frank!Elements 😄

nielsm5 (approved):

PR #6199, opened 2024-01-31
Try to determine if a datasource already does pooling

tnleeuw:
Only works for non-XA datasources currently, for XA datasources we need to still consider what is best to do as many XA drivers already implement pooling even if it's not configured.

I'm not sure about these changes, want to see if all works well in the Jenkins matrix builds. Am open to suggestions for better way to check.

### DIFF
--- a/core/src/main/java/nl/nn/adapterframework/jndi/PoolingJndiDataSourceFactory.java
+++ b/core/src/main/java/nl/nn/adapterframework/jndi/PoolingJndiDataSourceFactory.java
@@ -19,6 +19,7 @@ import java.sql.Connection;
 import java.time.Duration;
 
 import javax.sql.CommonDataSource;
+import javax.sql.ConnectionPoolDataSource;
 import javax.sql.DataSource;
 
 import org.apache.commons.dbcp2.ConnectionFactory;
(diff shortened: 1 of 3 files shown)


core/src/main/java/nl/nn/adapterframework/jndi/PoolingJndiDataSourceFactory.java
```diff
@@ -65,13 +66,23 @@ public PoolingJndiDataSourceFactory() {
...
 		}
 		log.info("Pooling not configured, using datasource [{}] without augmentation", dataSourceName);
 		return (DataSource) dataSource;
 	}
 
-	protected DataSource createPool(DataSource dataSource) {
+	private static boolean isPooledDataSource(CommonDataSource dataSource) {
+		return dataSource instanceof ConnectionPoolDataSource
+				|| dataSource.getClass().getName().startsWith("org.apache.tomcat")
+				;
```
jkosternl, line 78: At least we fix the Tomcat issues with this solution, which is the most important application server used by F!F in the wild. 

Cosmetic minor remark: this can also be put at one line?

Also something like this can be added: 
`// Datasource from Tomcat is already doing pooling` or perhaps that's already obvious.
Or a Javadoc before the method: `True if Tomcat connection pooling is used`

jkosternl (approved):

nielsm5 (approved):

nielsm5 (approved):

Issue #5491, opened 2023-10-03
timeout attribute casing is different in different senders


alisihab:
HttpSender has timeout attribute. However, it is `timeOut` in EsbJmsSender.
I think the casing should be identical for consistency

**Reporter**
Ali

PR #6203, opened 2024-02-01
Remove default values from Larva tests; use 5s timeout instead of 20s…

jkosternl:
…; improve Javadoc & code; enable ActiveMQ console

Just some small improvements

### DIFF
--- a/test/src/main/configurations/JMS/ConfigurationEsbJmsListenerSender.xml
+++ b/test/src/main/configurations/JMS/ConfigurationEsbJmsListenerSender.xml
@@ -13,9 +13,9 @@
 				className="org.frankframework.pipes.SenderPipe">
 				<sender className="org.frankframework.extensions.esb.EsbJmsSender"
 					destinationName="${tibco.rr_queue}" jmsRealm="qcf_tibco_esb_rr"
-					messageProtocol="RR">
+					messageProtocol="RR" timeOut="5000">
 					<param name="SOURCE" value="${hostname}_esbJmsRR" />
-					</sender>
+				</sender>
 				<inputWrapper
 					className="org.frankframework.extensions.esb.EsbSoapWrapperPipe"
 					mode="i2t" addOutputNamespace="true">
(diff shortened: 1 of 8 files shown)


test/src/main/configurations/JMS/ConfigurationEsbJmsListenerSender.xml
```diff
@@ -13,9 +13,9 @@
 				className="org.frankframework.pipes.SenderPipe">
 				<sender className="org.frankframework.extensions.esb.EsbJmsSender"
 					destinationName="${tibco.rr_queue}" jmsRealm="qcf_tibco_esb_rr"
-					messageProtocol="RR">
+					messageProtocol="RR" timeOut="5000">
```
tnleeuw, line 16: I hope this more limited timeout is not going to cause tests to fail when server is running slow.

Timeouts are only a problem when the reply is not going to come at all, it's not an issue when the reply will come. But a timeout that's too restrictive can be an issue when the server is not as fast for any reason.

jkosternl: The default of 20 secs is very long. The ActiveMQ is already running fine, when it reaches this point of the code, so I figured after 5 secs, the response is already given and it is already long enough. So, it indeed just helps to fail fast when we made a mistake.

tnleeuw (approved):

test/src/main/configurations/JMS/ConfigurationEsbJmsListenerSender.xml
nielsm5: Don't we have a timeout (lowercase o) property? If not, we need to sort that out soon...

jkosternl: I thought I tried with `timeout` firstly, and that failed. Than I looked it up and used `timeOut` and it works. I'll recheck.

nielsm5: I found this: https://github.com/frankframework/frankframework/issues/5491
Perhaps we can pick this up soon 😄

jkosternl: Ah.. valid point! 👍🏻

nielsm5 (approved):

PR #6210, opened 2024-02-01
Convert some more junit tests

nielsm5:


tnleeuw (approved):
Ik heb niet alles in detail bekeken maar over het algemeen ziet het er goed uit.

core/src/test/java/org/frankframework/pipes/MoveFilePipeTest.java
```diff
@@ -1,366 +1,318 @@
...
+		Files.createFile(testFolderDest.resolve("cantmove.sc"));
+		Files.createFile(testFolderDest.resolve("notcompatible.asd"));
+		Files.createFile(testFolderDest.resolve("toBeAppended.txt"));
+		Files.createFile(testFolderCantDelete.resolve("deletable.sd"));
+		Files.createFile(testFolderCantDelete.resolve("deletionInterrupter.mz"));
+		Files.createFile(testFolderDelete.resolve("moveAndDeleteDirectory.txt"));
+	}
+
+	@Test
+	public void nonExistingFileWithSourceAndTargetDirectories() throws ConfigurationException, PipeStartException, PipeRunException {
```
jkosternl, line 77: Since you've removed the `@Order` annotations, does this class now run fine in any random order? The directory setup is only performed once, and the tests don't alter the content?

core/src/test/java/org/frankframework/pipes/MoveFilePipeTest.java
```diff
@@ -1,366 +1,318 @@
...
+		PipeRunResult res = pipe.doPipe(null, session);
+
+		assertEquals(pipeForwardThen, res.getPipeForward().getName());
+	}
+
+	@Test
+	public void cantMoveFilesWithWildcardTest() throws ConfigurationException, PipeStartException, PipeRunException {
+		// exception.expect(PipeRunException.class);
+		// exception.expectMessage("no files with wildcard [*.xd] found in directory
+		// ["+sourceFolderPath+"]");
```
jkosternl, line 267: Feel free to remove these 3 `//` lines.

jkosternl (approved):
Just a few remarks. 
Question: did you also check if the amount of unittests are equal or different (but explainable)?

core/src/test/java/org/frankframework/pipes/MoveFilePipeTest.java
nielsm5: It's an annotation from `Spring`, it has nothing to do with tests, and has never worked.

core/src/test/java/org/frankframework/jdbc/JdbcTableListenerTest.java
```diff
@@ -500,12 +515,14 @@ public void testParallelChangeProcessState(boolean mainThreadFirst) throws Excep
...
 	}
 
-	@Test
+	@DatabaseTest
 	public void testParallelChangeProcessStateMainThreadFirst() throws Exception {
 		testParallelChangeProcessState(true);
 	}
 
-	@Test
+//	Niels snapt deze tests niet helemaal... Dit kan niet, de tabel doet een rowlock, zoals het hoort.
```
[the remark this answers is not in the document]
tnleeuw: In addition Tim agrees with Niels the test cannot work in the form as it is, only worked b/c of error in older test code, and making it work properly in all scenarios is non-trivial due to multi threading control issues.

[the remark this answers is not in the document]
jkosternl: Let us clean it, if we don't see the added value. It costs only time to think about it, every time we need to maintain / read it.

core/src/test/java/org/frankframework/pipes/MoveFilePipeTest.java
jkosternl: Ah.. I assumed it's from JUnit 😉 Thanks.

nielsm5:
> Just a few remarks. Question: did you also check if the amount of unittests are equal or different (but explainable)?

Yup 😄

PR #6249, opened 2024-02-07
Improve SFTP implementation and unit tests

jkosternl:
Fix directory with ending slash.
Make SftpFileSystem working for absolute directory paths as well.


core/src/test/java/org/frankframework/filesystem/FileSystemTest.java
```diff
@@ -29,6 +30,30 @@ public void setUp() throws Exception {
 		autowireByName(fileSystem);
 	}
 
+	@Test
+	@Disabled //Works in master (8.1-SNAPSHOT), but not in lower versions, because of failures in AmazonS3FileSystemTest
```
tnleeuw, line 34: This extra information is best added to the `@Disabled("...")` annotation itself so that it can be logged by the test-runner for instance.

jkosternl: Ah yes, thanks!

nielsm5 (approved):

tnleeuw (approved):

PR #6255, opened 2024-02-08
Remove maven Jetty plugin

tnleeuw:
This no longer works and we do not intend to bring it back, as it's not used in practice anyway.


jkosternl (approved):
Does it also fail in 8.0 and 7.9? If so, we should backport it too.

nielsm5 (approved):

PR #6279, opened 2024-02-14
Initiele opzet http context sharing

nielsm5:


nielsm5:
Fixed in the 8.1
Unsure if this needs to be backported, closing for now.

PR #6284, opened 2024-02-15
Don't catch all exceptions but rethrow

nielsm5:
I believe I've got it all covered (test wise) but it does add an additional minute to our test suite...


tnleeuw (approved):

core/src/test/java/org/frankframework/receivers/TestReceiverOnError.java
```diff
@@ -0,0 +1,259 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+import org.junit.jupiter.params.provider.ValueSource;
+
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
+public class TestReceiverOnError {
+	private static TestConfiguration configuration = TransactionManagerType.DATASOURCE.create();
```
jkosternl, line 59:
```suggestion
	private static final TestConfiguration configuration = TransactionManagerType.DATASOURCE.create();
```

jkosternl (approved):
Each unit test takes 10 seconds to complete 👎🏻 .. But if this is the only good way to test it, probably we should take this loss.

PR #6314, opened 2024-02-19
Remove unused testtool class

nielsm5:


core/src/test/java/org/frankframework/util/FileHandlerTestBase.java
```diff
@@ -317,6 +321,7 @@ public void testReadJsonDecode(Class<? extends IFileHandler> clazz) throws Excep
 
 	@ParameterizedTest
 	@MethodSource("data")
+	@Disabled //Doesn't seem to work properly
```
tnleeuw, line 324: You can add the reason for disabling a test directly on the annotion: `@Disabled("Test doesn't seem to work properly")` for instance.

tnleeuw (approved):

core/src/test/java/org/frankframework/util/FileHandlerTestBase.java
nielsm5: Yea, although I didn't want to find out the real reason this test isn't working properly. Something is not being closed.. But if I were to write that it would trigger more questions then answers. In this case the underlaying code has been deprecated a while ago.

PR #6318, opened 2024-02-20
Make query-timeout more consistently for all query senders

tnleeuw:


jkosternl (approved):
Looking good. Do we need to add some verification in unit tests to see if the property was really set? If that is hard, this is perhaps sufficient.

PR #6320, opened 2024-02-20
Rename jdbc minPoolSize to minIdle

tnleeuw:


jkosternl (approved):
So, `minPoolSize` is gone as a property. It needs to be mentioned in 7.9.1 release notes and migration path: rename ...minPoolSize to ...minIdle

tnleeuw:
> So, `minPoolSize` is gone as a property. It needs to be mentioned in 7.9.1 release notes and migration path: rename ...minPoolSize to ...minIdle

Good point. For Narayana I removed it but it was a documented property.

For `transactionmanager.jdbc.connection.minPoolSize` it is somewhat different since it only existed so far in a number of hotfix / nightly builds.

I've now added back some defaults to the configuration so that the old property `transactionmanager.narayana.jdbc.connection.minPoolSize` will continue to work if configured.

nielsm5 (approved):

nielsm5:
Er gaat nog iets goed fout in de CI

tnleeuw:
> Er gaat nog iets goed fout in de CI

Dat was 1 kleine vergissing in the AppConstants.properties waardoor 2 properties recursief naar elkaar verwezen 🙈

nielsm5 (approved):

PR #6338, opened 2024-02-22
Add RuntimeSenderWrapper

nielsm5:


core/src/main/java/org/frankframework/senders/RuntimeSenderWrapper.java
```diff
@@ -0,0 +1,85 @@
...
+import org.frankframework.core.PipeLineSession;
+import org.frankframework.core.SenderException;
+import org.frankframework.core.SenderResult;
+import org.frankframework.core.TimeoutException;
+import org.frankframework.statistics.HasStatistics;
+import org.frankframework.statistics.StatisticsKeeperIterationHandler;
+import org.frankframework.stream.Message;
+
+/**
+ * Wrapper for senders, that opens 'runtime' before each sender action, and closes afterwards.
```
jkosternl, line 32:
```suggestion
 * Wrapper for senders, that opens 'runtime' before each sender action, and closes afterwards. Prevents long open connections inside Senders and possible connection failures. 
```

jkosternl (approved):

core/src/main/java/org/frankframework/senders/RuntimeSenderWrapper.java
```diff
@@ -0,0 +1,85 @@
...
+import org.frankframework.core.PipeLineSession;
+import org.frankframework.core.SenderException;
+import org.frankframework.core.SenderResult;
+import org.frankframework.core.TimeoutException;
+import org.frankframework.statistics.HasStatistics;
+import org.frankframework.statistics.StatisticsKeeperIterationHandler;
+import org.frankframework.stream.Message;
+
+/**
+ * Wrapper for senders, that opens 'runtime' before each sender action, and closes afterwards. Prevents long open connections inside Senders and possible connection failures.
```
tnleeuw, line 32:
```suggestion
 * Wrapper for senders, that opens the sender in 'runtime' before each sender action, and closes it afterwards. This prevents long open connections inside Senders and possible connection failures. 
```

core/src/main/java/org/frankframework/senders/RuntimeSenderWrapper.java
```diff
@@ -0,0 +1,85 @@
...
+import org.frankframework.stream.Message;
+
+/**
+ * Wrapper for senders, that opens 'runtime' before each sender action, and closes afterwards. Prevents long open connections inside Senders and possible connection failures. 
+ *
+ * @ff.parameters any parameters defined on the SenderWrapper will be handed to the sender, if this is a {@link ISenderWithParameters}
+ *
+ * @author  Niels Meijer
+ */
+public class RuntimeSenderWrapper extends SenderWrapperBase {
```
tnleeuw, line 38: I don't like the name because it doesn't tell you what it does, but I don't have a better name yet.

- `OpenOnDemandSenderWrapper`?
- `ConnectOnDemandSenderWrapper`?
- `RuntimeOpenSenderWrapper`?
- `DynamicOpenSenderWrapper`?

core/src/main/java/org/frankframework/senders/RuntimeSenderWrapper.java
```diff
@@ -0,0 +1,85 @@
...
+		} finally {
+			getSender().close(); //Only open to test the connection, close afterwards if all is ok.
+		}
+	}
+
+	@Override
+	public SenderResult doSendMessage(Message message, PipeLineSession session) throws SenderException, TimeoutException {
+		try {
+			getSender().open();
+			return sender.sendMessage(message, session);
```
tnleeuw, line 68: Why `getSender()` on one line, and just `sender` on the next line?

```suggestion
			getSender().open();
			return getSender().sendMessage(message, session);
```

This is more consistent with the rest of the code.

tnleeuw (changes requested):

core/src/main/java/org/frankframework/senders/RuntimeSenderWrapper.java
jkosternl: Good idea.

core/src/main/java/org/frankframework/senders/RuntimeSenderWrapper.java
jkosternl: Yes, I agree with you. I like this one the best `ConnectOnDemandSenderWrapper` or what about `OpenCloseSenderWrapper`? That's exactly what happens. Perhaps let Sena validate it, since she is exactly a person who directly needs to understand it.

core/src/main/java/org/frankframework/senders/RuntimeSenderWrapper.java
```diff
@@ -0,0 +1,93 @@
...
+import org.frankframework.core.SenderResult;
+import org.frankframework.core.TimeoutException;
+import org.frankframework.statistics.HasStatistics;
+import org.frankframework.statistics.StatisticsKeeperIterationHandler;
+import org.frankframework.stream.Message;
+
+import lombok.Setter;
+
+/**
+ * Wrapper for senders, that opens 'runtime' before each sender action, and closes afterwards.
```
tnleeuw, line 31: I still think that this change makes the documentation clearer:

```suggestion
 * Wrapper for senders, that opens the wrapped sender at runtime before each sender action, and closes it afterwards.
```

tnleeuw (approved):
Well I guess the name is not going to be changed to then Appove 😁

nielsm5:
> Well I guess the name is not going to be changed to then Appove 😁

I'll quickly interview some people.. So far I'm not convinced yet.

tnleeuw:
> > Well I guess the name is not going to be changed to then Appove 😁
> 
> I'll quickly interview some people.. So far I'm not convinced yet. (At least the documentation/implementation is ok now)

I'm also not sure either of these suggestions if really _right_ but to me they're better than `RuntimeSenderWrapper`. Either way, changing or not changing will not change my approval I just wanted to hear some reaction from you on that comment.

PR #6347, opened 2024-02-23
Rename jdbc minPoolSize to minIdle

tnleeuw:
* Make name of configuration property clearer and make it more clear that BTM specific options only apply to XA datasource connection pools.
* Revert old name `transactionmanager.narayana.jdbc.connection.minPoolSize` for `transactionmanager.narayana.jdbc.connection.minIdle`.

### DIFF
--- a/core/src/main/resources/AppConstants.properties
+++ b/core/src/main/resources/AppConstants.properties
@@ -115,8 +115,12 @@ jdbc.required=true
 jdbc.datasource.default=jdbc/${instance.name.lc}
 
 ## JDBC Connection Pool properties
-# Minimum number of physical connections to maintain.
+# Deprecated, use instead `transactionmanager.jdbc.connection.minIdle`
 transactionmanager.jdbc.connection.minPoolSize=0
+# Minimum number of physical connections that should be maintained in IDLE state when the pool is not full.
+# Keeping idle connections available in the pool can increase performance.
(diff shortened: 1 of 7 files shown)


nielsm5 (approved):

core/src/main/resources/AppConstants.properties
```diff
@@ -115,8 +115,12 @@ jdbc.required=true
 jdbc.datasource.default=jdbc/${instance.name.lc}
 
 ## JDBC Connection Pool properties
-# Minimum number of physical connections to maintain.
+# Deprecated, use instead `transactionmanager.jdbc.connection.minIdle`
```
jkosternl, line 118: Shouldn't we remove the deprecated option from 8.1 because we already marked it as deprecated in 7.9/8.0 and people are still going to migrate to that version still?

jkosternl (changes requested):
Question asked about deprecation.

core/src/main/resources/AppConstants.properties
nielsm5: Regardless when people migrate to newer versions, it's deprecated in this one, thus the tag should be there?

jkosternl: Ok, we'll leave it for 8.1 like this.

jkosternl (approved):

PR #6371, opened 2024-02-27
Remove unnecessary information from GrayBox view

MarkvdVorst:
Created AbstractBox for parent of GreyBox and BlackBox so that the BlackBox does not depend on GreyBox for better readability.

Created a fix for the greybox problem by looking in front and behind the info-, input-, or ouputpoint checkpoint to see if there is another startpoint on the same level in the tree structure. if true, don't show the checkpoint.


MarkvdVorst:
@jacodg

ladybug/src/main/java/org/frankframework/ibistesttool/filter/GrayBox.java
```diff
@@ -20,20 +20,22 @@
...
-public class GrayBox implements CheckpointMatcher {
-
+public class GrayBox extends AbstractBox {
+	@Override
 	public boolean match(Report report, Checkpoint checkpoint) {
 		if (checkpoint.getType() == Checkpoint.TYPE_INPUTPOINT || checkpoint.getType() == Checkpoint.TYPE_OUTPUTPOINT
 				|| checkpoint.getType() == Checkpoint.TYPE_INFOPOINT) {
 			List<Checkpoint> checkpoints = report.getCheckpoints();
 			ListIterator<Checkpoint> iterator = report.getCheckpoints().listIterator(checkpoints.indexOf(checkpoint));
+			if (hasStartPointOnLevel(report, checkpoint)) {
```
jacodg, line 36: This if can be moved two lines up so it is clear that this if doesn't use checkpoints and iterator.
Or maybe pass checkpoints and iterator to hasStartPointOnLevel() so that method doesn't have to reinvent the wheel but then we have to be sure the iterator is returned at the right position.

jacodg (commented):
The title of the PR could be more descriptive (what does it exactly fix?)

ladybug/src/main/java/org/frankframework/ibistesttool/filter/GrayBox.java
MarkvdVorst: I moved the if-statement up and changed the title of the PR

jacodg (approved):

nielsm5 (commented):
I find it difficult to judge/see what's changed.
Have you also considered creating some Unit Tests for this?

jkosternl (commented):
Looking good at first sight. Adding some unit tests should give more security that it is working fine.

ladybug/src/main/java/org/frankframework/ibistesttool/filter/GrayBox.java
```diff
@@ -46,31 +48,25 @@ public boolean match(Report report, Checkpoint checkpoint) {
...
 			}
-			Checkpoint lastCheckpoint = checkpoints.get(checkpoints.size() - 1);
-			if (checkpoint.equals(lastCheckpoint)) {
+		}
+		while (iterator.hasPrevious()) {
+			Checkpoint previousCheckpoint = iterator.previous();
+			if (previousCheckpoint.getLevel() < currentLevel) {
+				break;
+			}
+			if (previousCheckpoint.getType() == Checkpoint.TYPE_STARTPOINT) {
```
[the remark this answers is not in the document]
MarkvdVorst: in this section it is looking if there is a startpoint checkpoint behind the current checkpoint on the same level. on line 60 it just looks in front. if both sides pass we can confirm that there is no startpoint on the same level in the tree and that the checkpoint can be shown.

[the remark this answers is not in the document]
jkosternl: Ah clear, so it should stay like this. Hopefully you can add a unit test to validate this behavior.

MarkvdVorst:
Added unit tests for BlackBox, GrayBox and WhiteBox

Added a subclass for XmlStorage to get access to the readReportFromFile(File file) method for converting the test report xml file to a Report object.

WhiteBox is the control group to see if the report xml file is correctly imported as a Report object.

ladybug/src/test/java/org/frankframework/ibistesttool/filter/WhiteBoxTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+	@Test
+	public void shouldReturnAllCheckpoints() throws StorageException {
+		File reportFile = new File(Objects.requireNonNull(this.getClass().getClassLoader().getResource("testReport/Pipeline_HelloWorlds.report.xml")).getFile());
+		Report report = xmlStorage.readReportFromFile(reportFile);
+		List<Checkpoint> reportCheckpoints = report.getCheckpoints();
+		for (int index = 0; index < reportCheckpoints.size(); index++) {
+			assertEquals(expectedResult[index], reportCheckpoints.get(index).getUid());
+		}
+	}
```
jkosternl, line 31: Which class do you test with this unit test? Is this not already tested with the two the other unit test files? 

Secondly, can this class be merged with e.g. BlackBoxTest or GrayBoxTest?

ladybug/src/test/java/org/frankframework/ibistesttool/filter/GrayBoxTest.java
```diff
@@ -0,0 +1,37 @@
...
+		List<Checkpoint> checkpointsAfterFilter = new ArrayList<>();
+		for (Checkpoint checkpoint : report.getCheckpoints()) {
+			if (grayBox.match(report, checkpoint)) {
+				checkpointsAfterFilter.add(checkpoint);
+			}
+		}
+
+		for (int index = 0; index < checkpointsAfterFilter.size(); index++) {
+			assertEquals(expectedResult[index], checkpointsAfterFilter.get(index).getUid());
+		}
```
jkosternl, line 35: If `checkpointsAfterFilter` is empty, the implementation fails, but your test thinks it is fine, because the whole for-loop is not executed at all 😉 . Please add something like `assertEquals(5, checkpointsAfterFilter.size());` before the loop, to validate the list contents makes sense. 

Same for `BlackBoxTest.java`

jkosternl (changes requested):

ladybug/src/test/java/org/frankframework/ibistesttool/filter/WhiteBoxTest.java
MarkvdVorst: This is more to test the default view in ladybugfrontend so you can see if the checkpoints are correctly shown, but I now realise like you said that isn't really a necessary thing to do since the other tests would fail in that case. I'll remove it.

ladybug/src/test/java/org/frankframework/ibistesttool/filter/GrayBoxTest.java
MarkvdVorst: Thank you for the feedback. I forgot about that case. I have added the changes in the latest commit

jkosternl (approved):

nielsm5 (approved):

PR #6376, opened 2024-02-28
Add default property for forceMessageIdAsCorrelationId on RR EsbJmsListeners

jkosternl:
Also changed some Javadoc default to `@Default`, which is better.

### DIFF
--- a/nn-specials/src/main/java/org/frankframework/extensions/esb/EsbJmsListener.java
+++ b/nn-specials/src/main/java/org/frankframework/extensions/esb/EsbJmsListener.java
@@ -29,6 +29,7 @@ import javax.jms.JMSException;
 import javax.jms.Message;
 import javax.jms.TextMessage;
 
+import org.apache.commons.lang3.StringUtils;
 import org.frankframework.configuration.ConfigurationException;
 import org.frankframework.configuration.ConfigurationWarnings;
 import org.frankframework.configuration.SuppressKeys;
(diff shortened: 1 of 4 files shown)


tnleeuw:
Why is `@Default` better than `@ff.default`?

jkosternl:
> Why is `@Default` better than `@ff.default`?

`@ff.default` only works for FrankDoc. And `@Default` for both Javadoc and FrankDoc, as far as I can remember.

nn-specials/src/main/java/org/frankframework/extensions/esb/EsbJmsListener.java
```diff
@@ -62,6 +63,8 @@ public class EsbJmsListener extends JmsListener implements ITransactionRequireme
 
 	private static final AppConstants APP_CONSTANTS = AppConstants.getInstance();
 	private static final String MSGLOG_KEYS = APP_CONSTANTS.getProperty("msg.log.keys");
+	static final String JMS_RR_FORCE_MESSAGE_KEY = "jms.rr.forceMessageIdAsCorrelationId.default";
```
tnleeuw, line 66: Since this is EsbJms specific, perhaps the propertyname prefix should reflect this? `jms.esb.*` or `esbjms.*` or whatever.

tnleeuw (approved):
1 small suggestion.

nn-specials/src/main/java/org/frankframework/extensions/esb/EsbJmsListener.java
jkosternl: Done!

tnleeuw (approved):

nielsm5 (approved):

PR #6379, opened 2024-02-28
Console Frontend linting & code formatting

Matthbo:
Adds ESLint & Prettier to help format & lint while developing


Matthbo:
Every page needs to be checked for errors since not all errors occur during the build process, sadly

philipsens (approved):
There are still some bits of code that could be improved. But that's for another time, when refactoring code.

PR #6389, opened 2024-03-01
Use updated @FrankDocGroup

mhdirkse:


nielsm5:
If you could update the version to 3.1-SNAPSHOT, we can merge this one 😄

jkosternl (approved):
Tested by comparing the JS and XSD output. Seems to be the same, where expected.

PR #6408, opened 2024-03-05
Fix Javadoc generation for Java 21

jkosternl:
Real fix was specifying the version of Lombok to the plugin.

Also removed the `onMethod`, which we need to discuss in a meeting, if we should keep it or not. 
There are 2 options to write those:
up to JDK7: `@Getter(onMethod=@__({@AnnotationsGoHere}))`
from JDK8: `@Getter(onMethod_={@AnnotationsGohere}) // note the underscore after onMethod.`

Fixed some side-findings: vintage-engine, missing `groupId` for maven-compiler-plugin.

### DIFF
--- a/core/src/main/java/org/frankframework/validation/XSD.java
+++ b/core/src/main/java/org/frankframework/validation/XSD.java
@@ -32,7 +32,6 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import javax.annotation.Nullable;
 import javax.xml.stream.XMLEventReader;
 import javax.xml.stream.XMLStreamConstants;
 import javax.xml.stream.XMLStreamException;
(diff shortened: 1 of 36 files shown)


core/src/main/java/org/frankframework/validation/XSD.java
```diff
@@ -83,7 +82,7 @@ public abstract class XSD implements IXSD, Comparable<XSD> {
 	private final @Getter Set<String> importedNamespaces = new HashSet<>();
 	private @Getter String xsdTargetNamespace;
 	private @Getter String xsdDefaultNamespace;
-	private @Getter(onMethod_ = {@Nullable, @Override}) @Setter IXSD importParent;
```
nielsm5, line 86: Dit was de enige nuttige annotatie, misschien deze getter wel uitschrijven?
Voor de rest zie ik alleen maar positieve changes

nielsm5 (approved):

core/src/main/java/org/frankframework/validation/XSD.java
jkosternl: Prima, zal ik weer toevoegen.

tnleeuw (approved):

nielsm5 (approved):

PR #6430, opened 2024-03-12
Close Sender later, when PipeLineSession is closed

jkosternl:


core/src/main/java/org/frankframework/senders/ReconnectSenderWrapper.java
```diff
@@ -65,20 +65,36 @@ public void open() throws SenderException {
 			sender.open();
 			super.open();
 		} finally {
+			super.close();
```
nielsm5, line 68: It should only close the sender, not it self.

core/src/main/java/org/frankframework/senders/ReconnectSenderWrapper.java
```diff
@@ -65,20 +65,36 @@ public void open() throws SenderException {
...
 	@Override
 	public SenderResult doSendMessage(Message message, PipeLineSession session) throws SenderException, TimeoutException {
-		try {
-			sender.open();
-			return sender.sendMessage(message, session);
-		} finally {
-			sender.close();
-		}
+		sender.open();
+		session.scheduleCloseOnSessionExit(new AutoCloseableSenderWrapper(sender), ReconnectSenderWrapper.class.getSimpleName());
```
nielsm5, line 76:
```suggestion
		session.scheduleCloseOnSessionExit(new AutoCloseableSenderWrapper(sender), this);
```
The sole purpose of the requester argument is for traceability in the logging. It might be best to just print the entire object (which includes the hashCode).

core/src/test/java/org/frankframework/senders/ReconnectSenderWrapperTest.java
```diff
@@ -32,39 +36,57 @@ public void basic() throws Exception {
...
+		sender.setSender(senderMock);
+		sender.configure();
+		sender.open();
+
+		// Act
+		sender.sendMessageOrThrow(Message.nullMessage(), session).asString();
+
+		// Assert
+		verify(senderMock, Mockito.times(2)).open();
+		verify(senderMock, Mockito.times(1)).close();
```
nielsm5, line 53: Perhaps it's an idea to close the session, and verify that close is called once more.

nielsm5 (changes requested):

core/src/main/java/org/frankframework/senders/ReconnectSenderWrapper.java
jkosternl: Yes, you're right. I'll remove it.

core/src/main/java/org/frankframework/senders/ReconnectSenderWrapper.java
jkosternl: This suggestion does not compile, since the method only accepts a String. I'll fix it.

core/src/test/java/org/frankframework/senders/ReconnectSenderWrapperTest.java
jkosternl: Yes indeed, I've improved it with your idea.

nielsm5 (approved):

PR #6435, opened 2024-03-13
SFTP Unit testing: Test if port is free, before using it for unit test

jkosternl:
* Also clean up code and reuse setting up SFTP server

Prevents issues on GitHub builds with port is used

### DIFF
--- a/filesystem/src/test/java/org/frankframework/filesystem/SftpFileSystemTest.java
+++ b/filesystem/src/test/java/org/frankframework/filesystem/SftpFileSystemTest.java
@@ -6,15 +6,10 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.io.File;
 import java.io.IOException;
-import java.util.Collections;
-import java.util.random.RandomGenerator;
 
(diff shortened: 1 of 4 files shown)


filesystem/src/test/java/org/frankframework/filesystem/SftpFileSystemTest.java
```diff
@@ -46,28 +41,14 @@ private void startNewSshDaemon() throws IOException {
 		if("localhost".equals(host)) {
 			remoteDirectory = "/"; // See getTestDirectoryFS(), '/' is the SFTP HOME directory.
 
-			sshd = createSshServer(username, password);
+			sshd = SftpFileSystemTestHelper.createSshServer(username, password, port);
+			sshd.setFileSystemFactory(getTestDirectoryFS());
```
nielsm5, line 45: Deze zit nu in de helper toch?

jkosternl: Ja oeps, die had ik nog willen opruimen. Done. Thanks.

PR #6439, opened 2024-03-13
Unit tests for pipes, directives & utility functions

Matthbo:


console/frontend/src/main/frontend/src/app/components/pages/side-navigation.directive.spec.ts
```diff
@@ -1,8 +1,49 @@
...
+      declarations: [TestComponent, SideNavigationDirective],
+    }).createComponent(TestComponent);
+
+    fixture.detectChanges(); // initial binding
+  });
+
+  // TODO figure out a way to test metismenu
+  // https://angular.io/guide/testing-attribute-directives
+
+  // temporatory, remove when making actual tests
```
philipsens, line 45: When will this be done?

Matthbo: I changed my mind in trying to unit test this.
The [side-navigation directive](https://github.com/frankframework/frankframework/blob/issue/6336-console-write-unit-tests-for-pipes-directives-utility-functions/console/frontend/src/main/frontend/src/app/components/pages/side-navigation.directive.ts) is only a call to the start the metismenu jQuery plugin.
Its not worth it to unit test this because its not completely our responsibility that their code works but also metismenu will have to be replaced in the future because jQuery is going to be removed hopefully sooner rather than later

philipsens (approved):

PR #6446, opened 2024-03-14
Fix log4j2 configuration

nielsm5:
- Change package scan to an annotation processor


core/pom.xml
```diff
@@ -897,7 +896,7 @@
 						<phase>process-classes</phase>
 						<configuration>
 							<proc>only</proc>
-							<compileSourceRoots>${project.build.directory}/generated-sources/delombok</compileSourceRoots>
+							<compileSourceRoots>${project.build.outputDirectory}</compileSourceRoots>
```
jkosternl, line 899: Ah, did the output of this plugin get overwritten by the delombok step afterwards?

core/src/main/java/org/frankframework/logging/FrankLogConfigurationFactory.java
```diff
@@ -33,28 +34,32 @@
...
-import org.frankframework.util.StreamUtil;
 import org.frankframework.util.StringResolver;
 
 /**
- * This ConfigurationFactory is loaded after the log4j2.properties file has been initialised.
+ * This ConfigurationFactory is loaded after the log4j2.properties file has been initialized.
  * Both Configurations are then combined via a CompositeConfiguration
+ * 
+ * NOTE:
+ * The use of Lombok is not allowed!
```
jkosternl, line 53: In this class or in this package? Can you also add a why, in the line?

core/src/main/java/org/frankframework/management/bus/endpoints/SecurityItems.java
```diff
@@ -227,8 +227,8 @@ private ArrayList<Object> addSapSystems() {
 
 			sapSystems = (List) factoryGetRegisteredSapSystemsNamesAsList.invoke(sapSystemFactory, (Object[])null);
 			factoryGetSapSystemInfo = c.getMethod("getSapSystemInfo", String.class);
-		} catch (Exception e) {
-			log.debug("Caught NoClassDefFoundError, just no sapSystem available: " + e.getMessage());
+		} catch (Throwable t) { //Don't change this throwable, AI doesn't know better...
```
jkosternl, line 230: Instead of blaming AI, I think a `@SuppressWarning` would be better, because static code analysis tools will complain about it, for sure.

core/src/main/resources/log4j4ibis.xml
```diff
@@ -123,7 +127,7 @@
 		<Logger name="nl.nn.testtool" additivity="false" level="INFO">
 			<AppenderRef ref="ladybug"/>
 		</Logger>
-		<Logger name="nl.nn.ibistesttool" additivity="false" level="INFO">
+		<Logger name="org.frankframework.ibistesttool" additivity="false" level="INFO">
```
jkosternl, line 130: Previously, I tried to change this as well, in a PR. But I got rejected by other members of the Core team 😉 
The argument was that these packages are targeted at the external `Ladybug`.

ladybug/src/main/resources/application.properties
```diff
@@ -1,11 +1 @@
...
-logging.level.root=DEBUG
-logging.level.org.frankframework=DEBUG
-logging.level.org.apache.cxf=WARN
-logging.level.org.apache.coyote=WARN
-logging.level.org.apache.tomcat=WARN
-logging.level.org.apache.catalina=INFO
-logging.level.org.apache.jasper=WARN
-
-
 spring.main.allow-bean-definition-overriding=true
```
jkosternl, line 1: We can't move the last property to XML configuration, so we don't need this whole file anymore?

pom.xml
```diff
@@ -351,76 +361,28 @@
...
-			</dependency>
-			<dependency>
-				<groupId>org.springframework</groupId>
-				<artifactId>spring-aop</artifactId>
-				<version>${spring.version}</version>
-			</dependency>
-			<dependency>
-				<groupId>org.springframework</groupId>
-				<artifactId>spring-expression</artifactId>
+				<artifactId>spring-framework-bom</artifactId>
```
jkosternl, line 373: Ah... good catch. 👍🏻  
I didn't see it before.

pom.xml
```diff
@@ -432,6 +394,15 @@
...
 			</dependency>
 
+			<!-- import BOM to lock all versions -->
+			<dependency>
+				<groupId>org.apache.logging.log4j</groupId>
+				<artifactId>log4j-bom</artifactId>
+				<version>${log4j2.version}</version>
+				<type>pom</type>
+				<scope>import</scope>
+			</dependency>
```
jkosternl, line 404: Can we now remove the 7 mentioned imports of this groupId as well? Or are they not included in this bom?

core/src/main/java/org/frankframework/logging/IbisNdcPatternConverter.java
```diff
@@ -28,7 +28,7 @@
 public class IbisNdcPatternConverter extends LogEventPatternConverter {
 
 	private IbisNdcPatternConverter() {
-		super("IbisNdcPatternConverter", "ndc");
+		super("IbisNdcPatternConverter", "TC");
```
jkosternl, line 31: I don't know what the difference between 'ndc' and 'TC' is. Can you perhaps add a comment on this line on TC?

core/pom.xml
nielsm5: No, this directory only exists when running maven with the delombok step. Normal class info can be used here, but both paths could be added if needed.

core/src/main/java/org/frankframework/lifecycle/IbisApplicationContext.java
```diff
@@ -248,6 +248,7 @@ private void lookupApplicationModules() {
 		modulesToScanFor.add("frankframework-batch");
 		modulesToScanFor.add("frankframework-cmis");
 		modulesToScanFor.add("frankframework-commons");
+		modulesToScanFor.add("frankframework-filesystem");
```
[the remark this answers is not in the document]
nielsm5: Only DBMS and security seem to be missing.

core/src/main/java/org/frankframework/logging/IbisNdcPatternConverter.java
nielsm5: A self made up shorthand for `PatternConverter`.

core/src/main/resources/log4j4ibis.xml
nielsm5: No, nl.nn.testtool is.

ladybug/src/main/resources/application.properties
nielsm5: No, these are spring properties, they have nothing to do with the Frank.

pom.xml
nielsm5: Yes we can!

ladybug/src/main/resources/application.properties
jkosternl: Ok, thanks.

jkosternl (approved):

PR #6467, opened 2024-03-18
Add all sender AOP to ladybug

nielsm5:


ladybug/src/main/resources/springIbisDebuggerAdvice.xml
```diff
@@ -104,6 +104,22 @@
...
 					"
+				method="debugSenderWithForwardInputOutputAbort"
+				arg-names="message, session"
+			/>
+			<aop:around
+				pointcut=
+					"
+					execution(
+						*
+						org.frankframework.core.ISender.sendMessageOrThrow(
```
jkosternl, line 115: So, we've missed this method in the Ladybug debugger advice, for a long time?

jkosternl (approved):

ladybug/src/main/resources/springIbisDebuggerAdvice.xml
nielsm5: Yea, I think since we moved from/to OutputSteaming support.

jkosternl (approved):

PR #6474, opened 2024-03-19
Fix Ladybug doesn't wait for database connection

jacodg:


jacodg:
Fix for https://github.com/wearefrank/ladybug/issues/223

PR #6481, opened 2024-03-20
Remove '/*' from single endpoint ApiListener in physicalDestinationName

jkosternl:
And also clean some 'throws Exceptions'.
I'll add a comment to the real fix.


core/src/main/java/org/frankframework/http/rest/ApiListener.java
```diff
@@ -187,7 +187,8 @@ private void buildPhysicalDestinationName() {
 			ServletConfiguration config = servletManager.getServlet(ApiListenerServlet.class.getSimpleName());
 			if (config != null) {
 				if (config.getUrlMapping().size() == 1) {
-					builder.append(config.getUrlMapping().get(0));
+					// Replace /api/* with /api. Doing this in ApiListenerServlet.getUrlMapping() is not possible.
+					builder.append(config.getUrlMapping().get(0).replace("/*", ""));
```
jkosternl, line 191: The only fix in this PR.

nielsm5 (approved):

PR #6484, opened 2024-03-20
Add ObfuscatePipe to Tibco module

jkosternl:
* Decryption does not always work, so disabled 2 unit tests

### DIFF
--- a/tibco/pom.xml
+++ b/tibco/pom.xml
@@ -71,6 +71,7 @@
 
 	<build>
 		<plugins>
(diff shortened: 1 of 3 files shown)


nielsm5 (approved):

nielsm5:
> * Decryption does not always work, so disabled 2 unit tests

Or perhaps the encryption doesn't always work 😄

nielsm5:
Supersedes #6021

PR #6494, opened 2024-03-22
Remove `reuseForks` to fix surefire forks crashing during build 

philipsens:


nielsm5:
I thought the default was [true](https://github.com/apache/maven-surefire/blob/bae016d754f3ed4066d88f15eed869d03d7bcebc/maven-surefire-common/src/main/java/org/apache/maven/plugin/surefire/AbstractSurefireMojo.java#L532)?

nielsm5:
I've removed `T 1.5C` and replaced it with `T 6`. Seems to be stable now.

PR #6513, opened 2024-03-25
Show toast on adapter reload

Matthbo:
Shows a toast when the `upSince` of one or multiple adapters have changed
image

Also a fix for manual config reload where poller in `startPollingForConfigurationStateChanges` didn't stop


console/frontend/src/main/frontend/src/app/app.component.ts
```diff
@@ -565,6 +578,15 @@ export class AppComponent implements OnInit, OnDestroy {
     }
   }
 
+  hasAdapterReloaded(adapter: Adapter): boolean {
+    const oldAdapter =
+      this.appService.adapters[`${adapter.configuration}/${adapter.name}`];
+    if (adapter.upSince > oldAdapter?.upSince) {
```
philipsens, line 584: It's possible to return this condition instead of true or false?

Matthbo: You're correct, it was a void function before with a loop for all adapters.
Didn't think it more through after I changed it to this

philipsens (approved):

PR #6514, opened 2024-03-26
Revert "Bump org.apache.activemq:artemis-jms-client-all from 2.32.0 to 2.33.0"

nielsm5:
Reverts frankframework/frankframework#6509


jkosternl (approved):
Should we lock it too?

PR #6518, opened 2024-03-26
Make iaf-test (Spring) bootable

nielsm5:
I move a lot around to clean up the code as well as use it independently.
Spring automatically finds the `WebApplicationInitializer`s when running in a traditional WAR environment.
When using Spring Boot these implementations are ignored (!).

They will need to be either manually added as `PrimarySource` on the `SpringApplication` or defined in a bean config somewhere. Spring Boot listens to the `ServletContextInitializer` interface instead, which a traditional WAR environment does not find. At some point we need to make our 'runnables' a bit more generic so they can implement both interfaces, for now I've opted to not do that due to them being in the spring-boot-starter dependency, which is not part of our default classpath.

### DIFF
--- a/ladybug/src/main/java/org/frankframework/ibistesttool/runner/TestToolWarInitializer.java
+++ b/ladybug/src/main/java/org/frankframework/ibistesttool/runner/TestToolWarInitializer.java
@@ -40,7 +40,7 @@ import org.springframework.web.context.WebApplicationContext;
  * @author Niels Meijer
  */
 @Order(500)
-public class WarInitializer extends SpringBootServletInitializer {
+public class TestToolWarInitializer extends SpringBootServletInitializer {
 	private static final Logger APPLICATION_LOG = LogUtil.getLogger("APPLICATION");
 
 	@Configuration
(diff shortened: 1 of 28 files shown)


ladybug/src/main/java/org/frankframework/ibistesttool/runner/TestToolWarInitializer.java
tnleeuw, line 1: Is deze WarInitializer voor de Larva of de Ladybug test tool?
Ik kan het zien aan het project maar meestal zie ik alleen de class naam.

Ik denk dat het duidelijker is om LadyBug expliciet in de naam te zetten, ipv. "TestTool".

tnleeuw (approved):

test/pom.xml
```diff
@@ -235,13 +269,18 @@
 		In newer versions of Eclipse (at least version 20181214-0600 onwards) this seems to have been resolved.
 		-->
 		<profile>
-			<id>m2e</id>
+			<id>spring-boot</id>
```
jkosternl, line 272: Can the large comment block already be removed, line 259 - 270? Or in a later PR. 
It seems to be quite outdated.

jkosternl (approved):

PR #6531, opened 2024-03-27
Fix elementToMove preserve

nielsm5:
Add test to validate if elementToMove works correctly, and if the original message is stored in the ErrorStorage.

### DIFF
--- a/core/src/main/java/org/frankframework/receivers/Receiver.java
+++ b/core/src/main/java/org/frankframework/receivers/Receiver.java
@@ -1186,15 +1186,12 @@ public class Receiver<M> extends TransactionAttributes implements IManagable, IM
 			log.debug("{} received message with messageId [{}] correlationId [{}]", logPrefix, messageWrapper.getId(), messageWrapper.getCorrelationId());
 
 			String messageId = ensureMessageIdNotEmpty(messageWrapper.getId());
-			Message message = compactMessageIfRequired(messageWrapper.getMessage(), session);
-
 			final String businessCorrelationId = getBusinessCorrelationId(messageWrapper, messageId, session);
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/frankframework/receivers/Receiver.java
```diff
@@ -1186,15 +1186,12 @@ private Message processMessageInAdapter(MessageWrapper<M> messageWrapper, PipeLi
 			log.debug("{} received message with messageId [{}] correlationId [{}]", logPrefix, messageWrapper.getId(), messageWrapper.getCorrelationId());
 
 			String messageId = ensureMessageIdNotEmpty(messageWrapper.getId());
-			Message message = compactMessageIfRequired(messageWrapper.getMessage(), session);
-
 			final String businessCorrelationId = getBusinessCorrelationId(messageWrapper, messageId, session);
 			session.put(PipeLineSession.CORRELATION_ID_KEY, businessCorrelationId);
 
-			MessageWrapper<M> businessMessage = new MessageWrapper<>(messageWrapper, message, messageId, businessCorrelationId);
+			MessageWrapper<M> messageWithMessageIdAndCorrelationId = new MessageWrapper<>(messageWrapper, messageWrapper.getMessage(), messageId, businessCorrelationId);
```
nielsm5, line 1192: This is not the best solution, but I've added tests to verify it's functionality so the next time we refactor this we won't bump into the same problems.
The new wrapper is solely used for the MessageLog and ErrorStorage. The 'compactedMessage' is solely for the pipeline and thus I've moved it down to make this more obvious.

tnleeuw (approved):

tnleeuw (approved):

PR #6538, opened 2024-03-28
Only AOP when ladybug module is present

nielsm5:


jkosternl (approved):
Verify good idea, to enable it only when needed.

PR #6545, opened 2024-03-29
Add a role overview table

philipsens:
Added a table containing API actions and which role is able to do it.

I'm wondering if the name and placement of the table is correct, and if actions without description should be shown or not. If not, should the action name be kept or is the description enough?

I did not implement a way to highlight the current role.

Preview:
image


Matthbo (approved):
Je bent een topper!

nielsm5 (approved):
Ziet er fantastisch uit!

PR #6558, opened 2024-04-02
Do not add the Content-Type header when it already exists

jkosternl:


core/src/main/java/org/frankframework/http/authentication/OAuthAccessTokenManager.java
```diff
@@ -184,7 +184,10 @@ protected HttpRequestBase convertToApacheHttpRequest(HTTPRequest httpRequest) th
 				break;
 			case POST:
 				apacheHttpRequest = new HttpPost(httpRequest.getURL().toExternalForm());
-				apacheHttpRequest.addHeader("Content-Type", "application/x-www-form-urlencoded;charset=UTF-8");
+				// add Content-Type if not present in the incoming request
+				if (httpRequest.getHeaderMap().keySet().stream().noneMatch(h -> h.equalsIgnoreCase("Content-Type"))) {
+					apacheHttpRequest.addHeader("Content-Type", "application/x-www-form-urlencoded;charset=UTF-8");
+				}
```
tnleeuw, line 190: If the header already exists, should it perhaps be replaced with new header? (depending on what old/new value is).

jkosternl: I am diving deeper into the code to answer your question and it seems that I fixed the unit tests, but the production code can never add headers to the original `httpRequest`, because it can only be created by `TokenRequest request = createRequest(credentials);` at method `retrieveAccessToken(credentials)`. 
So, probably the scenario's in the unit test are not like in production code. I'll take a look into the 7.9 branch now, to verify it perhaps can happen there.

jkosternl: The fix should be fine, like this. Changed `addHeader` to `setHeader` to be sure that in the future, no duplicate Content-Type-headers can exist.

tnleeuw (approved):

PR #6561, opened 2024-04-02
Add ladybug to iaf-test runner and fix add-opens classloading problem

nielsm5:


tnleeuw (approved):

nielsm5:
> A great find! Good that it fixes my issue. The frank-runner changes can perhaps be reverted then.

We should probably leave those in place for people working on snapshot versions ;)

jkosternl:
> > A great find! Good that it fixes my issue. The frank-runner changes can perhaps be reverted then.
> 
> We should probably leave those in place for people working on snapshot versions ;)

Yeah, can remove it later as well.

PR #6564, opened 2024-04-02
Fix browser not going back with back action on logging page

philipsens:
It is still backwards compatible with the old URL's

### DIFF
--- a/console/frontend/src/main/frontend/src/app/app-routing.module.ts
+++ b/console/frontend/src/main/frontend/src/app/app-routing.module.ts
@@ -142,6 +142,22 @@ const routes: Routes = [
     },
   },
   {
+    path: 'logging/:directory',
+    component: LoggingComponent,
+    title: 'Logging',
+    data: {
+      breadcrumbs: 'Logging > Log Files',
+    },
+  },
+  {
+    path: 'logging/:directory/:file',
+    component: LoggingComponent,
+    title: 'Logging',
+    data: {
+      breadcrumbs: 'Logging > Log Files',
+    },
+  },
+  {
     path: 'jms/send-message',
     component: JmsSendMessageComponent,
     title: 'Send JMS Message',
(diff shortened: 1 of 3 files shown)


console/frontend/src/main/frontend/src/app/app-routing.module.ts
```diff
@@ -141,6 +141,22 @@ const routes: Routes = [
       breadcrumbs: 'Logging > Log Settings',
     },
   },
+  {
+    path: 'logging/:directory',
+    component: LoggingComponent,
+    title: 'Logging',
+    data: {
+      breadcrumbs: 'Logging > Log Files',
```
nielsm5, line 149:
```suggestion
      breadcrumbs: 'Logging > Show Directory (/xyz)',
```

console/frontend/src/main/frontend/src/app/app-routing.module.ts
```diff
@@ -141,6 +141,22 @@ const routes: Routes = [
...
+    },
+  },
+  {
+    path: 'logging/:directory/:file',
+    component: LoggingComponent,
+    title: 'Logging',
+    data: {
+      breadcrumbs: 'Logging > Log Files',
```
nielsm5, line 157:
```suggestion
      breadcrumbs: 'Logging > Show Directory (/name) > Show File (name.png)',
```

nielsm5 (changes requested):

console/frontend/src/main/frontend/src/app/app-routing.module.ts
Matthbo: This should be possible by using `AppService.customBreadcrumbs(breadcrumbs: string): void` in the initialisation of the component, this will overwrite the current breadcrumbs until the user switches to a different page or the function is called again.

console/frontend/src/main/frontend/src/app/app-routing.module.ts
Matthbo: This should be possible by using `AppService.customBreadcrumbs(breadcrumbs: string): void` in the initialisation of the component, this will overwrite the current breadcrumbs until the user switches to a different page or the function is called again.

philipsens: Thank you for this recommendation. I've implemented it.

Matthbo (approved):

nielsm5 (approved):

PR #6576, opened 2024-04-03
Remove CleanupOldFilesPipe and UploadFilePipe

tnleeuw:


tnleeuw:
> You'll need to remove the "MoveFilePipe" adapter as well I think?

I thought we would not yet remove the `MoveFilePipe` in 8.1? So I did not remove that adapter, but fixed it. (Some tests in IAF-Test would break otherwise).

jkosternl (approved):

nielsm5:
> > You'll need to remove the "MoveFilePipe" adapter as well I think?
> 
> I thought we would not yet remove the `MoveFilePipe` in 8.1? So I did not remove that adapter, but fixed it. (Some tests in IAF-Test would break otherwise).

The adapter / test in iaf-test. not the code :)

tnleeuw:
> > > You'll need to remove the "MoveFilePipe" adapter as well I think?
> > 
> > 
> > I thought we would not yet remove the `MoveFilePipe` in 8.1? So I did not remove that adapter, but fixed it. (Some tests in IAF-Test would break otherwise).
> 
> The adapter / test in iaf-test. not the code :)

And that's the adapter which I fixed in the PR, isn't it?

PR #6580, opened 2024-04-03
Do not add the Content-Type header when it already exists (v2)

nielsm5:
I don't know where to start..
In #2683 I warned about the HttpEntity being wrong, using a StringEntity was not the wrong choice here, but the failure of an attempt to set the content-type was. In #3980 tests were created, which actually prove it's 'faultiness', because the test returns a `text/plain; charset=ISO-8859-1` content-type!

In #3980 we can also see what was intended with the `authenticatedTokenRequest` attribute; The idea being that it uses the provided clientId and secret as a header opposed to (post url-encoded) form-data fields.
In #2597 the preAuthentication methods where 'reorganized' but the beforementioned #3980 severely changed this behavior.

### DIFF
--- a/core/src/main/java/org/frankframework/http/HttpSessionBase.java
+++ b/core/src/main/java/org/frankframework/http/HttpSessionBase.java
@@ -78,6 +78,7 @@ import org.frankframework.http.authentication.HttpAuthenticationException;
 import org.frankframework.http.authentication.OAuthAccessTokenManager;
 import org.frankframework.http.authentication.OAuthAuthenticationScheme;
 import org.frankframework.http.authentication.OAuthPreferringAuthenticationStrategy;
+import org.frankframework.http.authentication.OAuthAccessTokenManager.AuthenticationType;
 import org.frankframework.lifecycle.ConfigurableLifecycle;
 import org.frankframework.util.ClassUtils;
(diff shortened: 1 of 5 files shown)


nielsm5:
Supersedes #6558

jkosternl (approved):

core/src/main/java/org/frankframework/http/HttpSessionBase.java
```diff
@@ -362,7 +364,7 @@ public void configureConnectionManager() {
 		connectionManager.setDefaultMaxPerRoute(getMaxConnections());
 
 		if (isStaleChecking()) {
-			log.info("set up connectionManager, setting stale checking ["+isStaleChecking()+"]");
+			log.info("set up connectionManager, setting stale checking [true]");
```
tnleeuw, line 367: Why not actually move that log line out of the if-condition? We don't log this message now if stale-checking is false. That's odd.

tnleeuw (approved):

core/src/main/java/org/frankframework/http/HttpSessionBase.java
nielsm5: Yea, saw it in JJ's PR, thought it was at least better this way, but I agree it's a bit silly...

PR #6588, opened 2024-04-04
Change package name for BusMessages

nielsm5:


management-gateway/src/main/java/org/frankframework/management/bus/message/EmptyResponseMessage.java
```diff
@@ -13,9 +13,9 @@
    See the License for the specific language governing permissions and
    limitations under the License.
 */
-package org.frankframework.management.bus;
+package org.frankframework.management.bus.message;
 
-public class EmptyResponseMessage extends StringResponseMessage {
+public class EmptyResponseMessage extends StringMessage {
```
jkosternl, line 18:
```suggestion
public class EmptyMessage extends StringMessage {
```
Or is this one always a response?

jkosternl (approved):

management-gateway/src/main/java/org/frankframework/management/bus/message/EmptyResponseMessage.java
nielsm5: Yea sorry, forgot to rename that one...

PR #6600, opened 2024-04-05
Console rewrite frontend start & data flow

Matthbo:
Improves app-component, app-service & poller


console/frontend/src/main/frontend/src/app/services/poller.service.ts
```diff
@@ -23,243 +25,175 @@ export class PollerService {
...
-      this.poller = window.setTimeout(() => this.fn(), this.pollerInterval);
-      this.lastPolled = now;
-    } else
-      this.poller = window.setInterval(() => this.fn(), this.pollerInterval);
-  }
+    this.errorCount = 0;
+    this.runningSubscription = this.http.get<T>(this.url).subscribe({
+      next: this.callback,
+      error: (error) => {
+        this.errorCount += 1;
```
philipsens, line 177: Maybe?

```suggestion
        this.errorCount++;
```

philipsens (approved):
Good readable and understandable code. I have no questions or complaints.

console/frontend/src/main/frontend/src/app/services/poller.service.ts
Matthbo: I dont think there is any difference to which one you use but in my opinion is `+=` much clearer than using `var++`. 
We're not writing in C/C++ and this looks more explicit. Also then we dont have to argue about using `++var` and if thats more appropriate compared to `var++` (yes there technically is a difference)

PR #6616, opened 2024-04-08
Fix no Ladybug / slf4j logging

jacodg:


nielsm5:
Cannot check out branch, updated #6633. Thanks for pointing this out, forgot to update it in #6544!

PR #6625, opened 2024-04-09
Add HasStatistics interface to AbstractPipe

nielsm5:


tnleeuw (approved):

core/src/main/java/org/frankframework/lifecycle/ShowLogDirectory.java
```diff
@@ -45,10 +45,19 @@
...
+	private String defaultLogDirectory;
+	private String defaultLogWildcard = AppConstants.getInstance().getProperty("log.viewer.wildcard");
+	private boolean showDirectories = AppConstants.getInstance().getBoolean("log.viewer.showdirectories", true);
+	private int maxItems = AppConstants.getInstance().getInt("log.viewer.maxitems", 500);
+
+	public ShowLogDirectory() {
+		String logdir = AppConstants.getInstance().getProperty("log.dir");
+		if(StringUtils.isEmpty(logdir)) {
+			throw new IllegalStateException("unknown log directory, property [log.dir] has not been set!");
+		}
```
jkosternl, line 57: Should we also check if it exists?

core/src/main/java/org/frankframework/pipes/AbstractPipe.java
```diff
@@ -82,7 +83,7 @@
  *
  * @see PipeLineSession
  */
-public abstract class AbstractPipe extends TransactionAttributes implements IPipe, EventThrowing, ApplicationContextAware, IWithParameters {
+public abstract class AbstractPipe extends TransactionAttributes implements IPipe, EventThrowing, ApplicationContextAware, IWithParameters, HasStatistics {
```
jkosternl, line 86: Adding it here enables us to remove the declaration of this interface at e.g. `MessageSendingPipe` right? Otherwise it is duplicate.

jkosternl (changes requested):

core/src/main/java/org/frankframework/lifecycle/ShowLogDirectory.java
nielsm5: What do you mean? if it exists, it wont throw?

core/src/main/java/org/frankframework/pipes/AbstractPipe.java
nielsm5: I can confirm there is indeed 1 duplicate now.

core/src/main/java/org/frankframework/lifecycle/ShowLogDirectory.java
jkosternl: If the directory to log to exists.

jkosternl (approved):

PR #6629, opened 2024-04-10
Fix DirectoryCleaner documentation

nielsm5:


core/src/main/java/org/frankframework/util/Misc.java
```diff
@@ -293,6 +293,9 @@ public static String getDurationInMs(long value) {
 		return duration + "ms";
 	}
 
+	/**
+	 * @return 'age' in MS.
```
jkosternl, line 297:
```suggestion
	 * @return 'age' in milliseconds.
```
I don't like Microsoft 😉

core/src/main/java/org/frankframework/util/DirectoryCleaner.java
```diff
@@ -113,7 +113,8 @@ public String getDirectory() {
 	}
 
 	/**
-	 * time (with suffix 'd', 'h', 'm' or 's' in milliseconds) that must have passed at least before a file will be deleted
+	 * Time (with suffix 'd', 'h', 'm' or 's') that must have passed at least before a file will be deleted.
```
jkosternl, line 116:
```suggestion
	 * Time (with suffix 'd', 'h', 'm' or 's') that must have passed before a file will be deleted.
```

jkosternl (approved):

core/src/main/java/org/frankframework/util/DirectoryCleaner.java
tnleeuw:
```suggestion
	 * Minimum amount of time (with suffix 'd', 'h', 'm' or 's') that must have passed before a file will be deleted.
```

tnleeuw (approved):

PR #6640, opened 2024-04-10
Improve FrankDoc on EhCache attribute diskPersistent

Laurens-makel:


nielsm5 (approved):

Laurens-makel:
@nielsm5 What about configuration reloads? Does the cache persist to disk and 'survive' the reload, or is it refreshed? Might be worth adding this to the docs.

nielsm5:
If it's persistent to disk it will survive everything as long as the file on disk is not removed 😃

Laurens-makel:
> If it's persistent to disk it will survive everything as long as the file on disk is not removed 😃

I understand, however I wasn't entirely sure when the cache was written to disk :) I'll modify the docs.

nielsm5 (approved):

PR #6646, opened 2024-04-11
Full makeover of iDin step 1

nielsm5:
Since they (rightfully so) finally uploaded the source code to GitHub I've taken the liberty to copy (and transform) some of their (4) JUnit tests. Yes, the official app only has 4 working tests, and only 1 was actually useable.

More (or less) info can be found here: https://github.com/Currence-Online/iDIN-libraries-java


idin/src/main/java/org/frankframework/extensions/idin/IdinSender.java
```diff
@@ -172,11 +237,13 @@ public SenderResult sendMessage(Message message, PipeLineSession session) throws
 			throw new SenderException(e);
 		}
 
+		Communicator communicator = createCommunicator(session);
```
jkosternl, line 240: The creation of the Communicator needs to be done for every message send?

jkosternl (approved):

idin/src/main/java/org/frankframework/extensions/idin/IdinSender.java
nielsm5: Unfortunately yes, otherwise the returnURL cannot be made dynamic...

idin/src/main/java/org/frankframework/extensions/idin/IdinSender.java
```diff
@@ -301,13 +378,9 @@ else if("AUTHENTICATE".equals(getAction())) {
 				}
 			}
```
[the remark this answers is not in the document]
nielsm5: I've tried to keep the changes to a minimum, otherwise the PR will become too large. (hence it being part1 😄)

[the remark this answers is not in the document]
jkosternl: Yeah, good idea.

idin/src/main/java/org/frankframework/extensions/idin/IdinSender.java
jkosternl: Ok

tnleeuw (approved):

PR #6656, opened 2024-04-12
Improved non-view components & made them standalone

Matthbo:
Every component & directive in `src/app/components` should now be standalone and improved.
Removed as much of jquery as possible (only dependency required jquery is left)


philipsens (approved):
Looks good!

PR #6718, opened 2024-04-24
Show warning on low free disk space

### LINKED ISSUE #990: Show warning on low free space on disk
**Reporter**
Jaco de Groot

Related to https://github.com/ibissource/iaf/pull/881 and https://github.com/ibissource/iaf/pull/988

It would be nice to show a warning in the console when free space on disk drops below a certain threshold. In pull request #881 this was done for the Ladybug to prevent corrupt store files (besides showing a warning it doesn't write to disk anymore). See AppConstants.properties in #988 for the threshold:

\# file storage related properties (when freeSpaceMinimum is not specified it is calculated as maxFileSize * (maxBackupIndex + 1) * 10)
\#ibistesttool.maxFileSize=1MB
\#ibistesttool.maxBackupIndex=9
\#ibistesttool.freeSpaceMinimum=10MB

We could do a similar calculation for the console or just use a property with a fixed value.

It might also be nice to show a warning when the amount of free memory drops below a certain threshold. But that might be tricky because there's much more fluctuation. And probably needs a separate feature request.

\-

**Additional context**
\-

philipsens:
Closes #990

### DIFF
--- a/console/frontend/src/main/frontend/src/app/views/status/status.component.html
+++ b/console/frontend/src/main/frontend/src/app/views/status/status.component.html
@@ -12,11 +12,11 @@
     >
       Disk space is running low! Only
       {{ serverInfo.fileSystem.freeSpace | humanFileSize }} available, that's
-      {{ round(freeDiskSpacePercentage) }}% of the total disk size ({{
+      {{ freeDiskSpacePercentage }}% of the total disk size ({{
         serverInfo.fileSystem.totalSpace | humanFileSize
       }})
     </li>
-    <ng-template *ngIf="alerts.length > 0">
+    <ng-container *ngIf="alerts.length > 0">
       <li
         *ngIf="messageLog[selectedConfiguration].errorStoreCount === 1"
         class="list-group-item alert-danger"
@@ -49,7 +49,7 @@
           errorStore.count.show is not set to true</span
         >
       </li>
-      <ng-template [ngForOf]="alerts" let-alert ngFor>
+      <ng-container *ngFor="let alert of alerts">
         <li
           *ngIf="
             alert.configuration === '' ||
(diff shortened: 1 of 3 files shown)


console/frontend/src/main/frontend/src/app/views/status/status.component.ts
```diff
@@ -13,10 +13,12 @@ import {
   MessageLog,
   MessageSummary,
   Receiver,
+  ServerInfo,
   Summary,
 } from 'src/app/app.service';
 import { PollerService } from 'src/app/services/poller.service';
 import { getProcessStateIcon, getProcessStateIconColor } from 'src/app/utils';
+import { round } from '@popperjs/core/lib/utils/math';
```
Matthbo, line 21: Wouldn't it be better to use Math.round() instead of a dependency function?

console/frontend/src/main/frontend/src/app/views/status/status.component.html
```diff
@@ -1,66 +1,82 @@
...
-      *ngIf="messageLog[selectedConfiguration].errorStoreCount > 1"
-      class="list-group-item alert-danger"
-    >
-      <i class="m-r-md fa fa-info-circle"></i>
-      <span
-        >Errorlog contains
-        {{ messageLog[selectedConfiguration].errorStoreCount }} records. Service
-        management should check whether this record has to be resent or
-        deleted</span
+    <ng-template *ngIf="alerts.length > 0">
```
Matthbo, line 19: This *ngIf wont properly work on ng-template, see https://stackoverflow.com/a/44838187

Matthbo (changes requested):

Matthbo (approved):

PR #6754, opened 2024-05-02
Replace PrismJS with the Monaco Editor

philipsens:


Matthbo (changes requested):
[View adapter configuration XML](http://localhost:4200/#/configurations?name=IAF_Util&adapter=ManageFileSystem) links (found on the status page for each adapter) don't seem to work anymore

Matthbo (approved):

PR #6762, opened 2024-05-03
In FF! keep one character more when truncating then in Ladybug

mhdirkse:
Both the Frank!Framework's ladybug module and ladybug itself truncate messages when they are too long. Because the F!F does the truncation first, Ladybug does not see that truncation was necessary. Because of this, Ladybug did not report that a report message was truncated.

This PR lets the F!F keep one extra character such that Ladybug does reduce the message size when truncation is needed. With this PR, the old Ladybug GUI does show it when truncation has been done:

messageTruncatedOldLadybug

The new Ladybug GUI does not show truncation even with the present PR included in the F!F, see:

messageTruncatedNewLadybug

### DIFF
--- a/ladybug/src/main/java/org/frankframework/ibistesttool/MessageEncoder.java
+++ b/ladybug/src/main/java/org/frankframework/ibistesttool/MessageEncoder.java
@@ -25,15 +25,27 @@ import org.apache.commons.io.IOUtils;
 import org.apache.commons.io.input.BoundedInputStream;
 import org.apache.commons.io.input.BoundedReader;
 import org.frankframework.stream.Message;
+import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.annotation.Autowired;
 
 import lombok.SneakyThrows;
 import nl.nn.testtool.Checkpoint;
 import nl.nn.testtool.MessageEncoderImpl;
 
-public class MessageEncoder extends MessageEncoderImpl {
+public class MessageEncoder extends MessageEncoderImpl implements InitializingBean {
 
 	private @Autowired int maxMessageLength;
+
+	// Both Ladybug and the present class in the Frank!Framework
+	// truncate messages as configured through maxMessageLength.
+	// In this class we keep one character more such that
+	// Ladybug can detect when a message is being truncated.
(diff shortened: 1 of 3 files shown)


mhdirkse:
Ik zie een bot melding over code coverage. Als deze code wordt gemerged, dan kan ik later zorgen voor een integratie test in ladybug-ff-cypress-test. Is dit de moeite waard en heeft het prioriteit?

nielsm5:
Als de ladybug maxSize +1 krijgt dan weet de ladybug toch niet hoe groot het bericht was? Ik zie in het plaats '55 characters removed' staan. Ik neem aan dat dat dan altijd '1 characters removed' zal worden?

ladybug/src/main/java/org/frankframework/ibistesttool/MessageEncoder.java
```diff
@@ -25,16 +25,28 @@
...
 
-public class MessageEncoder extends MessageEncoderImpl {
+public class MessageEncoder extends MessageEncoderImpl implements InitializingBean {
 
 	private @Autowired int maxMessageLength;
 
+	// Both Ladybug and the present class in the Frank!Framework
+	// truncate messages as configured through maxMessageLength.
+	// In this class we keep one character more such that
+	// Ladybug can detect when a message is being truncated.
```
jacodg, line 42: It would be nice to not duplicate the comment in two classes. A possible solution would be to add something like the following with the comment to one of the classes and use this var in both classes:

protected static final long MAX_MESSAGE_LENGTH_ADJUSTMENT = 1L;

ladybug/src/main/java/org/frankframework/ibistesttool/MessageEncoder.java
```diff
@@ -25,16 +25,28 @@
...
+	private int truncateMessageToLength = -1;
+
```
jacodg, line 44: Instead of this extra var and method you could do "maxMessageLengt + MAX_MESSAGE_LENGTH_ADJUSTMENT" where you now used truncateMessageToLength (not sure what would be the best solution)

mhdirkse:
You can see from the screenshot above that this solution does not work.

PR #6799, opened 2024-05-09
Fix log levels and statements

nielsm5:


core/src/main/java/org/frankframework/util/RunStateManager.java
```diff
@@ -15,15 +15,15 @@
 */
 package org.frankframework.util;
 
-import org.apache.logging.log4j.Logger;
+import lombok.extern.log4j.Log4j2;
```
Matthbo, line 18: Log4j2 through Lombok?

Matthbo (approved):

jkosternl (approved):

PR #6813, opened 2024-05-13
Add data-cy* for testing whether configs have warnings

mhdirkse:


console/frontend/src/main/frontend/src/app/views/status/status.component.html
```diff
@@ -84,15 +85,21 @@
...
         </li>
         <li
           *ngFor="let configuration of configurations"
           [ngClass]="{ active: selectedConfiguration === configuration.name }"
         >
-          <a (click)="changeConfiguration(configuration.name)">{{
-            configuration.name
-          }}</a>
+          <a
+            [attr.data-cy-status-select-configuration]="configuration.name"
```
mhdirkse, line 99: See https://stackoverflow.com/questions/59453410/angular-adding-cypress-data-cy-attribute - explains that dynamic binding should be done like this.

console/frontend/src/main/frontend/src/app/views/status/status.component.html
```diff
@@ -84,15 +85,21 @@
     <div class="col-lg-12 tabs-container">
       <ul class="nav nav-tabs" role="menu" uib-dropdown-menu="">
         <li [ngClass]="{ active: selectedConfiguration === 'All' }">
-          <a (click)="changeConfiguration('All')">All</a>
+          <a
+            [attr.data-cy-status-select-configuration]="'All'"
+            (click)="changeConfiguration('All')"
+            >All</a
+          >
```
mhdirkse, line 92: I tested with a simple HTML file that HTML allows you to add additional spacing inside the closing tag. These extra spaces were required by ESLint.

console/frontend/src/main/frontend/src/app/views/status/status.component.html
Matthbo: This is correct since its binding a property in an object

mhdirkse:
Replaced by https://github.com/frankframework/frankframework/pull/6832 because I can no longer push on https://github.com/frankframework/master

PR #6826, opened 2024-05-15
Update Apache Tika

evandongen:
upgrade Tika dependency


aspose/pom.xml
```diff
@@ -27,197 +27,26 @@
...
-		</dependency>
 
 		<!-- Tika library -->
 		<dependency>
 			<groupId>org.apache.tika</groupId>
-			<artifactId>tika-parsers</artifactId>
-			<version>1.28.5</version>
+			<artifactId>tika-parser-microsoft-module</artifactId>
+			<version>2.9.2</version>
+			<type>pom</type>
```
nielsm5, line 36: type pom?

erikvandongen: fixed

nielsm5 (approved):

core/pom.xml
```diff
@@ -646,7 +646,7 @@
 		<dependency>
 			<groupId>org.apache.tika</groupId>
 			<artifactId>tika-core</artifactId>
-			<version>1.28.5</version>
+			<version>2.9.2</version>
 			<exclusions>
 				<exclusion><!-- excluded because we want to ignore this (1.2.2) version! this dependency handles javax.activation.* imports, and is NOT compatible with the IBM JRE -->
```
jkosternl, line 651: Are you sure this exclusion is still needed? In the above `aspose/pom.xml`, you've removed it.

jkosternl (approved):

core/pom.xml
jkosternl: Yes, but I was just thinking that this library isn't a dependency of the tika-core anymore.

evandongen: Good point, I'll remove the comment and exclusion

PR #6885, opened 2024-05-28
Fix output of MigratorTest

tnleeuw:
I believe the changes in this PR are correct and depend on driver version more than database version, but would appreciate if someone could run them against their locally running databases in Docker.

In principle only `org.frankframework.jdbc.migration.MigratorTest` (https://github.com/frankframework/frankframework/blob/8512a162779bcf4c5b7c7da52e6cbbb6efd6586c/core/src/test/java/org/frankframework/jdbc/migration/MigratorTest.java) needs to be ran, with any of the affected databases running.

(Some other tests might still fail and should be fixed in new PRs)


nielsm5:
Jenkins will test all databases, so I'm sure that if all is green it's ok?

nielsm5 (approved):

tnleeuw:
> Jenkins will test all databases, so I'm sure that if all is green it's ok?

Jenkins will not test databases in unit tests, only with Larva tests of IAF-Test.

Otherwise this problem would have been found long ago.

jkosternl (approved):

PR #6891, opened 2024-05-28
Remove FilePipe implementations and replace with LocalFileSystem where possible

evandongen:
I've migrated the FilePipe to LocalSystemFilePipe where possible. We have a few scenario's where this isn't possible. The FilePipe had an 'outputType' attribute which isn't available in the FixedResultPipe and/or the LocalFileSystemPipe. When reading binary data (sample.pdf for instance), the FilePipe is still used.


nielsm5 (approved):

test/src/main/configurations/MainConfig/ConfigurationCreateLargeFile.xml
```diff
@@ -51,13 +54,14 @@
 			<pipe name="SetEndIndex"
 				className="org.frankframework.pipes.XsltPipe"
 				getInputFromSessionKey="metrics"
-				xpathExpression="floor((number(substring(processMetrics/properties/property[@name='maxMemory'], 0, string-length(processMetrics/properties/property[@name='maxMemory'])))-160) div 32)"
+				xpathExpression="floor((number(substring(processMetrics/properties/property[@name='heapSize'], 0, string-length(processMetrics/properties/property[@name='heapSize']) - 2))) div 32)"
```
evandongen, line 57: This Configuration is disabled by default. The xpath expression was outdated. It tries to read the metrics from the jvm but that doesn't look like the expressions expects anymore. For now, I've changed this to heapSize to have a sensible default when this is enabled.

FYI, the xml looks like this:
```
<processMetrics>
	<properties>
		<property name="freeMemory">205MiB</property>
		<property name="totalMemory">452MiB</property>
		<property name="heapSize">246MiB</property>
		<property name="maxMemory">7GiB</property>
		<property name="currentTime">2024-05-31 11:51:26.106</property>
	</properties>
</processMetrics>
```
I think the "MiB" and "GiB" weren't present when this was written for an older JVM.

nielsm5:
Superseded by #6926

PR #6904, opened 2024-05-29
Improve quick-submit-form directive & apply to more textarea fields

Matthbo:


console/frontend/src/main/frontend/src/app/views/jdbc/jdbc-execute-query/quick-submit-form.directive.spec.ts
```diff
@@ -34,13 +36,13 @@ describe('QuickSubmitFormDirective', () => {
     );
   });
 
-  it('emits event when ctrl+enter is pressed', () => {
+  it('emits event when ctrl+enter is pressed', fakeAsync(() => {
```
philipsens, line 39: Could you check if this is still needed?

Matthbo: It wasnt needed anymore

philipsens (approved):

PR #6945, opened 2024-06-03
Make restoreMovedElements streaming, by using the FullXmlFilter

jkosternl:


core/src/main/java/org/frankframework/processors/InputOutputPipeProcessor.java
```diff
@@ -150,7 +160,7 @@ protected PipeRunResult processPipe(@Nonnull PipeLine pipeLine, @Nonnull IPipe p
 					handler.setContext(pipeLineSession);
 					try {
 						XmlUtils.parseXml(inputSource, handler);
-						pipeRunResult.setResult(xmlWriter.toString());
+						pipeRunResult.setResult(xmlWriter.getWriter());
```
jkosternl, line 163: Hoping this is a positive impact on performance.

core/src/main/java/org/frankframework/processors/InputOutputPipeProcessor.java
```diff
@@ -118,23 +118,33 @@ protected PipeRunResult processPipe(@Nonnull PipeLine pipeLine, @Nonnull IPipe p
...
+					XmlWriter xmlWriter = new XmlWriter();
+					RestoreMovedElementsHandler handler = new RestoreMovedElementsHandler(xmlWriter);
+					handler.setSession(pipeLineSession);
+					try {
+						XmlUtils.parseXml(inputSource, handler);
+						pipeRunResult.setResult(xmlWriter.getWriter());
+					} catch (Exception e) {
+						log.warn("Pipeline of adapter [{}] could not restore moved elements on the received message: {}", owner.getName(), e.getMessage());
 					}
 				}
```
tnleeuw, line 138: It would be nice to have this code extracted into a separate method, to clean up the main method and make the flow of that more readable.

core/src/main/java/org/frankframework/processors/InputOutputPipeProcessor.java
```diff
@@ -118,23 +118,33 @@ protected PipeRunResult processPipe(@Nonnull PipeLine pipeLine, @Nonnull IPipe p
...
+					}
+					XmlWriter xmlWriter = new XmlWriter();
+					RestoreMovedElementsHandler handler = new RestoreMovedElementsHandler(xmlWriter);
+					handler.setSession(pipeLineSession);
+					try {
+						XmlUtils.parseXml(inputSource, handler);
+						pipeRunResult.setResult(xmlWriter.getWriter());
+					} catch (Exception e) {
+						log.warn("Pipeline of adapter [{}] could not restore moved elements on the received message: {}", owner.getName(), e.getMessage());
 					}
```
tnleeuw, line 137: I think that since we are overwriting the result in the pipeRunResult, we should also close the original `result` message.

core/src/main/java/org/frankframework/processors/InputOutputPipeProcessor.java
tnleeuw: I think this is very tricky and not reliable. This relies on the embedded `Writer` in the `XmlWriter` being a `StringWriter`, on which `writer.toString()` can be called to get the proper XML message.

Of course, I do realise that `xmlWriter.getWriter()` is also just calling `writer.toString()`, which is not quite safe as that writer is sometimes a FileWriter.

I think best is to create the writer yourself then call the `.toString()` on that.

In future, we want to add some sort of MessageBuilder class that can be used to dynamically create messages from different kinds of streams.
This can then be easily plugged in place here.

tnleeuw (changes requested):

core/src/main/java/org/frankframework/processors/InputOutputPipeProcessor.java
jkosternl: The whole reason for doing this PR, is to avoid the whole message is loaded into memory. Otherwise, we could just do the previous implementation and some String replacements (exaggeratedly expressed). So, I hoped that using the writer could work. Or this PR is an in-between state, hoping the MessageBuilder will come in the future.

core/src/main/java/org/frankframework/processors/InputOutputPipeProcessor.java
jkosternl: Good idea, will do!

core/src/main/java/org/frankframework/processors/InputOutputPipeProcessor.java
tnleeuw: I think that at least this PR allows easily plugging in a sort of build in future.

Right now, since it uses a StringWriter in the XmlWriter, it does not yet address the memory requirements (unfortunately).

We'd need a writer that doesn't always writer everything into memory for that, but also does not always write everything to file always as that will slow down the system on small messages.

evandongen (approved):

core/src/main/java/org/frankframework/util/RestoreMovedElementsHandler.java
```diff
@@ -15,61 +15,108 @@
...
+
+	@Override
+	public void endElement(String uri, String localName, String qName) throws SAXException {
+		processCharacterData(false);
+		super.endElement(uri, localName, qName);
+	}
+
+	@Override
+	public void characters(char[] ch, int start, int length) throws SAXException {
+		charDataBuilder.append(ch, start, length);
```
nielsm5, line 58: Poeh, ik had hier eingelijk de find/replace verwacht. Niet in de start- en endElement methodes.

jkosternl: Ja, dat had ik ook, toen ik begon. En dan ga je het bouwen en zie je dat dat niet werkt. Je moet juist bij het starten van een element alle charData verzamelen en controleren van wat er voor ontvangen is.

tnleeuw (approved):

philipsens:
Could you rebase on master? The console docker image can't build because it is missing in here.

tnleeuw (approved):

nielsm5 (approved):

PR #6956, opened 2024-06-05
Remove HttpOnly from CSRF cookie

Matthbo:


security/src/main/java/org/frankframework/lifecycle/servlets/ServletAuthenticatorBase.java
```diff
@@ -174,7 +174,8 @@ public SecurityFilterChain configureHttpSecurity(HttpSecurity http) {
 			//Apply defaults to disable bloated filters, see DefaultSecurityFilterChain.getFilters for the actual list.
 			http.headers(headers -> headers.frameOptions(HeadersConfigurer.FrameOptionsConfig::sameOrigin)); //Allow same origin iframe request
 			if(csrfEnabled){
-				CookieCsrfTokenRepository csrfTokenRepository = new CookieCsrfTokenRepository();
+				//HttpOnly needs to be false for Angular to read it
```
jkosternl, line 177: Angular supports HttpOnly, as what I for example read at this website: https://www.bezkoder.com/angular-13-jwt-auth-httponly-cookie/

Matthbo: Angular is not allowed to read an httponly cookie in some if not most browsers https://stackoverflow.com/a/8069697

nielsm5 (approved):

PR #6970, opened 2024-06-07
Add search & sort to uploaded configurations table

Matthbo:
Also fixes the configuration upload endpoint, using an old function I didnt think was needed anymore so I had deleted it when moving over to Spring

Yes I know making an `anyCompare` that doesnt properly compare any `keyof T` type where T is the array of table items is kind of bad but I rewrote it like 3 times already, each time overengineering it too much for what is currently needed.
This is now a nice basis that can later be expanded if its _really_ needed.

### DIFF
--- a/console/backend/src/main/java/org/frankframework/management/web/FrankApiBase.java
+++ b/console/backend/src/main/java/org/frankframework/management/web/FrankApiBase.java
@@ -16,6 +16,7 @@
 package org.frankframework.management.web;
 
 import jakarta.annotation.Nonnull;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.frankframework.management.bus.OutboundGateway;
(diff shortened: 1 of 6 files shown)


console/backend/src/main/java/org/frankframework/management/web/FrankApiBase.java
```diff
@@ -92,4 +95,11 @@ protected <T> T getProperty(String key, T defaultValue) {
 	protected final boolean allowDeprecatedEndpoints() {
 		return getProperty(DeprecationInterceptor.ALLOW_DEPRECATED_ENDPOINTS_KEY, false);
 	}
+
+	protected String getUserPrincipalName(Principal principal) {
+		if(principal != null && StringUtils.isNotEmpty(principal.getName())) {
+			return principal.getName();
+		}
+		return null;
+	}
```
nielsm5, line 104: I'm afraid this will conflict with my changes in #6973. I think we should merge this, then the POC (without the test endpoint). And ensure that all users are being retrieved from Spring Security opposed to the http Principal.

nielsm5 (approved):

philipsens (approved):

PR #6973, opened 2024-06-08
Minor cleanups and add WebSocket POC

nielsm5:
Ideally I'd like to use the ...web.filters, web.socket and web.configuration folders.
Might be best to move packages around in a different pr...
(We also need to ensure that the only class loaded by the parent is the ServletDispatcher.)


Matthbo (commented):
So if I understand it correctly, setting the namespace means it wont inherit the parent's mappings and setDetectAllHandlerMappings to false was what blocked websocket handlers from being registered?
I'm very happy we can connect to the endpoint now :)

nielsm5:
> So if I understand it correctly, setting the namespace means it wont inherit the parent's mappings and setDetectAllHandlerMappings to false was what blocked websocket handlers from being registered? I'm very happy we can connect to the endpoint now :)

There were a two main things going on, in the `DeprecationInterceptor` the handler had a ClassCastException.
and the mapping detection mechanism was disabled so it wouldn't inherit from the parent but the `setApplicationContext` method being overwritten did the same thing. This caused the Dispatcher to not pick up all handlers.

The namespace is just for debugging purposes.

jkosternl (approved):

evandongen (approved):

PR #6997, opened 2024-06-12
Add option to disable larva save button

nielsm5:


jkosternl (approved):
Good idea!

larva/src/main/java/org/frankframework/larva/LarvaServlet.java
```diff
@@ -111,7 +113,7 @@ protected void doPost(HttpServletRequest req, HttpServletResponse resp) throws I
 		if("/".equals(path) || "/index.jsp".equalsIgnoreCase(path)) {
 			handleIndex(req, resp);
 			return;
-		} else if ("/saveResultToFile.jsp".equals(path)) {
+		} else if (ALLOW_SAVE && "/saveResultToFile.jsp".equals(path)) {
```
tnleeuw, line 116: 1. User gets no feedback that saving is not allowed?
2. Do we want to also disable or remove it in the front-end of Larva?

nielsm5: The user gets a 404 page (can't be a 401 because that triggers a logout).
We need to redo the entire frontend 😄 I think it's best to leave that for now tbh, as the 'frontend' is being generated by a very complicated backend...

tnleeuw (approved):

PR #7023, opened 2024-06-18
Remove feedback and add star in header and buttons to Discussions/Bug report in navigation

### LINKED ISSUE #7010: Change the `Send Feedback` button to make use of GitHub Discussions

Currently, the place where all the feedback ends up, is not open/public to the community. This makes it unclear to community members what the feedback is. It could also be a good source of testimonials or a good source of stuff we need to improve.

### Describe the Feature

Change the feature to make use of GitHub Discussion.

It could still be a modal in the Frank!Console, but then send the values to a New Discussion like this https://github.com/frankframework/frankframework/discussions/new?category=feedback&title=Example&body=This+is+an+example+for+this+issue

No, community resolution only

### Anything else?

The same could be done with `Report a Bug`

### LINKED ISSUE #7011: Add GitHub Star button in console

Enhancing user engagement by integrating a GitHub Star button directly in the console encourages users to support and promote the Frank!Framework. 

### Describe the Feature

A visible GitHub Star button in the console header that replaces the current 'leave feedback' stars.
Clicking the button redirects to the Frank!Framework GitHub repository for starring.

Something like [this](https://buttons.github.io/), or just create a link to leave a GitHub Star.

No, community resolution only

### Anything else?

_No response_

philipsens:
Closes #7010 and closes #7011


Matthbo:
Weren't we going to try to add the environment information automatically to the report a bug button redirect?

The spacing feels a little weird between all these elements, maybe this could use a bit of improvement
image

philipsens:
> Weren't we going to try to add the environment information automatically to the report a bug button redirect?

Yes, I've created #7024. It was just a bit too much for right now. The `serverInfo` should probably be gathered through a service and a method for formatting should be added. In the `environment-modal`, we copy the `innerHTML` for the Environment info, which is nice, because the layout can be defined in the HTML itself. I'm not sure how we can change it, so the text can be used there and also for the issue template. Any ideas?

> The spacing feels a little weird between all these elements, maybe this could use a bit of improvement image

I thought that too here: https://github.com/frankframework/frankframework/issues/7011#issuecomment-2175881917

The header has always looked bad in my opinion, but I'm guessing that it will be redesigned in the future 👀

Matthbo (approved):

PR #7028, opened 2024-06-19
Improve performance of hideRegex in logging

tnleeuw:


commons/src/main/java/org/frankframework/util/StringUtil.java
```diff
@@ -157,9 +156,17 @@ public static String hideAll(String inputString, String regex) {
...
 	public static String hideAll(String inputString, String regex, int mode) {
+		return hideAll(inputString, Pattern.compile(regex), mode);
+	}
+
+	/**
+	 * Hides the input string according to the given regex and mode.
+	 * If mode is set to 1, then the first half of the string gets hidden.
+	 * Else, all of it.
+	 */
+	public static String hideAll(String inputString, Pattern regex, int mode) {
```
jkosternl, line 167: We still need both (overloaded) methods?

jkosternl (approved):

core/src/main/java/org/frankframework/logging/IbisMaskingLayout.java
```diff
@@ -58,12 +58,12 @@ public abstract class IbisMaskingLayout extends AbstractStringLayout {
 	 * Set of regex strings to hide locally, meaning for specific threads/adapters.
 	 * This is required to be set for each thread individually.
 	 */
-	private static final ThreadLocal<Set<String>> threadLocalReplace = new ThreadLocal<>();
+	private static final ThreadLocal<Map<String, Pattern>> threadLocalReplace = new ThreadLocal<>();
```
evandongen, line 61: Perhaps this is too much for now, but why not refactor this to a Set<Pattern>? That way you can only access the compiled Patterns via the ThreadLocal

commons/src/main/java/org/frankframework/util/StringUtil.java
tnleeuw: For now, there are still callers that pass a string where I didn't feel it necessary to optimize.
That can perhaps come in a future PR, if the overloaded version with a `String regex` is no longer used.

core/src/main/java/org/frankframework/logging/IbisMaskingLayout.java
tnleeuw: Let me see -- I thought that under some circumstances, we still wanted to be able to look at the original regexes or be able to remove by original regex. Although that method is perhaps never actually used.

evandongen (approved):

core/src/main/java/org/frankframework/logging/IbisMaskingLayout.java
tnleeuw: I looked at it, and I think this is a good suggestion. I will update this.

nielsm5 (approved):

PR #7039, opened 2024-06-20
Some code refactoring that I couldn't stop myself from doing.

tnleeuw:


core/src/main/java/org/frankframework/management/bus/endpoints/GetSchedules.java
```diff
@@ -302,7 +301,7 @@ private List<Map<String, Object>> getJobMessages(JobDetail jobDetail) {
 			MessageKeeperMessage job = jobdef.getMessageKeeper().getMessage(t);
 
 			message.put("text", job.getMessageText());
-			message.put("date", job.getMessageDate());
+			message.put("date", Date.from(job.getMessageDate()));
```
evandongen, line 304: You convert the Instant to a Date here, but when reading it in `ServerStatistics` you assume it's an Instant, is this correct?

evandongen (changes requested):

core/src/main/java/org/frankframework/management/bus/endpoints/GetSchedules.java
tnleeuw: `message` here is a map, that's sent to the front-end.

In `ServerStatistics` we're not reading that map but again looking at the original, typed, `MessageKeeperMessage` that therefore has an `Instant`.

For the Map I'm doing the conversion to `Date` because otherwise there'll be impact on the other side of the bus and the mapping of items, which I didn't want to dive into and touch now.

So I think this change is correct, but there's room for more improvement in the future.

evandongen (approved):

jkosternl (approved):

nielsm5 (approved):

PR #7042, opened 2024-06-20
fix: add closing tag for head

philipsens:


Matthbo (approved):
This seems like a useful thing to have 😂

philipsens:
> This seems like a useful thing to have 😂

image

PR #7067, opened 2024-06-25
Cleanup Misc class

nielsm5:


filesystem/src/main/java/org/frankframework/filesystem/ForEachAttachmentPipe.java
```diff
@@ -179,9 +180,29 @@ public void setExcludeProperties(String excludePropertiesList) {
...
 	}
 	public Set<String> getExcludePropertiesSet() {
 		return excludeProperties;
 	}
+
+	/**
+	 * Adds items on a string, added by comma separator (ex: "1,2,3"), into a list.
+	 * @param collectionDescription description of the list
+	 */
+	void addItemsToList(Collection<String> collection, String list, String collectionDescription) {
```
jkosternl, line 193: To match the first line of the Javadoc, I'd change the names of the vars to `list` for the first one, and rename the 2nd var from `list` to `input` or at least something that not mixes with the output list.

larva/src/main/java/org/frankframework/extensions/test/IbisTester.java
```diff
@@ -435,7 +440,7 @@ private static String getWebContentDirectory() {
 							}
 						}
 					}
-				} catch (AccessControlException e) {
+				} catch (SecurityException e) {
```
jkosternl, line 443: Ey, die heb ik in mijn Sonar-security PR ook veranderd 😉

nn-specials/src/main/java/org/frankframework/extensions/bis/BisUtils.java
```diff
@@ -208,23 +208,27 @@ public Message prepareReply(Message rawReply, String messageHeader, String resul
...
 				messages.add(result);
-				payload = Misc.listToString(messages);
+				payload = listToString(messages);
 			}
 		}
 		return new Message(payload);
 	}
 
+	public static String listToString(List<String> list) {
+		return String.join("", list);
```
jkosternl, line 229: Mag je ook gewoon inlinen, wat mij betreft.

jkosternl (approved):

nn-specials/src/main/java/org/frankframework/extensions/bis/BisUtils.java
nielsm5: Tja, ik heb puur dingen verplaatst. Ik vond het belachelijke methodes voor `Misc`, dus mijn doel was meer ze daar wegkrijgen.

tnleeuw:
PS: I think there's also a method `Misc.cleanseMessage()`. I think you can also remove that message from the Misc class. It's only used to wrap around `String.hideRegex()` and only in a single place.

But if you don't then I'll do it as part of my continued work on overall `hideRegex` improvements.

tnleeuw (approved):

evandongen (approved):

PR #7079, opened 2024-06-26
Move logout functionality to Spring Security

nielsm5:
Step 1 to cleanup the old endpoint and fix the logout functionality.
Need to move the console servlets out of this class in a next pr.


security/src/main/java/org/frankframework/lifecycle/servlets/ServletAuthenticatorBase.java
```diff
@@ -209,6 +215,27 @@ public SecurityFilterChain configureHttpSecurity(HttpSecurity http) {
 		}
 	}
 
+	// Match when the client matches '<servlet-path>/logout'.
+	private boolean requestMatcher(HttpServletRequest request) {
+		return ("GET".equals(request.getMethod()) && "/logout".equals(request.getPathInfo()));
```
Matthbo, line 220: What would be the servlet path here?
`/iaf/logout` or `/logout`?

nielsm5: Servlet mapping, so `iaf/api/` + `logout` or `iaf/gui/` + `logout`.
In a new PR I want to move this, as well as CSRF (and session management) to a different configurer for the Console, so both endpoints (frontend/backend) are seen as 1, as well as not using all this silly stuff for all web-service endpoints.

jkosternl (approved):
Should we also proof its effectiveness in a unit test or integration test?

nielsm5:
> Should we also proof its effectiveness in a unit test or integration test?

I'm not quite sure how I can test the logout functionality in a JUnit test 😞

security/src/main/java/org/frankframework/lifecycle/servlets/ServletAuthenticatorBase.java
Matthbo: Okay, so angular also tries to work with a route on `/iaf/gui/#/logout`
This shouldnt be an issue but it could be, maybe I should change that route in a next pr so I can be sure it wont ever conflict

Matthbo (approved):

PR #7113, opened 2024-07-05
Remove the interface IAdapter, use Adapter instead everywhere

tnleeuw:
As discussed last Wednesday, the IAdapter interface can go.


test/src/main/resources/DeploymentSpecifics.properties
```diff
@@ -257,7 +257,5 @@ imapmail.logFolder=Log
 # configure an alias when you want to run the scenarios in secure mode
 framework.api.user.alias=
 
-msg.log.xPath.value=/test/value
```
tnleeuw, line 260: Property removed here because it is duplicate, higher up in the file it is already declared with the same value.

nielsm5 (approved):

evandongen (approved):

PR #7115, opened 2024-07-07
Minor Larva bugfixes

nielsm5:
Turns multithreaded into opt-in.

Fixes:
image

### DIFF
--- a/larva/src/main/java/org/frankframework/larva/LarvaTool.java
+++ b/larva/src/main/java/org/frankframework/larva/LarvaTool.java
@@ -179,7 +179,7 @@ public class LarvaTool {
 		if (paramAutoScroll == null && paramLogLevel != null) {
 			config.setAutoScroll(false);
 		}
-		if (("on".equals(paramMultithreaded) || Boolean.parseBoolean(paramMultithreaded)) && paramLogLevel != null) {
+		if (StringUtils.isNotEmpty(paramMultithreaded) && Boolean.parseBoolean(paramMultithreaded) && paramLogLevel != null) {
 			config.setMultiThreaded(true);
 		}
 
@@ -452,7 +452,7 @@ public class LarvaTool {
 		writeHtml(TR_CLOSING_TAG, false);
 		writeHtml(TR_STARTING_TAG, false);
 		writeHtml(TD_STARTING_TAG, false);
-		writeHtml("<input type=\"checkbox\" name=\"multithreaded\"", false);
+		writeHtml("<input type=\"checkbox\" name=\"multithreaded\" value=\"true\"", false);
 		if (config.isMultiThreaded()) {
 			writeHtml(" checked", false);
 		}
--- a/larva/src/main/java/org/frankframework/larva/LarvaTool.java
+++ b/larva/src/main/java/org/frankframework/larva/LarvaTool.java
@@ -179,8 +179,8 @@ public class LarvaTool {
 		if (paramAutoScroll == null && paramLogLevel != null) {
 			config.setAutoScroll(false);
 		}
-		if (paramMultithreaded == null && paramLogLevel != null) {
-			config.setMultiThreaded(false);
+		if (("on".equals(paramMultithreaded) || Boolean.parseBoolean(paramMultithreaded)) && paramLogLevel != null) {
+			config.setMultiThreaded(true);
 		}
 
 		if (!silent) {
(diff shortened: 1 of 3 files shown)


larva/src/main/java/org/frankframework/larva/LarvaTool.java
```diff
@@ -448,11 +448,11 @@ public void printHtmlForm(List<String> scenariosRootDirectories, List<String> sc
 		writeHtml("<span style=\"float: left; font-size: 10pt; width: 0px\">&nbsp; &nbsp; &nbsp;</span>", false);
 		writeHtml("<table style=\"float:left;height:50px\">", false);
 		writeHtml(TR_STARTING_TAG, false);
-		writeHtml("<td>Multi threaded (beta)</td>", false);
+		writeHtml("<td>Multi Threaded (experimental)</td>", false);
 		writeHtml(TR_CLOSING_TAG, false);
 		writeHtml(TR_STARTING_TAG, false);
 		writeHtml(TD_STARTING_TAG, false);
-		writeHtml("<input type=\"checkbox\" name=\"multithreaded\" value=\"true\"", false);
+		writeHtml("<input type=\"checkbox\" name=\"multithreaded\"", false);
```
tnleeuw, line 455: This is not consistent with other checkboxes, which all send the value `"true"` when checked.
Writing value `true` doesn't mean the checkbox is checked, that is done on line 457.

So I suggest reverting this and like 182, to be consistent with all other checkboxes in the file.

See for instance:
https://stackoverflow.com/questions/11424037/do-checkbox-inputs-only-post-data-if-theyre-checked

larva/src/main/java/org/frankframework/larva/LarvaTool.java
```diff
@@ -179,8 +179,8 @@ public int runScenarios(IbisContext ibisContext, String paramLogLevel, String pa
 		if (paramAutoScroll == null && paramLogLevel != null) {
 			config.setAutoScroll(false);
 		}
-		if (paramMultithreaded == null && paramLogLevel != null) {
-			config.setMultiThreaded(false);
+		if (("on".equals(paramMultithreaded) || Boolean.parseBoolean(paramMultithreaded)) && paramLogLevel != null) {
+			config.setMultiThreaded(true);
```
tnleeuw, line 183: I think this check should not depend on `paramLogLevel`. I also think we can revert the `"on"` check to a `null` check, and set the checkbox directly from `paramMultiThreaded != null`, 

Something like this:
```suggestion
		config.setMultiThreaded(paramMultithreaded != null);
```

larva/src/main/java/org/frankframework/larva/TestConfig.java
```diff
@@ -21,24 +21,22 @@
...
+	private @Getter @Setter LarvaLogLevel logLevel = LarvaLogLevel.WRONG_PIPELINE_MESSAGES;
 
-	private boolean autoScroll = true;
-	private boolean useHtmlBuffer = false;
-	private boolean useLogBuffer = true;
-	private boolean multiThreaded = true;
+	private @Getter @Setter boolean autoScroll = true;
+	private @Getter @Setter boolean useHtmlBuffer = false;
+	private @Getter @Setter boolean useLogBuffer = true;
+	private @Getter @Setter boolean multiThreaded = false;
```
tnleeuw, line 36: This was the real problem.

tnleeuw (changes requested):

larva/src/main/java/org/frankframework/larva/TestConfig.java
nielsm5: One of 4 problems.

larva/src/main/java/org/frankframework/larva/LarvaTool.java
nielsm5: Due to the poor design of this method, you need to be able to change a value without larva running all scenario's.
Not why why this was like this, it's not been changed, and I do not intend to do so.

larva/src/main/java/org/frankframework/larva/LarvaTool.java
nielsm5: What does this achieve?

larva/src/main/java/org/frankframework/larva/LarvaTool.java
```diff
@@ -670,7 +670,7 @@ public void wrongPipelineMessage(String stepDisplayName, String message, String
 		writeLog("<pre id='" + diffBoxId + "' class='diffBox'></pre>", method, false);
 		writeLog("</div>", method, false);
 
-		if (config.getLogLevel().shouldLog(LarvaLogLevel.SCENARIO_PASSED_FAILED)) {
+		if (config.getLogLevel() == LarvaLogLevel.SCENARIO_PASSED_FAILED) {
```
evandongen, line 673: Hiermee vergelijk je de reference en niet per se de inhoud.
```suggestion
		if (LarvaLogLevel.SCENARIO_PASSED_FAILED.equals(config.getLogLevel())) {
```

evandongen (changes requested):

larva/src/main/java/org/frankframework/larva/LarvaTool.java
nielsm5: Dit is een enum :)

larva/src/main/java/org/frankframework/larva/LarvaTool.java
tnleeuw: Consistency with rest of the code, that writes checkboxes with `value="true"` also when not checked.

larva/src/main/java/org/frankframework/larva/LarvaTool.java
evandongen: Oeps

evandongen (approved):

tnleeuw (approved):

PR #7121, opened 2024-07-08
Display logout button in console when properly logged in

Matthbo:
The button in the frontend does display now when a username is gotten from the backend.

PR #7101 should've made the `/logout` endpoint work but somehow it won't work in Firefox and is iffy in Chrome, this should be looked into.

### DIFF
--- a/console/frontend/src/main/frontend/src/app/views/login/login.component.html
+++ b/console/frontend/src/main/frontend/src/app/views/login/login.component.html
@@ -14,7 +14,7 @@
     <form
       class="m-t FrankLogin"
       role="form"
-      (submit)="login(credentials)"
+      (submit)="login()"
       ngNativeValidate
     >
       <div class="form-group">
@@ -27,6 +27,7 @@
           placeholder="Username"
           required
           autofocus
+          disabled
         />
       </div>
       <div class="form-group">
@@ -38,6 +39,7 @@
           name="password"
           placeholder="Password"
           required
+          disabled
         />
       </div>
       <div class="form-group">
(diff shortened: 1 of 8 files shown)


console/frontend/src/main/frontend/src/app/views/login/login.component.html
```diff
@@ -27,6 +27,7 @@ <h3>Frank<strong class="text-frank">!</strong>Console</h3>
           placeholder="Username"
           required
           autofocus
+          disabled
```
nielsm5, line 30: ?

Matthbo: The form is currently not being used so I disabled the inputs. This way you only have to submit to go to the main view of the application

console/frontend/src/main/frontend/src/app/views/login/login.component.ts
```diff
@@ -24,19 +25,26 @@ export class LoginComponent implements OnInit {
...
+    private router: Router,
   ) {}
 
   ngOnInit(): void {
-    this.authService.loggedin(); // Check whether the client is logged in.
+    if (this.authService.isLoggedIn()) {
+      this.router.navigate(['/']);
+    }
 
     window.setTimeout(() => {
```
philipsens, line 36: Couldn't this timeout be substituted with ngAfterViewInit?

Matthbo: I guess, I'm not fully sure on what the intent was originally
I'll update it

philipsens (approved):

PR #7125, opened 2024-07-09
Autofill environment field when fling a bug report

philipsens:


console/frontend/src/main/frontend/src/app/views/status/status.component.ts
```diff
@@ -322,18 +326,14 @@ export class StatusComponent implements OnInit, OnDestroy {
...
-          ) / 10;
-      });
+    this.serverInfoService.serverInfo$.pipe(first()).subscribe((serverInfo) => {
+      this.serverInfo = serverInfo;
+      this.freeDiskSpacePercentage =
+        Math.round(
+          (serverInfo.fileSystem.freeSpace / serverInfo.fileSystem.totalSpace) *
+            1000,
+        ) / 10;
+    });
```
Matthbo, line 336: Doesn't this only run when serverinfo is refreshed?
So if you navigate from another view to the status view, does it still trigger this?

Matthbo: That's fair, I was thinking over time it might be nice to have it check every so often but maybe that's a little to extra for this particular issue

Matthbo (approved):

console/frontend/src/main/frontend/src/app/views/status/status.component.ts
philipsens: I can remove the first. Might be usefull for when people are refreshing to see disk space changes

Matthbo: That wouldn't hurt no

PR #7142, opened 2024-07-11
Remove FileRecordListener and move fileToString to nn-specials

evandongen:


nn-specials/src/main/java/org/frankframework/extensions/rekenbox/Utils.java
evandongen, line 1: I might have misinterpreted the ticket on this, but I assumed it should be a new Utils class since there wasn't one in the nn-specials module yet. If this should be in another package, let me know

tnleeuw: If this code is only called from one place, the RekenBoxCaller, then in my opinion it can also be put just in that class.

tnleeuw (approved):

nielsm5 (approved):

PR #7144, opened 2024-07-11
Websocket support & events for warning & adapter updates

Matthbo:
Adds websocket support, send adapter & server/configuration warnings updates over websocket connection.
STOMP is used  for websocket messaging, SockJS is used as fallback which emulates the websocket protocol over http


console/backend/src/main/resources/FrankConsoleContext.xml
```diff
@@ -11,7 +11,7 @@
 
 	<context:property-placeholder />
 
-	<context:component-scan base-package="org.frankframework.management.web,org.frankframework.web">
+	<context:component-scan base-package="org.frankframework.management.web,org.frankframework.web, org.frankframework.management.web.socket">
```
nielsm5, line 14: Dit zou niet moeten hoeven, hij pakt automatisch alle subfolders van `org.frankframework.management.web`

core/src/main/java/org/frankframework/management/bus/endpoints/ServerStatistics.java
```diff
@@ -211,6 +211,12 @@ public Message<String> getApplicationWarnings() {
 		return new JsonMessage(returnMap);
 	}
 
+	@ActionSelector(BusAction.UPDATES)
+	@PermitAll
+	public Message<String> getApplicationWarningsUpdates(){
+		return getApplicationWarnings();
+	}
```
nielsm5, line 218: Wat is hier het nut van?

Matthbo: The current authorisation works by checking the user's roles as if it was a http request with user session, but this information doesn't exist when the backend triggers a scheduled method and sending a gateway message to retrieve the data from core so I had to make this or find a way to put the right roles on a non-user to still get past role authorisation

console/backend/src/main/resources/FrankConsoleContext.xml
Matthbo: Ah this was needed before a big merge from master where the file was deleted and remade, I copied this over to be sure it would still work

nielsm5:
Authentication now has been fixed. The only thing that remains is on how to deal with multiple backend nodes. Though I think it's best to tackle that problem in a new PR 😄.

nielsm5 (approved):

tnleeuw (approved):

PR #7171, opened 2024-07-16
Add an in-mem to disk overflowing outputstream buffer

nielsm5:


core/src/main/java/org/frankframework/compression/ZipWriter.java
```diff
@@ -110,21 +107,20 @@ public MessageZipEntry createPart(Message input, PipeLineSession session, Parame
 
 	@Override
 	public Message build(List<MessageZipEntry> parts) throws IOException {
-		File file;
```
tnleeuw, line 113: 1. These changes remove functionality (to return a filename instead of the file). Probably deprecated, but then the deprecation should be done in 9.0, not 8.3
2. Currently these changes also break 4 tests in IAF-Test that expect the file to exist on disk at some point.

core/src/main/java/org/frankframework/stream/OverflowToDiskOutputStream.java
```diff
@@ -0,0 +1,145 @@
...
+		}
+	}
+
+	private OutputStream createFileOnDisk() throws IOException {
+		fileLocation = Files.createTempFile(tempDirectory, "msg", "dat");
+		return Files.newOutputStream(fileLocation);
+	}
+
+	private OutputStream flushBufferToDisk() throws IOException {
+		if (count == 0 && fos != null) { //buffer has been reset, and fos exists.
```
tnleeuw, line 62: Heel verwarrend om nu in 1 method 2 verschillende variables `fos` te hebben.

core/src/main/java/org/frankframework/stream/OverflowToDiskOutputStream.java
```diff
@@ -0,0 +1,145 @@
...
+	/** 
+	 * Doesn't do anything if the message is kept in memory
+	 */
+	@Override
+	public void flush() throws IOException {
+		flush(false);
+	}
+
+	/**
+	 * Doesn't do anything if the message is kept in memory.
```
tnleeuw, line 119: That comment is not quite correct b/c it depends on value of parameter.

core/src/main/java/org/frankframework/stream/MessageBuilder.java
```diff
@@ -43,12 +43,15 @@ public MessageBuilder() throws IOException {
...
 	/**
 	 * Directly stores to disk.
-	 * Mainly for legacy implementations
+	 * Mainly for legacy implementations which allows the user to choose where to save the file.
 	 */
 	public MessageBuilder(Path file) throws IOException {
-		if(!Files.exists(file)) throw new IOException("file ["+file+"] does not exist");
 		if(Files.isDirectory(file)) throw new IOException("location ["+file+"] may not be a folder");
 
+		if(!Files.exists(file)) {
```
tnleeuw, line 51: Je hoeft niet `Files.createFile` aan te roepen hier, `Files.newOutputStream` doet dat al volgens de JavaDoc.

Ik denk dat het probleem ergens anders zit.

tnleeuw (changes requested):

core/src/main/java/org/frankframework/compression/ZipWriter.java
nielsm5: I've kept all functionality as-is. Nothing has been removed or is broken.

core/src/main/java/org/frankframework/stream/MessageBuilder.java
nielsm5: Ik weet niet precies welk probleem er zou zijn, maar ik kan/zal hem weghalen 😄 .

core/src/main/java/org/frankframework/compression/ZipWriter.java
tnleeuw: It looked to me as if some functionality regarding returning filename or using filename was gone. I may have misinterpreted the code. Looking at the failures in IAF-Test on a missing file that was expected.

tnleeuw (approved):

PR #7174, opened 2024-07-17
Test a pipeline clearing configuration also clears adapter

Matthbo:


Matthbo:
Discussed this with Niels and we both agreed it would be better to always clear the adapter input when the configuration input has changed

MatthijsSmets (approved):

Issue #4326, opened 2023-01-11
supply bucketname to AmazonS3Sender via parameter


LeroyJenkinss:
**❗ please do not add sensitive information in issues, you can provide extra information via email using issue number as reference ❗**

**Describe the issue**
The Amazon3ssender should be able to set the attribute bucketName through a parameter. Yet I can't.
I would like this to be made available.

**Reporter**
Mr Tarik

gvanbrakel:
#4474 and #4477 provide another way to supply the bucketname at runtime: by prefixing it to the filename, separated from it by '|'

PR #7182, opened 2024-07-19
Fix bucketName prefix when creating files

nielsm5:
In https://github.com/frankframework/frankframework/issues/4326 we added the ability to dynamically change the bucketName, this however was very poorly documented and implemented!

This PR fixes the original behavior but whether we must keep it...


tnleeuw (approved):

nielsm5:
> Ideally the bucket prefix should be used identically for all file operations.

I completely agree, unfortunatelly this requires API changes, eg. IWriteableFilesystem#createFolder(String) to createFolder(F).

PR #7188, opened 2024-07-23
Use an AWS S3 file reference instead of misusing the S3Object

nielsm5:
The goal here is purely to fix the folder (name) issues and allow file properties to be set/shared between methods.

### DIFF
--- a/aws/src/main/java/org/frankframework/filesystem/S3FileRef.java
+++ b/aws/src/main/java/org/frankframework/filesystem/S3FileRef.java
@@ -29,6 +29,7 @@ import com.amazonaws.services.s3.model.S3Object;
 import com.amazonaws.services.s3.model.S3ObjectSummary;
 
 import jakarta.annotation.Nonnull;
+import jakarta.annotation.Nullable;
 import lombok.Getter;
 import lombok.Setter;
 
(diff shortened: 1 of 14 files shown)


aws/src/main/java/org/frankframework/filesystem/S3FileRef.java
```diff
@@ -0,0 +1,137 @@
...
+ * 
+ * @author Niels Meijer
+ */
+public class S3FileRef {
+
+	private static final String FILE_DELIMITER = "/";
+	public static final String BUCKET_OBJECT_SEPARATOR = "|";
+
+	private final String name;
+	private final String folder;
```
tnleeuw, line 54: Perhaps you can annotate `name` and `folder` with `@Nullable`, so that it is always clear you have to do a null-check with these? Also for code-analysis in the IDE.

tnleeuw (approved):

aws/src/main/java/org/frankframework/filesystem/S3FileRef.java
nielsm5: Name may be empty, but folder can be nullable, I'll add the annotations ;)

PR #7197, opened 2024-07-25
Speed up startup and shutdown of configurations

tnleeuw:


core/src/main/java/org/frankframework/receivers/Receiver.java
```diff
@@ -1760,7 +1756,7 @@ private String sendResultToSender(Message result) {
 		String errorMessage = null;
 		try(PipeLineSession session = new PipeLineSession()) {
 			log.debug("Receiver [{}] sending result to configured sender [{}]", this::getName, this::getSender);
-			getSender().sendMessageOrThrow(result, null); // sending correlated responses via a receiver embedded sender is not supported
+			getSender().sendMessageOrThrow(result, session); // sending correlated responses via a receiver embedded sender is not supported
```
nielsm5, line 1759: What the...

nielsm5 (approved):

core/src/main/java/org/frankframework/receivers/Receiver.java
tnleeuw: Inderdaad 😬

PR #7255, opened 2024-08-06
Remove (almost) all instances of getLogPrefix() from logging/exception messages.

### LINKED ISSUE #7202: Remove 'getLogPrefix()' in SenderBase

Right now for every log message (in combination with `getLogPrefix()`) we have a useless compute of the classname which is already part of the log message... (pipe/sender-name is part of the log context which is also available).

### Describe the Feature

Less compute, cleaner code.

Should start with SenderBase (ISender).

No, community resolution only

### Anything else?

_No response_

thomaspj10:
Closes #7202 

I have removed almost all instances of `getLogPrefix()` from logging and exceptions with the exception of the following:

- There are instances where the prefix contains additional information and doesn't use reflection to get the classname. For example:
```java
	public String getLogPrefix() {
		return "WebServiceSender ["+getName()+"] to ["+getPhysicalDestinationName()+"] ";
	}
```

- Where the prefix is the entire error message. For example:
```java
throw new SenderException(getLogPrefix(), e);
```


core/src/main/java/org/frankframework/http/HttpSender.java
```diff
@@ -209,12 +209,12 @@ protected HttpRequestBase getMethod(URI uri, Message message, ParameterValueList
 			case GET:
 				if (parameters!=null) {
 					queryParametersAppended = appendParameters(queryParametersAppended,relativePath,parameters);
-					if (log.isDebugEnabled()) log.debug("{}path after appending of parameters [{}]", getLogPrefix(), relativePath);
+					if (log.isDebugEnabled()) log.debug("path after appending of parameters [{}]", relativePath);
```
nielsm5, line 212:
```suggestion
					log.debug("path after appending of parameters [{}]", relativePath);
```

core/src/main/java/org/frankframework/http/HttpSender.java
```diff
@@ -209,12 +209,12 @@ protected HttpRequestBase getMethod(URI uri, Message message, ParameterValueList
...
 				}
 
 				HttpGet getMethod = new HttpGet(relativePath+(parameters==null && BooleanUtils.isTrue(getTreatInputMessageAsParameters()) && !Message.isEmpty(message)? message.asString():""));
 
-				if (log.isDebugEnabled()) log.debug("{}HttpSender constructed GET-method [{}]", getLogPrefix(), getMethod.getURI().getQuery());
+				if (log.isDebugEnabled()) log.debug("HttpSender constructed GET-method [{}]", getMethod.getURI().getQuery());
```
nielsm5, line 217:
```suggestion
				log.debug("HttpSender constructed GET-method [{}]", () -> getMethod.getURI().getQuery());
```

core/src/main/java/org/frankframework/http/WebServiceListener.java
```diff
@@ -189,7 +189,7 @@ public Message processRequest(Message message, PipeLineSession session) throws L
 
 		if (isSoap()) {
 			try {
-				if (log.isDebugEnabled()) log.debug("{}received SOAPMSG [{}]", getLogPrefix(), message);
+				log.debug("listening on [{}] received SOAPMSG [{}]", this::getPhysicalDestinationName, () -> message);
```
nielsm5, line 192: 🚀

core/src/main/java/org/frankframework/http/HttpSender.java
thomaspj10: There are roughly 350 occurences of such an if statement. Is it okay if I remove them in another PR, so the scope of this PR can stay relatively small?

nielsm5: They should only be removed when using a lambda or when it's reference directly to a field (so not a method call).

core/src/main/java/org/frankframework/jms/JMSFacade.java
```diff
@@ -336,7 +331,7 @@ public void close() {
 					log.trace("Closing messaging source - will synchronize (lock) on {}", messagingSource::toString);
 					messagingSource.close();
 				} catch (IbisException e) {
-					log.warn("{} caught exception closing messaging source", (Supplier<?>) this::getLogPrefix, e);
+					log.warn("{} caught exception closing messaging source", e);
```
nielsm5, line 334:
```suggestion
					log.warn("caught exception closing messaging source", e);
```

core/src/main/java/org/frankframework/jms/PullingJmsListener.java
```diff
@@ -214,7 +214,7 @@ public RawMessageWrapper<Message> getRawMessage(@Nonnull Map<String, Object> thr
 	public RawMessageWrapper<Message> getRawMessage(String correlationId, Map<String,Object> threadContext) throws ListenerException, TimeoutException {
 		RawMessageWrapper<Message> msg = getRawMessageFromDestination(correlationId, threadContext);
 		if (msg==null) {
-			throw new TimeoutException(getLogPrefix()+" timed out waiting for message with correlationId ["+correlationId+"]");
+			throw new TimeoutException(" timed out waiting for message with correlationId ["+correlationId+"]");
```
nielsm5, line 217: Nu klopt de zin niet meer.

core/src/main/java/org/frankframework/jms/PushingJmsListener.java
```diff
@@ -100,13 +100,13 @@ public class PushingJmsListener extends JmsListenerBase implements IPortConnecte
 	public void configure() throws ConfigurationException {
 		super.configure();
 		if (jmsConnector==null) {
-			throw new ConfigurationException(getLogPrefix()+" has no jmsConnector. It should be configured via springContext.xml");
+			throw new ConfigurationException(" has no jmsConnector. It should be configured via springContext.xml");
```
nielsm5, line 103: Hier ook.
```suggestion
			throw new ConfigurationException("no jmsConnector found. It should be wired/configured via Spring");
```

core/src/main/java/org/frankframework/jms/PushingJmsListener.java
```diff
@@ -221,10 +221,10 @@ public int getDeliveryCount(RawMessageWrapper<Message> rawMessage) {
 			if (log.isDebugEnabled()) log.debug("determined delivery count [{}]", value);
 			return value;
 		} catch (NumberFormatException nfe) {
-			if (log.isDebugEnabled()) log.debug("{}NumberFormatException in determination of DeliveryCount", getLogPrefix());
+			if (log.isDebugEnabled()) log.debug("NumberFormatException in determination of DeliveryCount");
```
nielsm5, line 224:
```suggestion
			log.debug("NumberFormatException in determination of DeliveryCount");
```

core/src/main/java/org/frankframework/monitoring/Monitor.java
nielsm5, line 1: It's best to revert changes made in this package (monitoring) for now. These messages /events are not based on runtime-messages. These other changes will need to eventually happen but I think for now it's best to scope it to senders.

nielsm5 (changes requested):

core/src/main/java/org/frankframework/senders/JavascriptSender.java
```diff
@@ -138,7 +138,7 @@ public void open() throws SenderException {
 				pvl = getParameterList().getValues(message, session);
 			}
 		} catch (ParameterException e) {
-			throw new SenderException(getLogPrefix() + " exception extracting parameters", e);
+			throw new SenderException(" exception extracting parameters", e);
```
nielsm5, line 141:
```suggestion
			throw new SenderException("unable to extract parameters", e);
```

core/src/main/java/org/frankframework/senders/ParallelSenders.java
```diff
@@ -74,7 +74,7 @@ public SenderResult doSendMessage(Message message, PipeLineSession session) thro
 				message.preserve();
 			}
 		} catch (IOException e) {
-			throw new SenderException(getLogPrefix() + " could not preserve input message", e);
+			throw new SenderException(" could not preserve input message", e);
```
nielsm5, line 77:
```suggestion
			throw new SenderException("could not preserve input message", e);
```

core/src/main/java/org/frankframework/senders/XsltSender.java
```diff
@@ -117,7 +117,7 @@ public void configure() throws ConfigurationException {
 			transformerPool = TransformerPool.configureTransformer0(this, getNamespaceDefs(), getXpathExpression(), getStyleSheetName(), getOutputType(), !omitXmlDeclaration, getParameterList(), getXsltVersion());
 		}
 		else if(StringUtils.isEmpty(getStyleSheetNameSessionKey())) {
-			throw new ConfigurationException(getLogPrefix()+" one of xpathExpression, styleSheetName or styleSheetNameSessionKey must be specified");
+			throw new ConfigurationException(" one of xpathExpression, styleSheetName or styleSheetNameSessionKey must be specified");
```
nielsm5, line 120:
```suggestion
			throw new ConfigurationException("one of xpathExpression, styleSheetName or styleSheetNameSessionKey must be specified");
```

nielsm5 (changes requested):
Probeer echt alleen de Senders over te houden. En die try-with-resources closableThreadContext moet echt in de SenderPipe anders gaat dit niet goed werken 😮

thomaspj10:
The commit above can be reverted to use as a starting point for when we want to extend the removal of `getLogPrefix()` to other parts of the codebase.

nielsm5 (approved):

nielsm5 (approved):

PR #7272, opened 2024-08-07
Add an example to the DirectoryListener and fix an incorrect reference.

### LINKED ISSUE #7219: MessageType in FileSystemListener schould be a enum (and contents should be content)
### Current Behavior

The documentation of the directory listener is not very clear

### Expected Behavior

Possible improvements
- Make the message Type an enum instead of a string.
- The name MessageType is not very descriptive for what it does. 
- The meaning of content and contents is very clos to each other but in my opinion it should be content of filecontent
- example in the frankdoc would be welcome
### Environment Information

ff 8.2.0

### Steps To Reproduce

https://frankdoc.frankframework.org/#/Listeners/DirectoryListener

### Configuration

_No response_

### Input

_No response_

_No response_

_No response_

### Relevant Log Output

_No response_

### Anything else?

_No response_

thomaspj10:
Closes #7219 

The usage of a DirectoryListener is vague at the moment. I added a clear example to the documentation. Read #7219 on why not every point is implemented.


filesystem/src/main/java/org/frankframework/receivers/DirectoryListener.java
```diff
@@ -21,6 +21,25 @@
 import org.frankframework.filesystem.FileSystemListener;
 import org.frankframework.filesystem.LocalFileSystem;
 
+/**
+ * Listener that looks in a FileSystem for files. <code>messageType</code> dictates what information of the file is passed in the message.
```
nielsm5, line 25:
```suggestion
 * Listener that looks for files in a LocalFileSystem. The DirectoryListener keeps track of the file process by storing it in different folders. The application may create the folders if you (a) set the <code>root</code> attribute and (b) set the attribute <code>createFolders</code> to true.
 
 The attribute <code>messageType</code> dictates what information of the file is passed to the pipeline. This may be the name, canonical path, the entire file, or the file's metadata.
```

nielsm5 (commented):
Issue #7219 talks about the possibility to turn messageType into an enum, I think that would simplify the use of this listener. Both functionally for the Frank!Developers as well as technically (see below code block).


```
			if (StringUtils.isEmpty(getMessageType()) || getMessageType().equalsIgnoreCase("name")) {
				return new Message(getFileSystem().getName(file));
			}
			if (StringUtils.isEmpty(getMessageType()) || getMessageType().equalsIgnoreCase("path")) {
				return new Message(getFileSystem().getCanonicalName(file));
			}
			if (getMessageType().equalsIgnoreCase("contents")) {
				return getFileSystem().readFile(file, getCharset());
			}
			if (getMessageType().equalsIgnoreCase("info")) {
				return new Message(FileSystemUtils.getFileInfo(getFileSystem(), file, getOutputFormat()));
			}
```

thomaspj10:
I didn't like my old implementation of making the `methodType` an Enum, but I think I managed to come up with a nice solution after all. The interface `IMessageType` is required because there are actually two different Enum interfaces required for DirectoryListener (and a few others) and MailListener. This way they both have a different enum, while still working correctly with their parent class.

filesystem/src/main/java/org/frankframework/filesystem/FileSystemListener.java
```diff
@@ -304,28 +317,25 @@ public Message extractMessage(@Nonnull RawMessageWrapper<F> rawMessage, @Nonnull
...
+				case "ATTRIBUTE", "HEADER" -> {
+					Map<String,Object> attributes = getFileSystem().getAdditionalFileProperties(file);
+					if (attributes != null) {
+						Object result = attributes.get(getSearchKey());
+						if (result != null) {
+							yield Message.asMessage(result);
+						}
+					}
+					log.warn("no attribute [{}] found for file [{}]", getMessageType(), getFileSystem().getName(file));
+					yield null;
```
nielsm5, line 335: Dit mag er uit hoor. Slaat nergens op om een attribute ergens uit te halen dat in `INFO` zit. Dan moet men maar een xpath oid doen.

nielsm5: Ik zie dat je dan wel info in het mail fs moet zetten. Ik weet niet of daar al tests voor zijn. Ik ga er van uit dat `FileSystemUtils.getFileInfo` ook gewoon werkt voor het mailfilesystem. (En daar zitten de 'additionalProperties' in.)

nielsm5: De releases.md heb ik in https://github.com/frankframework/frankframework/pull/7290 aangemaakt. Maar dat is nog niet gemerged (dus je hebt er nog niet veel aan).

thomaspj10: Dus nog eventjes wachten tot jouw issue op master staat

filesystem/src/main/java/org/frankframework/receivers/ExchangeMailListener.java
```diff
@@ -30,7 +30,7 @@
 import org.frankframework.filesystem.MailListener;
 
 /**
- * Microsoft Exchange Implementation of a {@link MailListener}.
+ * Microsoft Exchange Implementation of a {@link ImapListener}.
```
tnleeuw, line 33: This doesn't look correct?

test/src/main/configurations/FileHandling/Configuration_MFH.xml
```diff
@@ -12,7 +12,7 @@
 				minStableTime ="${mfh.flow1.minstabletime}"
 				fileTimeSensitive="true"
 				name="MFHIn"
-				messageType="info"
+				messageType="INFO"
```
tnleeuw, line 15: If using an enum means the value now has to be uppercase then that is another backwards incompatible change.

But doesn't the configuration parser automatically parse enums case-insensitive?

(Backwards incompatible changes shouldn't be in minor releases, right?)

filesystem/src/main/java/org/frankframework/receivers/ExchangeMailListener.java
thomaspj10: Clicking on the link results in a "Error loading Frank!Element" page (https://frankdoc.frankframework.org/#/All/ExchangeMailListener). But I see that it is not entirely true that this pipe is an implementation of an ImapListener.

tnleeuw: It is indeed not an ImapListener. The error is because no Frank!Doc is generated for the MailListener class, but since it's a `{@link ...}` there is still a link.

I think that it doesn't make sense for the Frank!Doc to have documentation pages for abstract classes that Frank! developers cannot use in their configurations, so it is probably best to just rephrase this.

```suggestion
 * Microsoft Exchange Implementation of a mail filesystem.
```

test/src/main/configurations/FileHandling/Configuration_MFH.xml
thomaspj10: I tested a lowercase enum locally and it seems to work correctly, so that is backwards compatible. The only breaking change is the one mentioned above where I removed the posibility to get attributes or headers based on a search key.

tnleeuw: Thanks, I wanted to make sure of this.

The one change mentioned above that is breaking, I think that it something we can get away with in practice because I don't think that anyone actually uses any of the other possible values.

filesystem/src/main/java/org/frankframework/receivers/ExchangeMailListener.java
thomaspj10: Agreed. The suggestion seems good to me

filesystem/src/main/java/org/frankframework/filesystem/FileSystemListener.java
```diff
@@ -529,14 +526,6 @@ public void setOverwrite(boolean overwrite) {
...
 	}
 
-	/**
-	 * Determines the contents of the message that is sent to the pipeline. Can be 'name', for the filename, 'path', for the full file path, 'contents' for the contents of the file, 'info' for file information. For any other value, the attributes of the file are searched and used
-	 * @ff.default path
-	 */
-	public void setMessageType(String messageType) {
-		this.messageType = messageType;
-	}
-
```
nielsm5, line 539: Deze moet je wel laten staan (met enum param) zodat de JavaDoc in de FrankDoc komt :)

filesystem/src/main/java/org/frankframework/receivers/FtpFileSystemListener.java
```diff
@@ -27,4 +27,12 @@ public FtpFileSystem createFileSystem() {
 		return FtpFileSystemDelegator.super.createFileSystem();
 	}
 
+	/**
+	 * Determines the contents of the message that is sent to the pipeline. The value of the attribute matching the searchKey is returned when using <code>ATTRIBUTE</code>
+	 * @ff.default PATH
+	 */
+	public void setMessageType(MessageType messageType) {
```
nielsm5, line 34: Hier kan je dan de ReferTo Gebruiken.

filesystem/src/main/java/org/frankframework/receivers/Samba2Listener.java
```diff
@@ -27,4 +27,12 @@ protected Samba2FileSystem createFileSystem() {
 		return new Samba2FileSystem();
 	}
 
+	/**
+	 * Determines the contents of the message that is sent to the pipeline. The value of the attribute matching the searchKey is returned when using <code>ATTRIBUTE</code>
+	 * @ff.default PATH
+	 */
+	public void setMessageType(MessageType messageType) {
```
nielsm5, line 34: Hier kan je dan de ReferTo Gebruiken.

filesystem/src/main/java/org/frankframework/receivers/SftpFileSystemListener.java
```diff
@@ -27,4 +27,12 @@ public SftpFileSystem createFileSystem() {
 		return SftpFileSystemDelegator.super.createFileSystem();
 	}
 
+	/**
+	 * Determines the contents of the message that is sent to the pipeline. The value of the attribute matching the searchKey is returned when using <code>ATTRIBUTE</code>
+	 * @ff.default PATH
+	 */
+	public void setMessageType(MessageType messageType) {
```
nielsm5, line 34: Hier kan je dan de ReferTo Gebruiken.

nielsm5 (changes requested):
Vergeet de RELEASES.MD niet 😉

filesystem/src/main/java/org/frankframework/filesystem/FileSystemListener.java
thomaspj10: Ik heb deze verwijderd, omdat deze class geërfd wordt door twee soorten listeners die beide een andere enum nodig hebben (MailListener vs Directory, (s)Ftp en Samba Listeners). Deze functie laten staan kan niet, omdat de children de signature van de functie moeten aanpassen (omdat het 2 verschillende enums zijn). Vandaar dat ik deze functie naar de children heb gekopieerd.

nielsm5 (approved):

PR #7283, opened 2024-08-08
Add an unsafe annotation for unsafe attributes.

### LINKED ISSUE #2727: Give configurationwarning on unsafe attribute values

**Describe the issue**
I would like to have warnings in the console if unsafe attributes are used. With unsafe attributes, I mean attributes like:
- allowSelfSignedCertificates=true
- verifyHostname=false
- password=not empty

**Reporter**
Ricardo

**To Reproduce**

**Screenshots**

**Configuration**
```
```

**Input**
```
```

**Environment**
```
```

**Additional Environment**

**Additional Context**

thomaspj10:
Closes #2727 

Will be used for the following:
- A warning in the console when an unsafe attribute is used.
- A warning in the frankdoc to alert them that the use of this attribute might be unsafe.

The annotation sadly isn't as sophisticated as I would like it to be. A lambda in the annotation which returns true when the value is actually unsafe would be my prefered solution, but that is not possible because members of the annotation must be primatives. This means that this annotation is only really useful for boolean values with a safe default. Checking the length of a password is currently not possible for example, because it would mark the usage of the attribute as unsafe while that may not be the case.

The current non exhaustive list of unsafe attributes are the following:
- verifyHostname 
- allowSelfSignedCertificates
- ignoreCertificateExpiredException

An Frankdoc issue can be made to further implement this into the Frankdoc when this is merged.

### DIFF
--- a/cmis/src/main/java/org/frankframework/extensions/cmis/CmisSender.java
+++ b/cmis/src/main/java/org/frankframework/extensions/cmis/CmisSender.java
@@ -1144,7 +1144,7 @@ public class CmisSender extends SenderWithParametersBase implements HasKeystore,
 		return sessionBuilder.getTrustManagerAlgorithm();
 	}
 
-	@Unsafe(description = "verifyHostname is unsafe and should not be used in a production environment.")
+	@Unsafe
 	@Override
 	public void setVerifyHostname(boolean verifyHostname) {
 		sessionBuilder.setVerifyHostname(verifyHostname);
@@ -1154,7 +1154,7 @@ public class CmisSender extends SenderWithParametersBase implements HasKeystore,
 		return sessionBuilder.isVerifyHostname();
 	}
 
-	@Unsafe(description = "allowSelfSignedCertificates is unsafe and should not be used in a production environment.")
+	@Unsafe
 	@Override
 	public void setAllowSelfSignedCertificates(boolean testModeNoCertificatorCheck) {
 		sessionBuilder.setAllowSelfSignedCertificates(testModeNoCertificatorCheck);
@@ -1164,7 +1164,7 @@ public class CmisSender extends SenderWithParametersBase implements HasKeystore,
 		return sessionBuilder.isAllowSelfSignedCertificates();
 	}
 
-	@Unsafe(description = "ignoreCertificateExpiredException is unsafe and should not be used in a production environment.")
+	@Unsafe
 	@Override
 	public void setIgnoreCertificateExpiredException(boolean ignoreCertificateExpiredException) {
 		sessionBuilder.setIgnoreCertificateExpiredException(ignoreCertificateExpiredException);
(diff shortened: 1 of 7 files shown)


cmis/src/main/java/org/frankframework/extensions/cmis/CmisSender.java
```diff
@@ -1143,6 +1144,7 @@ public String getTrustManagerAlgorithm() {
 		return sessionBuilder.getTrustManagerAlgorithm();
 	}
 
+	@Unsafe(description = "verifyHostname is unsafe and should not be used in a production environment.")
```
nielsm5, line 1147: The text that in the annotation is computable. Perhaps it's better to have an empty annotation and generate the text.

core/src/main/java/org/frankframework/configuration/digester/ValidateAttributeRule.java
```diff
@@ -61,36 +62,41 @@ protected void handleAttribute(String name, String value, Map<String, String> at
...
-			} else { //Only check for default values for non-property values
-				Method readMethod = pd.getReadMethod();
-				if(readMethod != null) { //And if a read method (getter) exists
-					checkTypeCompatibility(readMethod, name, value, attributes);
-				}
+			return;
+		}
+
+		checkDeprecationAndConfigurationWarning(name, value, m); //check if the setter or enum value is deprecated
+		checkUnsafe(m);
```
nielsm5, line 76:
```suggestion
		checkIfMethodIsMarkedAsUnsafe(m, name);
```

core/src/main/java/org/frankframework/configuration/digester/ValidateAttributeRule.java
```diff
@@ -148,6 +154,16 @@ private void checkDeprecationAndConfigurationWarning(String name, String value,
...
 
+	private void checkUnsafe(Method setterMethod) {
+		Unsafe unsafe = AnnotationUtils.findAnnotation(setterMethod, Unsafe.class);
+
+		if (unsafe == null) {
+			return;
+		}
+
+		addSuppressibleWarning(unsafe.description(), SuppressKeys.UNSAFE_ATTRIBUTE_SUPPRESS_KEY);
+	}
```
nielsm5, line 165:
```suggestion
	private void checkIfMethodIsMarkedAsUnsafe(Method setterMethod, String attributeName) {
		Unsafe unsafe = AnnotationUtils.findAnnotation(setterMethod, Unsafe.class);

		if (unsafe == null) {
			return;
		}

		String warning = "[" + attributeName + "] is unsafe and should not be used in a production environment.";
		addSuppressibleWarning(warning, SuppressKeys.UNSAFE_ATTRIBUTE_SUPPRESS_KEY);
	}
```

nielsm5 (changes requested):

cmis/src/main/java/org/frankframework/extensions/cmis/CmisSender.java
thomaspj10: I choose for a description, because there could be use cases where it is nice to have a different text, but I guess for now it's not really necessary.

commons/src/main/java/org/frankframework/doc/Unsafe.java
```diff
@@ -0,0 +1,32 @@
+/*
+   Copyright 2022 WeAreFrank!
```
nielsm5, line 2:
```suggestion
   Copyright 2024 WeAreFrank!
```

nielsm5 (approved):

PR #7358, opened 2024-08-20
Split status page sections into child components

Matthbo:


philipsens (approved):
Code looks nicely formatted now!
I'm glad that Niels showed me how to review individual commits. 
The commits were separated in a good way so I could review it without problems. Thank you.

PR #7365, opened 2024-08-21
Use more configurable lifecycles

nielsm5:


nielsm5:
Sonar seems to be a bit stupid here, claiming that my interface implementation is duplicate code.
I could create either a new interface of an abstract class to hide these methods but I'm not sure if that will improve anything.

Issue #7401, opened 2024-08-27
ibis4example shows error on IAF_Util flow generation


jjansenvr:
### Current Behavior

![image](https://github.com/user-attachments/assets/8b4a6f2c-d955-45a5-8322-7623e2504123)

### Expected Behavior

flow

### Environment Information

FF! 8.3.0-20240827.042327: Frank2Example4

jjansenvr:
the error is caused by the Util config not having adapters and especially forwards

line 34-41 in the configuration2mermaid.xsl need to be encapsulated with a If checking for forwards

something like this 

```
<xsl:if test="count($preproccessedConfiguration//forward) > 0)">
	<xsl:text>	linkStyle </xsl:text>
	<xsl:value-of select="$preproccessedConfiguration//forward/(position() - 1)" separator=","/>
	<xsl:text> stroke:#8bc34a,stroke-width:3px,fill:none;</xsl:text>
	<xsl:text>&#10;</xsl:text>
	<!-- The code below gives back the preprocessed configuration This is for testing purposes. To test, also change method(line 3) to xml instead of text-->
	<!-- <xsl:copy-of select="$preproccessedConfiguration"/>-->
</xsl:if> 
```

lenardvandermaas:
#7471 will fix this

PR #7471, opened 2024-09-10
fix mermaid xsl issues for config and adapter

lenardvandermaas:
this PR will fix the mermaid issue when a config has no forwards, and will also fix an issue when exits have their state attribute capitalized


lenardvandermaas:
fixes #7401

nielsm5 (approved):

PR #7477, opened 2024-09-11
Update code style documentation

evandongen:
Update coding guidelines

### DIFF
--- a/AdditionalCodingGuidelines.md
+++ b/AdditionalCodingGuidelines.md
@@ -111,7 +111,7 @@ final String x = (condition) ? "value" : null;
 	when not using an IDE.
 9. Java Optionals.
 
-	You can use Java `Optional` to indicatej that the return value of a method can be `null`.
+	You can use Java `Optional` to indicate that the return value of a method can be `null`.
 	It is not custom to use this for parameters, only for return-values.
 	However I find that code is not necessarily more readable when using `Optional` so use
 	them at your own discretion, and see if you find the code calling your methods becomes
--- a/AdditionalCodingGuidelines.md
+++ b/AdditionalCodingGuidelines.md
@@ -103,10 +103,15 @@ final String x = (condition) ? "value" : null;
 	 - The idea is to put focus the actual operations your code does instead of burying that
 		in the ceremony around it.
 
+8. Usage of 'var'
+	Since Java 9, the usage of `var` is supported. This is still strongly typed, but notation is less verbose:
+	`String string = new String()` can become `var string = new String()`.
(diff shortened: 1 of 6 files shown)


nielsm5 (approved):

AdditionalCodingGuidelines.md
```diff
@@ -103,8 +103,13 @@ final String x = (condition) ? "value" : null;
 	 - The idea is to put focus the actual operations your code does instead of burying that
 		in the ceremony around it.
 
-
-8. Java Optionals.
+8. Usage of 'var'
+	Since Java 9, the usage of `var` is supported. This is still strongly typed, but notation is less verbose:
+	`String string = new String()` can become `var string = new String()`.
+	This can become less readable when using the return type of a method. To keep code as readable as possible, we only use 
+	this 'locally'. This means, a local value. Try to avoid using var for method return types, this makes code less readable
```
tnleeuw, line 110: This is not clear. It reads as if the method return type can be specified as `var`, which is not true. And a local variable can also be assigned a method return value.

```suggestion
	this when the type is clear, such as a direct assignment of a string or scalar value or constructor. Try to avoid using `var` for variables assigned from method return values, this makes code less readable
```

AdditionalCodingGuidelines.md
```diff
@@ -317,7 +322,7 @@ fun registerApplicationModules(modules: List<String>) =
 Since javadoc is used for generating documentation in the Frank!Doc, it's important to provide usable information in the class for Frank Developers.
 
 * Please provide documentation in the javadoc if there's been a (breaking) change and offer an example in the documentation. See [FixedResultPipe.java](https://github.com/frankframework/frankframework/blob/master/core/src/main/java/org/frankframework/pipes/FixedResultPipe.java) for instance.  
-* When using Code examples, please use `<pre>{@code ... }</pre>` blocks.
+* When using multiline Code examples, please use `<pre>{@code ... }</pre>` blocks.
 * When referring to classes, please use `{@link ClassName}`
-* When referring to a value, please use `{@value #VALUE}`
-* When referring to a literal, please use `{@literal null}`
+* When referring to a class value, please use `{@value #VALUE}`
+* When referring to a value (of a parameter, variable, etc), please use `{@literal null}`
```
tnleeuw, line 328: Perhaps you can add a clarification that the Frank!Doc treats this differently than regular JavaDoc so that the literal will stand out as a code literal?

tnleeuw (changes requested):

tnleeuw (approved):

PR #7480, opened 2024-09-12
Only pass an errorMessage if the http status code is an error

evandongen:


core/src/main/java/org/frankframework/http/HttpSender.java
```diff
@@ -65,32 +71,9 @@
...
- * <pre>
- *   param_name=param_value
- *   another_param_name=another_param_value
- * </pre>
- *
- * Note:
- * When used as MTOM sender and MTOM receiver doesn't support Content-Transfer-Encoding "base64", messages without line feeds will give an error.
- * This can be fixed by setting the Content-Transfer-Encoding in the MTOM sender.
- * </p>
+ * @inheritDoc
```
nielsm5, line 76:
```suggestion
 * {@inheritDoc}
```

nielsm5 (approved):

PR #7579, opened 2024-09-30
Validate pipe forwards during configure phase

nielsm5:
Based on #7562, validation should happen during the configure phase not during the config xml phase.

### DIFF
--- a/core/src/main/java/org/frankframework/pipes/AbstractPipe.java
+++ b/core/src/main/java/org/frankframework/pipes/AbstractPipe.java
@@ -22,7 +22,15 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import jakarta.annotation.Nonnull;
+import jakarta.annotation.Nullable;
+
+import lombok.Getter;
+import lombok.Setter;
 import org.apache.commons.lang3.StringUtils;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/frankframework/pipes/AbstractPipe.java
```diff
@@ -151,6 +159,48 @@ public void configure() throws ConfigurationException {
 		}
 	}
 
+	private void configureForward(PipeForward forward) throws ConfigurationException {
```
evandongen, line 162: Deze throwed niet meer.

core/src/main/java/org/frankframework/pipes/AbstractPipe.java
```diff
@@ -125,13 +128,18 @@ public abstract class AbstractPipe extends TransactionAttributes implements IPip
...
 	@Override
 	public void configure() throws ConfigurationException {
 		super.configure();
 		if(StringUtils.isNotEmpty(getName()) && getName().contains("/")) {
 			throw new ConfigurationException("It is not allowed to have '/' in pipe name ["+getName()+"]");
 		}
+
+		for(PipeForward forward : registeredForwards) {
+			configureForward(forward);
+		}
```
evandongen, line 141:
```suggestion
		registeredForwards.forEach(this::configureForward);
```

evandongen (approved):

PR #7586, opened 2024-10-01
Fix minimum page height & hide scrollbars on testing & custom views

Matthbo:
- Set minimum page height to 100vh as to fill the page
- Remove scrollbars on iframe pages like custom views, larva & ladybug

### DIFF
--- a/console/frontend/src/main/frontend/src/assets/css/style.css
+++ b/console/frontend/src/main/frontend/src/assets/css/style.css
@@ -3503,7 +3503,8 @@ body {
   font-size: 13px;
   color: #676a6c;
   overflow-x: hidden;
-  overflow-y: scroll !important;
+  /*overflow-y: scroll !important;*/
+  overflow-y: scroll;
 }
 html,
 body {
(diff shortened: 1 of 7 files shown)


console/frontend/src/main/frontend/src/assets/css/style.css
```diff
@@ -3503,7 +3503,8 @@ body {
   font-size: 13px;
   color: #676a6c;
   overflow-x: hidden;
-  overflow-y: scroll !important;
+  /*overflow-y: scroll !important;*/
```
MatthijsSmets, line 3506: I think you forgot to remove this

MatthijsSmets (approved):

console/frontend/src/main/frontend/src/assets/css/style.css
Matthbo: No I kept it to keep the original file sort of in tact, I don't like making changes in this mess hence why I also have a bootrap-fix.css

philipsens (approved):

PR #7612, opened 2024-10-04
Adapter can do without receiver when called from FrankSender

mhdirkse:


nielsm5:
Yes this is possible, but that doesn't mean we should endorse this (yet).

nielsm5:
Deze parkeren we even.

PR #7638, opened 2024-10-08
Upgrade frank-doc version and prep for release

nielsm5:


pom.xml
evandongen, line 1: 🎊

evandongen (approved):

PR #7667, opened 2024-10-10
Fix NPE when Ladybug rerun cannot find config

jacodg:
- Fix NPE when Ladybug rerun cannot find config
- Stop application when an error occurs during startup
- Change default property locations
- Allow XML storage to be stored on disk even when using DatabaseStorage
- Change exception messages when adapter or config cannot be find while in rerun
- Change whitespace in comment and solve Sonar issue

### DIFF
--- a/console/war/src/main/java/org/frankframework/console/runner/ConsoleStandaloneInitializer.java
+++ b/console/war/src/main/java/org/frankframework/console/runner/ConsoleStandaloneInitializer.java
@@ -19,14 +19,16 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
-import org.apache.tomcat.websocket.server.WsContextListener;
-import org.springframework.boot.SpringApplication;
-import org.springframework.boot.WebApplicationType;
-import org.springframework.boot.web.servlet.ServletContextInitializer;
-
 import jakarta.servlet.ServletContext;
 import jakarta.servlet.ServletContextEvent;
 import jakarta.servlet.ServletException;
+
+import org.apache.tomcat.websocket.server.WsContextListener;
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.WebApplicationType;
+import org.springframework.boot.context.event.ApplicationFailedEvent;
+import org.springframework.boot.web.servlet.ServletContextInitializer;
+import org.springframework.context.ApplicationListener;
 
 /**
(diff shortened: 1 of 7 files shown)


console/war/src/main/java/org/frankframework/console/runner/ConsoleStandaloneInitializer.java
```diff
@@ -49,15 +51,26 @@ public void onStartup(ServletContext servletContext) throws ServletException {
...
+		app.addListeners(new FailedInitializationMonitor());
 		Set<String> set = new HashSet<>();
 		set.add("SpringBootContext.xml");
 		app.setSources(set);
 		app.addPrimarySources(List.of(WsSciWrapper.class));
+
 		app.run(args);
 	}
 
+	/**
```
tnleeuw, line 65: I really don't like this.
This means that you break all  applications deployed to this Tomcat, not just your own.

Generally this is considered a really bad practice.

ladybug/debugger/src/main/java/org/frankframework/ladybug/LadybugDebugger.java
```diff
@@ -97,13 +99,21 @@ private List<Adapter> getRegisteredAdapters() {
...
 			String configName = name.substring(0, i);
 			Configuration config = ibisManager.getConfiguration(configName);
+			if (config == null) {
+				throw new IllegalStateException("Configuration ["+configName+"] does not exist.");
+			}
 			String adapterName = name.substring(i + 1);
-			return config.getRegisteredAdapter(adapterName);
+			Adapter adapter = config.getRegisteredAdapter(adapterName);
+			if (adapter == null) {
+				throw new IllegalStateException("Adapter ["+adapterName+"] does not exist in configuration ["+configName+"].");
```
tnleeuw, line 114: I think `IllegalArgumentException` would be a bit more appropriate?

ladybug/debugger/src/main/java/org/frankframework/ladybug/LadybugDebugger.java
```diff
@@ -112,7 +122,7 @@ private Adapter getRegisteredAdapter(String name) {
 				return adapter;
 			}
 		}
-		return null;
+		throw new IllegalStateException("Adapter ["+name+"] not found.");
```
tnleeuw, line 125: I think `IllegalArgumentException` would be a bit more appropriate?

tnleeuw (approved):

PR #7669, opened 2024-10-10
Minor optimisations 

tnleeuw:


commons/src/main/java/org/frankframework/util/StringUtil.java
```diff
@@ -266,7 +267,7 @@ public static Stream<String> splitToStream(@Nullable final String input) {
 	 *
 	 * @param input the string to split, can be {@literal null}.
 	 * @param delim the delimiters to split the string by
-	 * @return a {@link List} of strings. An empty list if the input was {@literal null}.
+	 * @return a (modifiable) {@link List} of strings. An empty list if the input was {@literal null}.
```
tnleeuw, line 270: Making the result an unmodifiable list would make Sonar a bit happier, but sometimes is a bit of an extra hassle so I decided to leave this as it is and document it.

core/src/main/java/org/frankframework/jdbc/datasource/FrankResources.java
```diff
@@ -99,9 +101,9 @@ private DataSource loadDataSourceThroughDriver(Class<?> clazz, String url, Prope
 	}
 
 	private @Nullable Resource findResource(@Nonnull String name) {
-		List<String> parts = StringUtil.split(name, "/");
-		String prefix = parts.remove(0);
-		String resourceName = String.join("/", parts);
+		int slashPos = name.indexOf('/');
+		String prefix = name.substring(0, slashPos);
+		String resourceName = name.substring(slashPos + 1);
```
tnleeuw, line 106: This bit of code broke when I tried to make the result of `StringUtil.split()` unmodifiable.

In the end I reverted the change there, but decided to leave it in here because it should be slightly more optimal than doing a regex search and all the list manipulations.

larva/src/main/java/org/frankframework/larva/ScenarioRunner.java
```diff
@@ -81,7 +81,7 @@ public ScenarioRunner(LarvaTool larvaTool, IbisContext ibisContext, TestConfig c
 		this.multipleThreads = config.isMultiThreaded();
 
 		String blackListDirs = AppConstants.getInstance().getProperty("larva.parallel.blacklistDirs", "");
-		parallelBlacklistDirs = new ArrayList<>(StringUtil.split(blackListDirs, ","));
+		parallelBlacklistDirs = StringUtil.split(blackListDirs);
```
tnleeuw, line 84: The result is already a list, don't need to make a new list with it. And don't need to specify the default delimiter.

evandongen (approved):

nielsm5 (approved):

PR #7672, opened 2024-10-11
Prepare for next release cycle

nielsm5:


core/src/test/java/org/frankframework/receivers/ReceiverTest.java
```diff
@@ -971,7 +970,7 @@ public void testStartTimeout(SlowListenerBase listener) throws Exception {
 
 	@Test
 	public void testStopAdapterWhileReceiverIsStillStarting() throws Exception {
-		assumeFalse(TestAssertions.isTestRunningOnCI() || TestAssertions.isTestRunningOnGitHub(), "For unknown reasons this test is unreliable on Github and CI so only run locally for now until we have time to investigate");
+		assumeFalse(TestAssertions.isTestRunningWithSurefire() || TestAssertions.isTestRunningOnCI(), "For unknown reasons this test is unreliable on Github and CI so only run locally for now until we have time to investigate");
```
nielsm5, line 973: This test always succeeds when running directly in my IDE, but always fails when using SureFire.. not sure why. Very iffy test to debug 😢 .

tnleeuw: Same here. But until recently it did run with Surefire. Very very weird but I haven't yet made time to try find out why.

tnleeuw (approved):

PR #7714, opened 2024-10-18
Rename registerXyz digester methods with addXyz

### LINKED ISSUE #1601: Improve multiplicity detection in FrankDoc generation
Should fix #2519

**Reporter**
Gerrit

A change in multiplicity of config children is now implemented with a hack in digester-rules.xml and the digester. Should be replaced by something proper.

thomaspj10:
Closes #1601 

The ultimate goal is to replace the digester-rules, but we decided to gradually reduce the complexity of the current situation to hopefully get into a position where a migration to another system is more feasible and less complex.

This PR replaces all `registerXyz` with `addXyz`. `setXyz` is used for singular use and `addXyz` is used for plural use. A few dumb unnecessary things have also been removed.

There are a few other things, which should be replaced, but aren't changed in this PR, because we simply don't know what to with it yet:
- Remove `selfRegisterMethod`
- See what we can do with `registerTextMethod`

### DIFF
--- a/aspose/src/test/java/org/frankframework/extensions/aspose/pipe/PdfPipeTest.java
+++ b/aspose/src/test/java/org/frankframework/extensions/aspose/pipe/PdfPipeTest.java
@@ -536,7 +536,7 @@ public class PdfPipeTest extends PipeTestBase<PdfPipe> {
 	public void multiThreadedMailWithWordAttachment() throws Exception {
 		pipe.setName("multiThreadedmailWithWordAttachment");
 		pipe.setAction(DocumentAction.CONVERT);
-		pipe.registerForward(new PipeForward("success", "dummy"));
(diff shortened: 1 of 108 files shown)


evandongen:
> There are a few other things, which should be replaced, but aren't changed in this PR, because we simply don't know what to with it yet:
> * Remove selfRegisterMethod
> * See what we can do with registerTextMethod

Is there a anoter ticket so we won't forget this?

evandongen (approved):
Poe, aardig klusje!

nielsm5:
> * Remove selfRegisterMethod
> * See what we can do with registerTextMethod

Summier maar #7725, zodat we dit niet vergeten...

nielsm5 (approved):

PR #7716, opened 2024-10-18
Remove method overloading complexity

nielsm5:


tnleeuw (changes requested):
Nice improvement.
I think that if you also update the class DB2XMLWriter, you should be able to get rid off the overloaded `toString()` method in `SaxElementBuilder` altogether.

nielsm5:
> Nice improvement. I think that if you also update the class DB2XMLWriter, you should be able to get rid off the overloaded `toString()` method in `SaxElementBuilder` altogether.

Yea, at this point I was happy everything was still working. But it definitely requires a few more iterations!
Do you want me to continue in this branch or in a new pr?

tnleeuw:
> > Nice improvement. I think that if you also update the class DB2XMLWriter, you should be able to get rid off the overloaded `toString()` method in `SaxElementBuilder` altogether.
> 
> Yea, at this point I was happy everything was still working. But it definitely requires a few more iterations! Do you want me to continue in this branch or in a new pr?

I think that there's only one use left of the `SaxElementBuilder.toString()` so I think it's worth removing that, and the overloaded `toString()`, in this PR.
Other changes and iterations can come in a new PR.

tnleeuw (approved):

core/src/main/java/org/frankframework/documentbuilder/XmlDocumentBuilder.java
```diff
@@ -27,12 +28,8 @@ public class XmlDocumentBuilder extends XmlNodeBuilder implements IDocumentBuild
 	private Writer writer;
 	private SaxDocumentBuilder saxDocumentBuilder;
 
-	public XmlDocumentBuilder(String rootElement) throws SAXException {
-		this(rootElement, new StringWriter(), true);
-	}
-
 	public XmlDocumentBuilder(String rootElement, Writer writer, boolean prettyPrint) throws SAXException {
-		this(new SaxDocumentBuilder(null, writer), rootElement);
+		this(new SaxDocumentBuilder(null, new XmlWriter(writer), true), rootElement);
```
evandongen, line 32:
```suggestion
		this(new SaxDocumentBuilder(null, new XmlWriter(writer), prettyPrint), rootElement);
```

?

evandongen (approved):

PR #7723, opened 2024-10-21
Fix equals check in FS Listeners

nielsm5:
(cherry picked from commit ebeb401d2c5f8f4151e7a97a14858dd60219ed7a)

### DIFF
--- a/filesystem/src/main/java/org/frankframework/filesystem/FileSystemListener.java
+++ b/filesystem/src/main/java/org/frankframework/filesystem/FileSystemListener.java
@@ -357,10 +357,10 @@ public abstract class FileSystemListener<F, FS extends IBasicFileSystem<F>> impl
 			if (attributes!=null) {
 				messageProperties.putAll(attributes);
 			}
-			if (!"path".equals(getMessageType())) {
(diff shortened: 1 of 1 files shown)


evandongen (approved):

filesystem/src/main/java/org/frankframework/filesystem/FileSystemListener.java
```diff
@@ -357,10 +357,10 @@ public Message extractMessage(@Nonnull RawMessageWrapper<F> rawMessage, @Nonnull
 			if (attributes!=null) {
 				messageProperties.putAll(attributes);
 			}
-			if (!"path".equals(getMessageType())) {
+			if (getMessageType() != MessageType.PATH) {
```
tnleeuw, line 360: How was this not caught by our tests? 🤔

tnleeuw: Since this was not caught by tests apparently, I actually wonder if the right fix should be 
```suggestion
			if (MessageType.PATH.name().equals(getMessageType().name())) {
```

PR #7761, opened 2024-10-24
Use CID in test pipeline when present

evandongen:
Backport to 7.9 from https://github.com/frankframework/frankframework/pull/7535

### DIFF
--- a/test/src/main/resources/DeploymentSpecifics.properties
+++ b/test/src/main/resources/DeploymentSpecifics.properties
@@ -211,8 +211,8 @@ larva.adapter.active=true
 larva.adapter.writeToLog=true
 larva.adapter.writeToSystemOut=false
 larva.adapter.execute=
-larva.adapter.mail.active=false
-larva.adapter.scheduler.active=false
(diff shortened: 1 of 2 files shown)


test/src/main/resources/DeploymentSpecifics.properties
```diff
@@ -211,8 +211,8 @@ larva.adapter.active=true
 larva.adapter.writeToLog=true
 larva.adapter.writeToSystemOut=false
 larva.adapter.execute=
-larva.adapter.mail.active=true
-larva.adapter.scheduler.active=true
+larva.adapter.mail.active=false
+larva.adapter.scheduler.active=false
```
nielsm5, line 215:
```suggestion
larva.adapter.mail.active=true
larva.adapter.scheduler.active=true
```

nielsm5 (approved):

test/src/main/resources/DeploymentSpecifics.properties
evandongen: Hier twijfelde ik inderdaad over

PR #7771, opened 2024-10-25
Fix Docker and add NPM for Dependabot

philipsens:
Notable changes:

- Fixed the Docker directories Glob
- Changed the versioning of the Docker images → so it gets picked up by Dependabot
- Added NPM package ecosystem
- Removed `open-pull-requests-limit: 5` → that's the default
- Added dependencies from the private Nexus to ignore list → versions don't change anyway. (And if so, we could add credentials for the private repository later)
- Added `org.frankframework:frankframework-ear` to ignore list → Dependabot breaks on this dep because it wants to search the private Nexus (which it can't)
  I'm not sure why this only happens with this module, I don't see anything irregular.
- Added labels for the languages
- Added maintainers as reviewer → this just looks nice and we get a notification.

Other information:

Snyk bot doesn't do a lot, yet it claims to be better, it made less PR's and excluding Docker images. Snyk doesn't show the logs of the bot and can't be configures as granular as Dependabot, that why I went for that option.


philipsens:
Fixed in #7771

PR #7810, opened 2024-10-29
For the Ladybug debug tab use DatabaseLogStorage

mhdirkse:


nielsm5:
Also needs to be done in [ladybug/rerunner/src/main/resources/springTestToolStandalone.xml](https://github.com/frankframework/frankframework/blob/774f7e1c8a3b7650eb0dde828a0039d8d659d480/ladybug/rerunner/src/main/resources/springTestToolStandalone.xml#L49)

PR #7831, opened 2024-10-31
Fix js exception when no listener in receiver

Thermalight:
Added a check to prevent null pointer to listener when none are available in the adapter status page.

### DIFF
--- a/console/frontend/src/main/frontend/src/app/views/status/adapter-status/adapter-status.component.html
+++ b/console/frontend/src/main/frontend/src/app/views/status/adapter-status/adapter-status.component.html
@@ -219,11 +219,15 @@
                   [ngClass]="{ 'text-danger': receiver.state === 'error' }"
                 >
                   <td>{{ receiver.name }}</td>
-                  <td #listenerDestination [title]="receiver.listener.destination">
-                    {{ receiver.listener.class }} ({{
-                      receiver.listener.destination | truncate: 150 : listenerDestination
-                    }})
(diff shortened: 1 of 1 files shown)


Thermalight:
Should fix #7830 (hopefully)

console/frontend/src/main/frontend/src/app/views/status/adapter-status/adapter-status.component.html
```diff
@@ -219,11 +219,15 @@ <h4>Receivers</h4>
...
-                  </td>
+                  <ng-container *ngIf="receiver.listener && receiver.listener.destination; else noListener">
+                    <td #listenerDestination [title]="receiver.listener.destination">
+                      {{ receiver.listener.class }} ({{ receiver.listener.destination | truncate: 150 : listenerDestination }})
+                    </td>
+                  </ng-container>
+                  
+                  <ng-template #noListener>
+                    <td>No listener found</td>
+                  </ng-template>
```
Matthbo, line 230:
```suggestion
                  <td #listenerDestination 
                    *ngIf="receiver.listener && receiver.listener.destination; else noListener" 
                    [title]="receiver.listener.destination"
                  >
                    {{ receiver.listener.class }} ({{ receiver.listener.destination | truncate: 150 : listenerDestination }})
                  </td>
                  <ng-template #noListener>No listener found, but is required</ng-template>
```

philipsens (approved):

PR #7853, opened 2024-11-01
Make the FixedResultPipe backwards compatible again

nielsm5:
In #6987 we removed a deprecated attribute which breaks property substitution in classpath files.


core/src/test/java/org/frankframework/pipes/FixedResultPipeTest.java
```diff
@@ -283,6 +283,31 @@ public void testEmptyReturnString() throws Exception {
 		assertEquals("", res.getResult().asString());
 	}
 
+	@Test
+	public void substitudeVarsOldWithInvalidAttribute() throws Exception{
```
tnleeuw, line 287:
```suggestion
	public void substituteVarsOldWithInvalidAttribute() throws Exception{
```

core/src/test/java/org/frankframework/pipes/FixedResultPipeTest.java
```diff
@@ -283,6 +283,31 @@ public void testEmptyReturnString() throws Exception {
...
+		pipe.addParameter(param);
+
+		pipe.setReturnString("This is ${param}");
+		pipe.setUseOldSubstitutionStartDelimiter(true);
+		ConfigurationException ce = assertThrows(ConfigurationException.class, pipe::configure);
+		assertEquals("attribute [useOldSubstitutionStartDelimiter] may only be used in combination with attribute [filename]", ce.getMessage());
+	}
+
+	@Test
+	public void substitudeVarsOldWithFilename() throws Exception{
```
tnleeuw, line 299:
```suggestion
	public void substituteVarsOldWithFilename() throws Exception{
```

tnleeuw (approved):

PR #7899, opened 2024-11-11
Update credential factory configuration for JBoss and WildFly

philipsens:
Replace the credential factory class and properties in both JBoss and 
WildFly configurations to use the PropertyFileCredentialFactory. 
Update the credentials properties file to correct the user prefix.

### DIFF
--- a/docker/JBoss/src/configuration/standalone.xml
+++ b/docker/JBoss/src/configuration/standalone.xml
@@ -47,8 +47,8 @@
 		<property name="web.protocol" value="http"/>
 		<property name="web.port" value="8080"/>
 		<property name="web.contextpath" value="/iaf-test"/>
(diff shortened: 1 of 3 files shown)


docker/JBoss/src/configuration/standalone.xml
```diff
@@ -47,8 +47,8 @@
 		<property name="web.protocol" value="http"/>
 		<property name="web.port" value="8080"/>
 		<property name="web.contextpath" value="/iaf-test"/>
-		<property name="credentialFactory.filesystem.root" value="/opt/frank/secrets"/>
-		<property name="credentialFactory.class" value="org.frankframework.credentialprovider.WildFlyCredentialFactory"/>
```
nielsm5, line 51: We need to keep using and testing this credential factory!

PR #7906, opened 2024-11-11
Fix uploading artifact to existing artifact in matrix workflow

philipsens:


tnleeuw:
Is this only needed for 7.9?

philipsens:
> Is this only needed for 7.9?

No, every branch that has a matrix workflow.

Do you think this artifact name is good, or should I do something like `java8/FrankDoc`?

nielsm5 (approved):

nielsm5 (approved):

PR #7922, opened 2024-11-12
Get server timezone from FF Core & handle in console frontend

Matthbo:
Server now has timezone information in `/server/info`.
Console frontend has been updated to use the new ServerTime service which keeps track of the framework's timezone & epoch time.
TopNavBar time display & TimeSince & ToDate pipes have been updated to use the framework time.

This does not do anything for the statistics page yet as that needs to be reworked and seem broken (unrelated) now anyway


philipsens:
Currently, I don't see the server time in the header, instead I see my local time. It would also be nice to add the timezone on hover of the server time and to add the timezone in the information modal.

philipsens (approved):

PR #7983, opened 2024-11-22
Delete Spring profile that was intended to test Ladybug

mhdirkse:


mhdirkse:
Deze file kan weg omdat we hem in [wearefrank/ladybug-ff-test-webapp](https://github.com/wearefrank/ladybug-ff-test-webapp) hebben

nielsm5 (approved):

PR #7997, opened 2024-11-25
Update AbstractFileSystemListener config warning

nielsm5:


filesystem/src/main/java/org/frankframework/filesystem/AbstractFileSystemListener.java
```diff
@@ -160,7 +160,7 @@ public void configure() throws ConfigurationException {
 		}
 		targetProcessStates = ProcessState.getTargetProcessStates(knownProcessStates);
 		if (!(knownProcessStates.contains(ProcessState.INPROCESS) && isFileTimeSensitive()) && !(isOverwrite() || getNumberOfBackups() > 0)) {
-			ConfigurationWarnings.add(this, log, "It is recommended to configure either an in-process folder and to set 'fileTimeSensitive', or configure 'overwrite' or 'numberOfBackups' to avoid problems when files with the same name are processed.");
+			ConfigurationWarnings.add(this, log, "It is recommended to either configure an in-process folder with 'fileTimeSensitive=true', or configure 'overwrite' or 'numberOfBackups', to avoid problems when files with the same name are processed.");
```
evandongen, line 163:
```suggestion
			ConfigurationWarnings.add(this, log, "It is recommended to either configure an in-process folder with 'fileTimeSensitive=true', or configure 'overwrite' or 'numberOfBackups' to avoid problems when files with the same name are processed.");
```

evandongen (approved):

PR #8005, opened 2024-11-26
Rename jobs so they're not all named "build" or "analyze"

tnleeuw:


philipsens:
Wouldn't it be nice to have a name attribute for the build job as well?

philipsens:
> > Wouldn't it be nice to have a name attribute for the build job as well?
> 
> For which build job? 😉
> 
> I'm not sure it would really add much value though to be honest.

We are changing the job keys, so the display name also changes and causes less confusion, right? So in that scenario, wouldn't it be nice to add a name attribute so we can use capital letters and spaces? That the job keys don't matter as much anymore.

philipsens:
https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idname

.github/workflows/angular-karma-tests.yml
```diff
@@ -16,7 +16,7 @@ on:
       - 'release/[0-9]+.[0-9]+'
 
 jobs:
-  build:
+  run-karma-tests:
```
philipsens, line 19:
```suggestion
  run-karma-tests:
    name: Run Karma tests
```

.github/workflows/compare-frankdoc.yml
```diff
@@ -10,7 +10,7 @@ on:
         required: true
 
 jobs:
-  build:
+  compare-frank-doc:
```
philipsens, line 13:
```suggestion
  compare-frank-doc:
    name: Compare Frank!Doc
```

.github/workflows/maven-build.yml
```diff
@@ -40,7 +40,7 @@ on:
       - "docker/**"
 
 jobs:
-  build:
+  build-maven-artefacts:
```
philipsens, line 43:
```suggestion
  build-maven-artifacts:
    name: Build Maven artifact
```

tnleeuw:
Github is configured that a job named "build" is required, but there's no more such job. So the commit-checks never complete.
Where should this be configured?

philipsens (approved):
Thank youu!

PR #8007, opened 2024-11-26
Errors in events should not propagate back to the source

tnleeuw:
Make event handling asynchronous and add an error-handler.
Adding the error-handler will prevent the propagation of errors from events back to their publisher. Making event asynchronous seems like a logical thing to do, so the main process can continue while events are handled in the background, but isn't really part of the original issue.

### DIFF
--- a/core/src/main/java/org/frankframework/management/bus/endpoints/Monitoring.java
+++ b/core/src/main/java/org/frankframework/management/bus/endpoints/Monitoring.java
@@ -22,13 +22,8 @@ import java.util.List;
 import java.util.Map;
 
 import jakarta.annotation.security.RolesAllowed;
+
 import org.apache.commons.lang3.StringUtils;
-import org.frankframework.management.bus.TopicSelector;
-import org.frankframework.management.bus.dto.MonitorDTO;
-import org.frankframework.management.bus.dto.TriggerDTO;
-import org.frankframework.management.bus.message.EmptyMessage;
-import org.frankframework.management.bus.message.JsonMessage;
-import org.frankframework.management.bus.message.StringMessage;
 import org.springframework.context.ApplicationContext;
 import org.springframework.http.MediaType;
 import org.springframework.messaging.Message;
@@ -40,6 +35,12 @@ import org.frankframework.management.bus.BusAware;
 import org.frankframework.management.bus.BusException;
 import org.frankframework.management.bus.BusMessageUtils;
 import org.frankframework.management.bus.BusTopic;
+import org.frankframework.management.bus.TopicSelector;
+import org.frankframework.management.bus.dto.MonitorDTO;
+import org.frankframework.management.bus.dto.TriggerDTO;
+import org.frankframework.management.bus.message.EmptyMessage;
+import org.frankframework.management.bus.message.JsonMessage;
+import org.frankframework.management.bus.message.StringMessage;
 import org.frankframework.monitoring.AdapterFilter;
 import org.frankframework.monitoring.EventThrowing;
 import org.frankframework.monitoring.EventType;
@@ -51,11 +52,8 @@ import org.frankframework.monitoring.Severity;
 import org.frankframework.monitoring.SourceFiltering;
 import org.frankframework.monitoring.Trigger;
 import org.frankframework.monitoring.events.ConsoleMonitorEvent;
-
 import org.frankframework.util.EnumUtils;
-
 import org.frankframework.util.JacksonUtils;
-
 import org.frankframework.util.SpringUtils;
 
 @BusAware("frank-management-bus")
@@ -131,7 +129,9 @@ public class Monitoring extends BusEndpointBase {
 		try {
 			monitor.configure();
 		} catch (ConfigurationException e) {
-			throw new BusException("unable to (re)configure Monitor", e);
+			log.info("Unable to (re)configure monitor [{}]", monitor.getName(), e);
+			// Throw this as Warning / Bad Request, not able to configure was likely due to bad user input and not due to internal server error
+			throw new BusException("unable to (re)configure Monitor: " + e.getMessage());
 		}
 
 		return EmptyMessage.created();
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/frankframework/monitoring/MonitorManager.java
```diff
@@ -52,7 +53,7 @@ public class MonitorManager extends AbstractConfigurableLifecyle implements Appl
 
 	private @Getter @Setter ApplicationContext applicationContext;
 	private final List<Monitor> monitors = new ArrayList<>();                            // All monitors managed by this MonitorManager
-	private final Map<String, Event> events = new HashMap<>();                           // All events that can be thrown
+	private final Map<String, Event> events = new ConcurrentHashMap<>();                 // All events that can be thrown
```
tnleeuw, line 56: Since I made the handling of event asynchronous, it turns out this map can now be updated from multiple different threads during startup.
So it needs to be concurrency-safe.

core/src/main/resources/SpringApplicationContext.xml
```diff
@@ -17,6 +17,12 @@
 
 	<import resource="springUnmanagedDeployment.xml"/>
 
+	<bean name="applicationEventMulticaster" class="org.springframework.context.event.SimpleApplicationEventMulticaster" autowire="byName">
+		<property name="errorHandler">
```
tnleeuw, line 21: Spring will create this bean if it's not defined, but without an error-handler or task-executor.

Creating our own bean we can control these properties and errors from event handlers won't be propagated back to the event sources, but instead flow to the error-handler.

core/src/main/java/org/frankframework/management/bus/endpoints/Monitoring.java
```diff
@@ -131,7 +129,9 @@ public Message<String> addMonitorOrTrigger(Message<?> message) {
 		try {
 			monitor.configure();
 		} catch (ConfigurationException e) {
-			throw new BusException("unable to (re)configure Monitor", e);
+			log.info("Unable to (re)configure monitor [{}]", monitor.getName(), e);
```
tnleeuw, line 132: NB: Despite the exception we catch, the invalid trigger has already been added and we are stuck with that invalid trigger until we update it.

Ideally we should validate the TriggerDTO before creating and adding the trigger, also on update.

Not sure if we should do that as part of this PR, or a follow-up PR.

nielsm5: Is it not possible to remove it if we get an exception?

core/src/main/java/org/frankframework/management/bus/endpoints/Monitoring.java
```diff
@@ -131,7 +129,9 @@ public Message<String> addMonitorOrTrigger(Message<?> message) {
...
+			// Throw this as Warning / Bad Request, not able to configure was likely due to bad user input and not due to internal server error
```
evandongen, line 133: Is this something to be solved in the new PR you mentioned? Otherwise you might return something like `return new EmptyMessage(400);` here?

core/src/main/java/org/frankframework/monitoring/MonitorManager.java
```diff
@@ -126,17 +127,12 @@ public Monitor getMonitor(int index) {
 		return monitors.get(index);
 	}
 	public Monitor findMonitor(String name) {
```
evandongen, line 129: Kan dit niet gelijk `public Optional<Monitor> findMonitor(String name)` worden? Zo te zien wordt dit alleen in `org.frankframework.management.bus.endpoints.Monitoring#getMonitor` gebruikt, en die is er ook makkelijk op aan te passen.

management-gateway/src/main/java/org/frankframework/management/bus/BusException.java
```diff
@@ -28,12 +30,12 @@
...
+	private static final @Serial long serialVersionUID = 1L;
 
 	private final @Getter int statusCode;
 
 	/**
-	 * Seen as WARNING
+	 * Seen as WARNING / HTTP Bad Request
 	 */
 	public BusException(String message) {
 		this(message, 400);
```
evandongen, line 41: Ah, dit komt op dezelfde http status neer als mijn EmptyMessage(400) suggestie vermoed ik. Laat die opmerking dan maar zitten ;-)

evandongen (approved):

core/src/main/resources/SpringApplicationContext.xml
nielsm5: CO0l

core/src/test/java/org/frankframework/receivers/ReceiverTest.java
```diff
@@ -606,6 +620,69 @@ public void run() {
 		verify(jmsMessage, atLeastOnce()).acknowledge();
 	}
 
+	@Test
```
nielsm5, line 623:
```suggestion
	@Test
	// also ensures that we use our own EventErrorHandler.
```

Maar dan moet je ook de `testApplicationContext.xml` aanpassen :P

core/src/main/java/org/frankframework/management/bus/endpoints/Monitoring.java
tnleeuw: Zoals de code nu in elkaar zit is dat niet makkelijk.

management-gateway/src/main/java/org/frankframework/management/bus/BusException.java
tnleeuw: Klopt, alleen is de message niet empty.

core/src/test/java/org/frankframework/receivers/ReceiverTest.java
tnleeuw: Als we de `testApplicationContext.xml` aanpassen dan passen we dat aan voor alle tests, nu alleen maar de 1 test waar we de error-handler nodig hebben. En definieren van die bean in de context geeft hem misschien ook automatisch een task-executor die uitvoering async maakt, en dat maakt de tests weer lastiger.

Dit leek mij de meest eenvoudige oplossing voor de test.

core/src/main/java/org/frankframework/management/bus/endpoints/Monitoring.java
tnleeuw: I now made a change to remove the trigger but it would feel better to me if we would actually have a validation-step before adding it.
I'll do that in a separate PR so you can see if you're OK with the changes I have in mind for that.

nielsm5 (approved):

PR #8023, opened 2024-11-29
7969 fix two bugs with monitoring

### LINKED ISSUE #7969: Monitoring: checkbox "Destinations" does not work
### Current Behavior

I edit my monitor by checking the "Destinations" checkbox, see picture

image

Then I press save and I see the XML. It looks as follows:

```
<Monitor name="monitor example 1a" type="TECHNICAL" destinations="0">
<AlarmTrigger severity="CRITICAL" threshold="5" period="60">
<Event>Receiver Shutdown</Event>
</AlarmTrigger>
</Monitor>
```

It says "destinations = 0".

There is a destinations in my Configuration.xml and it has a name:

```
	<Monitoring>
		<SenderMonitorAdapterDestination>
			<IbisLocalSender name="triggerIt" javaListener="reactToMonitor"></IbisLocalSender>
		</SenderMonitorAdapterDestination>
	</Monitoring>
```

### Expected Behavior

Related to the "Destinations" checkbox, I should see something about destination "triggerIt". I expect to see that that destination is added with my monitor.

### Environment Information

**FF! 9.0.0-20241115.042343**: Frank2Example1 
Running on **LAPTOP-R1SK28C** using **Apache Tomcat/10.1.24**
Java Version: **OpenJDK Runtime Environment (21.0.2+13-LTS)**
Heap size: **168.6 MiB**, total JVM memory: **276.0 MiB**
Free memory: **107.4 MiB**, max memory: **1.0 GiB**
Free disk space: **115.1 GiB**, total disk space: **464.3 GiB**

### Steps To Reproduce

_No response_

### Configuration

```xml
<Configuration
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:noNamespaceSchemaLocation="../FrankConfig.xsd"
	>
	<Adapter name="Adapter1a">
		<Receiver name="Receiver1a">
			<ApiListener name="Listener1a" uriPattern="service1a"/>
		</Receiver>
		<Pipeline>
			<EchoPipe name="HelloWorld" getInputFromFixedValue="Hello World!"/>
		</Pipeline>
	</Adapter>
	<Monitoring>
		<SenderMonitorAdapterDestination>
			<IbisLocalSender name="triggerIt" javaListener="reactToMonitor"></IbisLocalSender>
		</SenderMonitorAdapterDestination>
	</Monitoring>
</Configuration>
```

### Input

_No response_

_No response_

_No response_

### Relevant Log Output

_No response_

### Anything else?

_No response_

thomaspj10:
Fixes #7969 

There are two bugs at play here:

First, there is a bug in the frontend which doesn't send the name of the destination, but instead sends the index, thus adding a `0`.

The name of a destination is not required and will automatically be instantiated with the name of the class (`SenderMonitorAdapter` in this case). The destination is stored in a map with the name being a key. Two destinations without a name are overwritten in the map and only one shows up. This will be fixed by following Niels' suggestion and by making the name attribute mandatory.

### DIFF
--- a/console/frontend/src/main/frontend/src/app/views/monitors/monitors.component.ts
+++ b/console/frontend/src/main/frontend/src/app/views/monitors/monitors.component.ts
@@ -79,11 +79,12 @@ export class MonitorsComponent implements OnInit, OnDestroy {
       for (const monitor of this.monitors) {
         monitor.displayName = monitor.name;
         if (monitor.raised) this.totalRaised++;
-        monitor.activeDestinations = [];
+        monitor.activeDestinations = {};
         for (const index in this.destinations) {
           const destination = this.destinations[index];
-          monitor.activeDestinations[index] = monitor.destinations.includes(destination);
+          monitor.activeDestinations[destination] = monitor.destinations.includes(destination);
         }
+        console.log(monitor.activeDestinations);
       }
     });
   }
--- a/console/frontend/src/main/frontend/src/app/views/monitors/monitors.component.ts
+++ b/console/frontend/src/main/frontend/src/app/views/monitors/monitors.component.ts
@@ -84,7 +84,6 @@ export class MonitorsComponent implements OnInit, OnDestroy {
           const destination = this.destinations[index];
           monitor.activeDestinations[destination] = monitor.destinations.includes(destination);
         }
-        console.log(monitor.activeDestinations);
       }
     });
   }
--- a/console/frontend/src/main/frontend/src/app/views/monitors/monitors.service.ts
+++ b/console/frontend/src/main/frontend/src/app/views/monitors/monitors.service.ts
@@ -12,7 +12,7 @@ export type Monitor = {
   lastHit: string;
   triggers: Trigger[];
   displayName: string;
-  activeDestinations: boolean[];
+  activeDestinations: { [key: string]: boolean };
   edit: boolean;
   alarm: Alarm;
 };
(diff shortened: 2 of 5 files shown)


console/frontend/src/main/frontend/src/app/views/monitors/monitors.service.ts
```diff
@@ -12,7 +12,7 @@ export type Monitor = {
   lastHit: string;
   triggers: Trigger[];
   displayName: string;
-  activeDestinations: boolean[];
+  activeDestinations: { [key: string]: boolean };
```
Matthbo, line 15: Normally we use a util type called Record which means exactly the same as this one but is a bit stricter during type checking
```suggestion
  activeDestinations: Record<string, boolean>;
```

console/frontend/src/main/frontend/src/app/views/monitors/monitors.component.ts
```diff
@@ -79,11 +79,12 @@ export class MonitorsComponent implements OnInit, OnDestroy {
...
         monitor.displayName = monitor.name;
         if (monitor.raised) this.totalRaised++;
-        monitor.activeDestinations = [];
+        monitor.activeDestinations = {};
         for (const index in this.destinations) {
           const destination = this.destinations[index];
-          monitor.activeDestinations[index] = monitor.destinations.includes(destination);
+          monitor.activeDestinations[destination] = monitor.destinations.includes(destination);
         }
+        console.log(monitor.activeDestinations);
```
Matthbo, line 87: Remove `console.logs()` meant for debugging purposes

thomaspj10: Woops

nielsm5 (approved):

PR #8032, opened 2024-11-30
Move placeholder text on edit-trigger page to a 'hint'

nielsm5:
It's a lot of text, and it overflows if there is not enough space. It's not the best solution but for now it's the easiest...
Placeholders do not work when there is a default value. Though this not the best solution, and perhaps we should revamp all the forms, but at least it improves the current user-flow a bit...


Matthbo (commented):
Do the hints need to be so extensive?
Wouldn't something like `(Amount of triggers before alarming)` & `(Amount of seconds)` be clear enough?

evandongen (approved):

PR #8033, opened 2024-12-02
Show the timezone in the header and modal

philipsens:
Add the name of the timezone in the system information and on hover in the header.

I've tried to show the GMT+X, but it didn't work out with the current methods that angular provides. I still consider this a bug and an issue should be created.


philipsens:
Timezone in server time is lost after formatting

This is a bug because it will show the wrong timezone, if shown.

Using Etc/UTC, I see this:
image

But it should be +0000

**To reproduce:**
Set `console.dateFormat: yyyy-MM-dd HH:mm:ss zzzz` in environment variables

Matthbo (approved):

PR #8040, opened 2024-12-03
Hide navigation elements and check if user has access to routes

philipsens:
**Description:**

This pull request introduces significant enhancements to the authentication and navigation components of the application, focusing on access control and permission management.

**Key Changes:**

1. **Auth Guard Implementation:**
   - Added an authentication guard that utilizes the `appHasAccessTo` directive to control access specific navigation links. This ensures that only authorized users can view certain features, thereby improving security and user experience.

2. **Refactoring of AuthService:**
   - The `AuthService` has been refactored to leverage Angular's dependency injection, enhancing maintainability and readability. 
   - Introduced a new method for asynchronous permission loading, which reduces nested subscriptions and improves the overall structure of the code.

3. **Streamlined Permissions Loading:**
   - Implemented a permissions observable in the `AuthService` to manage permission loading more efficiently. 
   - Removed the boolean flag for permissions loaded and replaced it with a shared observable that caches results, simplifying the logic and ensuring permissions are fetched only once. This change enhances performance and maintainability.

These updates collectively enhance the application's security framework and improve the user experience by ensuring that navigation is appropriately restricted based on user permissions. 

Please review the changes and provide feedback. Thank you!

### DIFF
--- a/console/frontend/src/main/frontend/src/app/services/auth.service.ts
+++ b/console/frontend/src/main/frontend/src/app/services/auth.service.ts
@@ -1,8 +1,15 @@
 import { inject, Injectable } from '@angular/core';
 import { AppService } from '../app.service';
 import { HttpClient } from '@angular/common/http';
-import { Observable } from 'rxjs';
-import { Link, LinkName, SecurityItemsService } from '../views/security-items/security-items.service';
+import { combineLatest, map, Observable, of, tap } from 'rxjs';
+import {
+  Link,
+  LinkName,
+  SecurityItemsService,
+  SecurityItems,
+  SecurityRole,
+  Links,
+} from '../views/security-items/security-items.service';
 
 @Injectable({
   providedIn: 'root',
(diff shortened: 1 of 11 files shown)


console/frontend/src/main/frontend/src/app/services/auth.service.ts
```diff
@@ -1,55 +1,85 @@
...
-          sessionStorage.getItem('location') || 'status',
-        );
-    } else {
-      if (this.consoleState.init > 0) {
-        if (!this.router.url.includes('login'))
-          sessionStorage.setItem('location', this.router.url || 'status');
-        this.router.navigateByUrl('login');
-      }
-    }
-  } */
```
Matthbo, line 41: We could rewrite this for the 9.1 release but isnt it better to keep it commented here for now?

console/frontend/src/main/frontend/src/app/services/auth.service.ts
```diff
@@ -1,55 +1,85 @@
...
+    return this.permissionsObservable$;
+  }
+
+  private initializePermissionsObservable(): Observable<void> {
+    return this.fetchPermissionData().pipe(
+      tap(([securityItems, links]) => this.updatePermissions(securityItems, links)),
+      map(() => {}),
+      shareReplay(1),
+    );
+  }
```
Matthbo, line 51: It really feels like initializePermissionsObservable should be a promise that calls the fetchPermissionData() and subscribes to it to then call `updatePermissions` with the result instead of having a very weirdly chained together observable which doesnt give back any real value.
I think a promise is a better fit here

console/frontend/src/main/frontend/src/app/services/auth.service.ts
```diff
@@ -1,55 +1,85 @@
...
-      if (this.consoleState.init > 0) {
-        if (!this.router.url.includes('login'))
-          sessionStorage.setItem('location', this.router.url || 'status');
-        this.router.navigateByUrl('login');
-      }
-    }
-  } */
+  constructor() {
+    this.permissionsObservable$ = this.initializePermissionsObservable();
+  }
```
Matthbo, line 29: This doesn't have to be in the constructor, could be handled like the above properties

console/frontend/src/main/frontend/src/app/views/security-items/security-items.service.ts
```diff
@@ -81,16 +163,28 @@ export type Links = { links: Link[] };
...
+  private securityItemsCache: SecurityItems | null = null;
+  private endpointsWithRolesCache: Links | null = null;
+  private readonly http: HttpClient = inject(HttpClient);
+  private readonly appService: AppService = inject(AppService);
 
   getSecurityItems(): Observable<SecurityItems> {
-    return this.http.get<SecurityItems>(`${this.appService.absoluteApiPath}securityitems`);
+    if (this.securityItemsCache) {
+      return of(this.securityItemsCache);
+    }
```
Matthbo, line 174: Maybe this should have a way to be cleared or updated too, for example when a user logs out

console/frontend/src/main/frontend/src/app/views/security-items/security-items.service.ts
```diff
@@ -81,16 +163,28 @@ export type Links = { links: Link[] };
...
+    return this.http
+      .get<SecurityItems>(`${this.appService.absoluteApiPath}securityitems`)
+      .pipe(tap((data) => (this.securityItemsCache = data)));
   }
 
   getEndpointsWithRoles(): Observable<Links> {
-    return this.http.get<Links>(`${this.appService.absoluteApiPath}?allowedRoles=true`);
+    if (this.endpointsWithRolesCache) {
+      return of(this.endpointsWithRolesCache);
+    }
```
Matthbo, line 184: Maybe this should have a way to be cleared or updated too, for example when a user logs out

Matthbo (changes requested):

console/frontend/src/main/frontend/src/app/services/auth.service.ts
philipsens: hmmmmmmmmmmmmmmmmm I'm not a fan of comments. We could make an issue and reference a commit or something. 
I can add it back for now.

console/frontend/src/main/frontend/src/app/services/auth.service.ts
philipsens: I think I had it as a promise at first, but that I changed it to only update everything once and to notify everything that's waiting on it. Could you give an example how you'd accieve this with a promise?

console/frontend/src/main/frontend/src/app/views/security-items/security-items.service.ts
philipsens: Good point. I also thought to tie this into the auth service.

philipsens: Added a clear to the logout. This will also trigger if redirected to logout page.

console/frontend/src/main/frontend/src/app/views/security-items/security-items.service.ts
philipsens: Same.

console/frontend/src/main/frontend/src/app/services/auth.service.ts
philipsens: Thanks for helping me! It is way nicer now.

Matthbo (approved):

PR #8064, opened 2024-12-05
Change messageId encoding to base64

philipsens:
This pull request changes the encoding of `messageId` from double URL encoding to base64. This update is crucial for handling `messageId` values that may include '/' characters, which previously caused decoding issues. The changes ensure improved reliability and compatibility in message processing across both backend and frontend components, enhancing the overall functionality of the application.

### DIFF
--- a/commons/src/main/java/org/frankframework/util/HttpUtils.java
+++ b/commons/src/main/java/org/frankframework/util/HttpUtils.java
@@ -15,8 +15,8 @@
  */
 package org.frankframework.util;
 
-import java.io.UnsupportedEncodingException;
-import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
 import java.util.Enumeration;
 
 import jakarta.servlet.http.HttpServletRequest;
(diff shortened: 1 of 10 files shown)


commons/src/main/java/org/frankframework/util/HttpUtils.java
```diff
@@ -76,4 +78,14 @@ public static String urlDecode(String input) {
 			throw new IllegalArgumentException(e);
 		}
 	}
+
+	public static String decodeBase64(String input) {
```
evandongen, line 82: Op andere plekken in de code doen we bv `byte[] versiondata = Base64.decodeBase64(versiondataB64);` . Heb je daar echt een nieuwe methode voor nodig?

philipsens: Ah, java util vs Apache commons 🤷🏻

philipsens: Oh, and DRY ofc. The try/catch is probs a good idea for stuff coming from the frontend.

nielsm5: ( Perhaps this can be moved to StringUtil or Misc? It doesn't really have much to do with http, and I want to remove the HttpUtils at some point.  )

nielsm5 (approved):

Matthbo (approved):

PR #8102, opened 2024-12-16
Fix NPE when SecuritySettings.webSecurityEnabled was called before the class was initialized.

nielsm5:


security/src/test/java/org/frankframework/lifecycle/servlets/ServletConfigurationTest.java
```diff
@@ -8,13 +8,26 @@
...
+
+@SpringJUnitConfig(initializers = {SpringRootInitializer.class})
 public class ServletConfigurationTest {
 
+	public static class SpringRootInitializer implements ApplicationContextInitializer<ConfigurableApplicationContext> {
+		@Override
+		public void initialize(ConfigurableApplicationContext applicationContext) {
+			SecuritySettings.setupDefaultSecuritySettings(applicationContext.getEnvironment());
+		}
+	}
```
tnleeuw, line 29: Kon deze NPE alleen gebeuren in tests, en niet in productie Spring contexts?

PR #8134, opened 2024-12-19
Fix showing non-read stream in ladybug

evandongen:
When an adapter was called via a SendMessageJob, the result was never read, resulting in "Captured stream was closed without being read". By applying this change, the output will be 'read' in that case.

### DIFF
--- a/core/src/test/java/org/frankframework/util/StreamCaptureUtilsTest.java
+++ b/core/src/test/java/org/frankframework/util/StreamCaptureUtilsTest.java
@@ -3,10 +3,14 @@ package org.frankframework.util;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import java.io.BufferedOutputStream;
+import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
 import java.net.URL;
 import java.util.Arrays;
 
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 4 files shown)


core/src/test/java/org/frankframework/util/StreamCaptureUtilsTest.java
```diff
@@ -3,27 +3,62 @@
...
+			return value;
+		}
+	}
+
+	@Test
+	void assertCaptureUtilsCallsWriteMethod() throws IOException {
+		ByteArrayInputStream bis = new ByteArrayInputStream("Test input".getBytes());
+
+		// We need something effectively final here, we can't use a boolean directly unfortunately.
+		BooleanContainer container = new BooleanContainer();
```
tnleeuw, line 36: AtomicBoolean?

core/src/test/java/org/frankframework/util/StreamCaptureUtilsTest.java
```diff
@@ -3,27 +3,62 @@
...
+			public void write(byte[] b, int off, int len) {
+				container.setValue(true);
+				super.write(b, off, len);
+			}
+		};
+
+		InputStream inputStream = StreamCaptureUtils.captureInputStream(bis, baos, 16);
+		inputStream.close();
+
+		Assertions.assertTrue(container.getValue());
```
tnleeuw, line 49: static import van assertTrue? En liefst een message erbij die uitlegt wat je assert

tnleeuw (changes requested):

nielsm5 (approved):

PR #8142, opened 2024-12-20
Implement a custom name for the root element

### LINKED ISSUE #8010: Option to rename root element in json2xml conversion

While converting a json to an xml using the JsonPipe, it is possible to use the "addXmlRootElement" attribute to add a root element. It is currently limited to being named "root". It would be convenient to be able to rename the root element to something else.

### Describe the Feature

I would like to see an extra attribute added to the JsonPipe with which it is possible to rename the root element something other than "root".

No, community resolution only

### Anything else?

_No response_

thomaspj10:
Closes #8010

### DIFF
--- a/core/src/main/java/org/frankframework/pipes/JsonPipe.java
+++ b/core/src/main/java/org/frankframework/pipes/JsonPipe.java
@@ -114,7 +114,7 @@ public class JsonPipe extends FixedForwardPipe {
 					return convertJsonToXml(jr.read());
 				} catch (JsonException e) {
 					log.debug("cannot parse input as JsonStructure", e);
-					Message result = new Message("<" + rootElementName + ">"+message.asString()+"</" + rootElementName + ">");
+					Message result = new Message(String.format("<%s>%s</%s>", rootElementName, message.asString(), rootElementName));
 					result.getContext().withMimeType(MediaType.APPLICATION_XML);
 					return new PipeRunResult(getSuccessForward(), result);
 				}
--- a/core/src/main/java/org/frankframework/pipes/JsonPipe.java
+++ b/core/src/main/java/org/frankframework/pipes/JsonPipe.java
@@ -29,7 +29,7 @@ import jakarta.json.JsonValue;
 
 import lombok.Getter;
 
-import org.codehaus.plexus.util.StringUtils;
+import org.apache.commons.lang3.StringUtils;
 
 import org.frankframework.doc.Default;
 import org.frankframework.doc.EnterpriseIntegrationPattern;
--- a/core/src/main/java/org/frankframework/pipes/JsonPipe.java
+++ b/core/src/main/java/org/frankframework/pipes/JsonPipe.java
@@ -28,6 +28,8 @@ import jakarta.json.JsonReader;
 import jakarta.json.JsonValue;
 
 import lombok.Getter;
+
+import org.frankframework.doc.Default;
 import org.frankframework.doc.EnterpriseIntegrationPattern;
 
 import org.springframework.http.MediaType;
@@ -56,8 +58,12 @@ import org.frankframework.util.TransformerPool;
  */
 @EnterpriseIntegrationPattern(EnterpriseIntegrationPattern.Type.TRANSLATOR)
 public class JsonPipe extends FixedForwardPipe {
+
+	private static final String DEFAULT_ROOT_ELEMENT_NAME = "root";
+
 	private @Getter Direction direction = Direction.JSON2XML;
 	private Boolean addXmlRootElement = null;
+	private @Getter String rootElementName = DEFAULT_ROOT_ELEMENT_NAME;
 	private @Getter boolean prettyPrint = false;
 
 	private TransformerPool tpXml2Json;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/frankframework/pipes/JsonPipe.java
```diff
@@ -71,6 +77,11 @@ public enum Direction {
 	public void configure() throws ConfigurationException {
 		super.configure();
 
+		// rootElementName has been modified from its default value
+		if (getDirection() == Direction.XML2JSON && !DEFAULT_ROOT_ELEMENT_NAME.equals(rootElementName)) {
+			throw new ConfigurationException("rootElementName can not be used when direction is XML2JSON");
+		}
+
```
tnleeuw, line 84: I see that you addressed the Sonar complaint but now, in theory, somebody could manually configure `"root"` as the root element name.

Perhaps you can leave the field `null` by default and then you can check here for field being not null when direction is `XML2JSON`.
And if the direction is `JSON2XML` and it is `null` then set the default?
That should catch all cases.

core/src/main/java/org/frankframework/pipes/JsonPipe.java
```diff
@@ -161,14 +172,23 @@ public void setVersion(String version) {
...
 	public void setAddXmlRootElement(boolean addXmlRootElement) {
 		this.addXmlRootElement = addXmlRootElement;
 	}
 
-	/** Format the output in easy legible way (currently only for XML) */
+	/**
+	 * When direction is JSON2XML, specifies the name of the root element when {@literal addXmlRootElement} is {@code true}.
+	 * When direction is XML2JSON, can not be used.
+	 */
+	@Default(DEFAULT_ROOT_ELEMENT_NAME)
```
tnleeuw, line 186: I don't know which is preferred officially: '@Default` annotation, or `@ff.default` in the JavaDoc.
Personally I'd prefer the use of `@ff.default` in the JavaDoc, which is also more commonly used.

core/src/main/java/org/frankframework/pipes/JsonPipe.java
```diff
@@ -161,14 +172,23 @@ public void setVersion(String version) {
...
+	 * When direction is XML2JSON, it includes the name of the root element as a key in the converted message, thus preserving the structure of the original input message.
 	 * @ff.default TRUE when JSON2XML and FALSE when XML2JSON
 	 */
 	public void setAddXmlRootElement(boolean addXmlRootElement) {
 		this.addXmlRootElement = addXmlRootElement;
 	}
 
-	/** Format the output in easy legible way (currently only for XML) */
+	/**
+	 * When direction is JSON2XML, specifies the name of the root element when {@literal addXmlRootElement} is {@code true}.
```
tnleeuw, line 183: So the custom root name is ignored when `addXmlRootElement` is `false`?
Perhaps that should also be verified then in `configure()`? Not sure if that should also throw an error but I guess that it should?

core/src/main/java/org/frankframework/pipes/JsonPipe.java
thomaspj10: The annotation is actually way more useful IMO. It allows me to reference the value of the constant, which is not possible with `@ff.default`. This prevents the documentation from becoming out of sync with the actual values used within the code.

core/src/main/java/org/frankframework/pipes/JsonPipe.java
thomaspj10: Agreed! Will do

tnleeuw:
@nielsm5 In dit PR zie ik fouten in de Exchange Mail tests...?

evandongen:
> @nielsm5 In dit PR zie ik fouten in de Exchange Mail tests...?

Ik zie dit ook bij https://github.com/frankframework/frankframework/pull/8139

core/src/main/java/org/frankframework/pipes/JsonPipe.java
```diff
@@ -97,7 +114,7 @@ public PipeRunResult doPipe(Message message, PipeLineSession session) throws Pip
 					return convertJsonToXml(jr.read());
 				} catch (JsonException e) {
 					log.debug("cannot parse input as JsonStructure", e);
-					Message result = new Message("<root>"+message.asString()+"</root>");
+					Message result = new Message("<" + rootElementName + ">"+message.asString()+"</" + rootElementName + ">");
```
evandongen, line 117: Kan hier iets als

`Message result = new Message(String.format("<%s>%s</%s>", rootElementName, message.asString(), rootElementName);`

?

evandongen (approved):

thomaspj10:
De master is inderdaad rood door die Exchange tests die falen.

tnleeuw (approved):

tnleeuw:
> > @nielsm5 In dit PR zie ik fouten in de Exchange Mail tests...?
> 
> Ik zie dit ook bij #8139

Ik vraag me af of dit veroorzaakt wordt door meerdere CI builds van verschillende PRs die tegelijkertijd draaien en dezelfde Exchange Mailbox gebruiken, en dezelfde base-folder.

nielsm5 (approved):

PR #8146, opened 2024-12-23
Add create base folder to exchange tests

nielsm5:


filesystem/src/test/java/org/frankframework/filesystem/exchange/ExchangeFileSystemTestHelper.java
```diff
@@ -34,7 +34,7 @@
 
 @Log4j2
 public class ExchangeFileSystemTestHelper implements IFileSystemTestHelper {
-	public static final int WAIT_MILLIS = 50;
+	public static final int WAIT_MILLIS = 150;
```
nielsm5, line 37:
```suggestion
	public static final int WAIT_MILLIS = 250;
```

evandongen (approved):

Issue #5934, opened 2023-12-07
Console components rewrite


Matthbo:
### Goal
Rewrite component code based on the old console using latest ES coding styles and features.

### Description
Most of the code in the angular components are 'based on' or refactored code from the old angular.js console and have only been touched up to work inside the Angular framework, all the components, directives, services and other code called during the lifetime of the components should be redesigned when needed and rewritten to the latest ECMAScript standards and best practices in JS/TS and the Angular framework.

### Tasks 
- [x] #6334
- [x] #6411
- [x] #5982
- [x] #6335
- [x] #6336
- [x] #6627
- [x] #6658
- [x] #7929
- [x] Super sexy (multi) drop-down component (like #5893 but not browser dependent)

Per view:
- [x] #6337
- [x] #8249

philipsens:
The new dropdown can not yet handle multi select. But I was thinking it might be weird to implement that anyway. Because the input field should contain separated values or something. It might be better to use a normal dropdown with add button (or on 'enter') that adds a "chip".

PR #8149, opened 2024-12-24
Add dynamic drop down to select an adapter in "Test a Pipeline" faster 

### LINKED ISSUE #5893: Add dynamic drop-down to faster select an adapter in Test a Pipeline
**Current situation**

1. The screen 'Test a Pipeline' has a mandatory step to select a Configuration and then select an adapter from the whole list. That also involves a lot of scrolling. You need to know which Configuration is used in advance. 

**To Be situation**
1. The 'Test a Pipeline' screen has a dynamic drop-down, at 'select an adapter' which supports text filtering, where you can type in e.g. 'xslt' and all matching scenario's with '*xslt*' (case insensitive) are shown in the drop down. Easy to select directly the one you like. 
2. The 'Test a Pipeline' screen 'Configuration' pull-down' is not mandatory anymore. When nothing is selected, all adapters are shown/available to filter the result list at step 2. Also add an empty entry in that pull-down, to select nothing (empty the configuration filter). 

**Reporter**
Jacobjob

philipsens:
This PR adds a "combobox" component. It acts as a dropdown with autocomplete. 
It can be used with just the keyboard by typing using the keys and pressing enter (or by filling in the whole value by hand and tabbing away) 
It can also be used with the mouse only, by clicking on the input and scrolling + clicking on the preferred option. 
A combination of this techniques is also possible.
Closes #5893

### DIFF
--- a/console/frontend/src/main/frontend/src/app/app.module.ts
+++ b/console/frontend/src/main/frontend/src/app/app.module.ts
@@ -92,6 +92,7 @@ import { MonitorsNewComponent } from './views/monitors/monitors-new/monitors-new
 import { ConfigurationTabListComponent } from './components/tab-list/configuration-tab-list.component';
 import { TabListComponent } from './components/tab-list/tab-list.component';
 import { HasAccessToLinkDirective } from './components/has-access-to-link.directive';
+import { ComboboxComponent } from './components/combobox/combobox.component';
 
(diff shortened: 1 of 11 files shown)


Matthbo (approved):

Matthbo:
Related to #5934

PR #8150, opened 2024-12-24
Fix quick submit in forms

### LINKED ISSUE #6865: In the Frank!Console I can no longer press ctrl+enter to submit a form
**Describe the issue**
In the Frank!Console I can no longer press ctrl+enter to submit a form.
I've only tested this in the 8.1 and 8.2

**Reporter**
Niels

philipsens:
Closes #6865


Matthbo (approved):
This makes a lot of sense! Never stood still to think about having the submit be done like this :D

Matthbo:
Ah but maybe fixing the karma test should be done before merging

philipsens:
> Ah but maybe fixing the karma test should be done before merging

Yes, I kind of expected it to fail, but I couldn't do it at the time. I'll fix it today.

nielsm5 (approved):
FANTASTISCH! 💯  😄  🎆

Matthbo (approved):

PR #8163, opened 2025-01-02
JSON2XML "deepSearch" finds wrong field

### LINKED ISSUE #7937: Json2XmlInputValidator bug FF 8.0.2 and 8.0.3
### Current Behavior

Json2XmlInputValidator fails on xsd validation

### Expected Behavior

Json2XmlInputValidator xsd succeeds

### Environment Information

All

### Steps To Reproduce

XSD with structure
/root/call/type (is of type enum call_type_type)
/root/type (is of type enum type_type)

An xml input message:
/root/call/type=outbound
/root/type=beurs
Gets rejected because inputvalue beurs is not a value in in enum call_type_type

<message>cvc-enumeration-valid: Value 'beurs' is not facet-valid with respect to enumeration '[]'. It must be a value from the enumeration.</message>

### Configuration

```xml
Tried in FF 8.0.2 and 8.0.3 and it fails in both.
```

### Input

_No response_

_No response_

_No response_

### Relevant Log Output

_No response_

### Anything else?

_No response_

tnleeuw:
Change the algorithm for `deepSearch` from depth-first to breadth-first. This makes the mapping of elements more similar to the mapping when `deepSearch=false`, so results are more reliable and predictable.

Closes #7937.

Also includes changes from PR #8162


core/src/main/java/org/frankframework/align/Json2Xml.java
```diff
@@ -48,6 +48,7 @@
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.apache.commons.lang3.builder.ToStringStyle;
+import org.apache.commons.lang3.tuple.Pair;
```
thomaspj10, line 51: I personally like to create a dataclass instead of a pair for readability

thomaspj10 (approved):
I don't fully understand everything, but there doesn't seem to be anything weird

core/src/main/java/org/frankframework/align/Json2Xml.java
tnleeuw: That is actually a good idea. Done.

evandongen (approved):

tnleeuw:
Ik wou een test-case toevoegen waarvan ik had verwacht dat die zou werken, maar hij werkte niet. Totdat ik begrijp waarom niet en dat eventueel fix, maak ik dit PR weer even draft.

nielsm5 (approved):

evandongen (approved):

PR #8168, opened 2025-01-03
Dynamically modify the timeout for the JMS and HTTP sender using parameters

### LINKED ISSUE #5239: Ability to set dynamic values for timeout in HttpSender

**Describe the issue**
Currently, I am building a dispatcher that will read configuration files and forward the incoming requests to the appropriate service.
While sending the requests to the other systems, I need to read the timeout  setting from the config file then configure it in HttpSender. Therefore it would be nice to be able to configure the timeout information in HttpSender during runtime.

Moreover, the timeout setting should be dynamic for JmsSender as well.

**Reporter**
Ali

thomaspj10:
Closes #5239


nielsm5:
Closing this for now, until there is an imminent issue, in which case we can reopen this.

PR #8191, opened 2025-01-08
Change Angular template flow directives to built-in flow control

Matthbo:


console/frontend/src/main/frontend/src/app/views/monitors/monitors.component.html
```diff
@@ -46,216 +47,270 @@ <h4>
...
-          <div *ngIf="monitor.edit">
-            <form role="form" (submit)="edit(monitor)" appQuickSubmitForm>
-              <div class="form-group">
-                <label class="col-sm-3 control-label">Name</label>
-                <div class="col-sm-9">
-                  <input class="form-control m-b" name="name" type="text" [(ngModel)]="monitor.displayName" />
-                </div>
+  @if (monitors.length > 0) {
+    <div class="row">
+      @for (monitor of monitors; track monitor.name) {
```
philipsens, line 56: Couldn't two monitors have the same name in the 'All' tab?

console/frontend/src/main/frontend/src/app/views/status/adapter-status/adapter-status.component.html
```diff
@@ -18,422 +18,479 @@ <h5>
...
+              </div>
+            </div>
+          </div>
+        }
+        <div class="row">
+          <div class="col-lg-9 col-md-12 messages table-responsive">
+            <h4>Messages</h4>
+            <table class="table table-striped" style="min-width: 620px">
+              <tbody>
+                @for (message of adapter.messages; track message.message) {
```
philipsens, line 454: Content of a message can be duplicate. I don't really know how it works in this case. But I think it might be better to track the hole message or `$index`, that's a reasonable default anyway.

console/frontend/src/main/frontend/src/app/views/status/configuration-messages/configuration-messages.component.html
```diff
@@ -1,52 +1,54 @@
...
-                    ></i>
-                  </td>
-                  <td class="text-center" style="width: 200px">
-                    <span [time]="message.date" appToDate></span>
-                  </td>
-                  <td>{{ message.message }}</td>
-                </tr>
-              </ng-template>
+              @if (messageLog[selectedConfiguration] !== null) {
+                @for (message of getMessageLog(selectedConfiguration); track message.message) {
```
philipsens, line 28: Might be the same here

philipsens (changes requested):

console/frontend/src/main/frontend/src/app/components/custom-views/custom-views.component.html
```diff
@@ -1,7 +1,9 @@
 <ul class="nav">
-  <li *ngFor="let view of customViews" uiSrefActive="active">
-    <a [routerLink]="['/customView', view.name]" [state]="{ view }"
-      ><i class="fa fa-desktop"></i> <span class="nav-label">{{ view.name }}</span></a
-    >
-  </li>
+  @for (view of customViews; track view.url) {
```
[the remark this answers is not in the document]
Matthbo: > I think we should track $index instead, because people can give their own URL's and there is a change that it would be the same.

I guess that is fair, but I think tracking in this particular case is not really an issue since you need to reload the framework anyway when you change the customViews properties.

> In the past, I have seen people make 2 custom views, one to embed a certain page and one to open it in a new tab.

This is why we now have a pop out button...

console/frontend/src/main/frontend/src/app/views/jms/jms-browse-queue/jms-browse-queue.component.html
```diff
@@ -72,34 +78,44 @@ <h4>Browse JMS Queue</h4>
...
+                    <th>Date/Time</th>
+                    <th>MessageId</th>
+                    <th>CorrelationID</th>
+                    @if (messages[0].text) {
+                      <th>Payload</th>
+                    }
+                  </tr>
+                </thead>
+                <tbody>
+                  @for (message of messages; track message.id; let i = $index) {
```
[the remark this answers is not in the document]
Matthbo: Doesn't every message still have a unique id's regardless? If not maybe we should let the user set a unique column then to speed up diffing in this for-loop (the reason tracking is now mandatory in the first place)

console/frontend/src/main/frontend/src/app/views/monitors/monitors.component.html
Matthbo: I'm not sure if you can, but you shouldn't right? :)

console/frontend/src/main/frontend/src/app/views/status/adapter-status/adapter-status.component.html
Matthbo: Pretty sure you can't track by $index since that is a variable only exists during for loop iteration and not while during the diffing process, but I could be wrong

console/frontend/src/main/frontend/src/app/components/custom-views/custom-views.component.html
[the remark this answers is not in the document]
philipsens: > This is why we now have a pop out button...

I know, I wanted to say that too in my comment, but let's be honest, if people want it, they'll do it. 

> I guess that is fair, but I think tracking in this particular case is not really an issue since you need to reload the framework anyway when you change the customViews properties.

Oh yeah, that's true. But maybe there is an edge case somewhere tho, like switching applications with Hazelcast that have different (or the same) links all together. Or maybe we could implement hot reloading properties in the future...

console/frontend/src/main/frontend/src/app/components/th-sortable.directive.spec.ts
```diff
@@ -14,14 +13,16 @@ import { NgForOf } from '@angular/common';
         </tr>
       </thead>
       <tbody>
-        <tr *ngFor="let item of items">
-          <td>{{ item.name }}</td>
-          <td>{{ item.value }}</td>
-        </tr>
+        @for (item of items; track item.name) {
```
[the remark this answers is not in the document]
Matthbo: It is meant to be unique, but I'm not enforcing that. Either I should to improve performance or accept there is no unique property and have a bit worse performance

console/frontend/src/main/frontend/src/app/views/monitors/monitors.component.html
philipsens: Ofc you shouldn't 👀 

You never know, I've seen an integration with multiple configurations that all have the same boilerplate. I wouldn't know if it also included monitors tho.

console/frontend/src/main/frontend/src/app/views/status/adapter-status/adapter-status.component.html
Matthbo: How interesting the migration set all track properties to the item itself rather than $index, I wonder if that has a reason or if $index is still fine to use

philipsens: Yeah good question, maybe because they don't want to make assumptions, and this way they cover advanced situations as well. I would think using `$index` is more performant when you want to update everything always.

console/frontend/src/main/frontend/src/app/components/th-sortable.directive.spec.ts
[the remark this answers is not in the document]
philipsens: I don't know where this is used, but I can imagine rows of a table having the same name. For example:

Activities
| Thingy      | Action      |
| ------------- | ------------- |
| Car | Created |
| Car | Updated |

console/frontend/src/main/frontend/src/app/components/custom-views/custom-views.component.html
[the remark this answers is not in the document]
Matthbo: I could solve this by using view.view since that is the index/key of the view in the customViews property

console/frontend/src/main/frontend/src/app/components/th-sortable.directive.spec.ts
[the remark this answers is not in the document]
Matthbo: I just realised, this for loop is inside a karma test so it doesnt matter much but yes its unique. See line 28.

console/frontend/src/main/frontend/src/app/views/status/adapter-status/adapter-status.component.html
Matthbo: Tracking message itself here since its a small and flat object to compare

console/frontend/src/main/frontend/src/app/components/th-sortable.directive.spec.ts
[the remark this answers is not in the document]
philipsens: Thank you for making it clear :)

philipsens (approved):

PR #8262, opened 2025-01-19
Update FrankConfig XSD

nielsm5:
We are now getting the follow XSD validation error. I think that in https://github.com/frankframework/frank-doc/pull/242 elements (used in batch processing) and ErrorMessageFormatters broke...

`
Validation error in [classpath:abcdef.xml] at line [3] when validating against schema [/xml/xsd/FrankConfig-compatibility.xsd]: cvc-complex-type.3.1: Value 'org.frankframework.errormessageformatters.SoapErrorMessageFormatter' of attribute 'className' of element 'ErrorMessageFormatter' is not valid with respect to the corresponding attribute use. Attribute 'className' has a fixed value of 'org.frankframework.errormessageformatters.ErrorMessageFormatter'.
`

### DIFF
--- a/core/src/main/resources/xml/xsd/FrankConfig-compatibility.xsd
+++ b/core/src/main/resources/xml/xsd/FrankConfig-compatibility.xsd
@@ -8502,12 +8502,15 @@
     <xs:choice>
       <xs:element name="ErrorMessageFormatter">
         <xs:complexType>
-          <xs:complexContent>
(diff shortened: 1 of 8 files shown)


evandongen (approved):

nielsm5:
> Might have something to do with #8240 as well

Na, it's definitely related to the doclet changes 😭

PR #8268, opened 2025-01-20
Fix CacheKey determined from parent-element name

nielsm5:


core/src/main/java/org/frankframework/cache/AbstractCacheAdapter.java
```diff
@@ -62,9 +62,9 @@ public abstract class AbstractCacheAdapter<V> implements ICache<String,V>, Frank
 	private TransformerPool valueTp=null;
 
 	@Override
-	public void configure(String ownerName) throws ConfigurationException {
+	public void configure() throws ConfigurationException {
 		if (StringUtils.isEmpty(getName())) {
-			setName(ownerName+"_cache");
+			setName(applicationContext.getId()+"_cache");
```
nielsm5, line 67: This is the adapter name, which was previously set in the pipeline (and passed through in configure).

tnleeuw (approved):

evandongen (approved):

tnleeuw (changes requested):
There are about 15 failing Larva tests -- these have to be fixed before I can approve the PR again 😬

tnleeuw (approved):

PR #8279, opened 2025-01-22
Copy fix from 7.9, with some minor changes

### LINKED ISSUE #8258: PipelineSession closeables keep growing when session keys are overwritten
### Current Behavior

In a special for-each case where PipelineSession keys are overwritten, the sub-adapter keys are removed (good!). However some of those keys are also stored in the closeables list, which is not 'closed' until the parent closes it's session.

This may cause the list of closeables to grow exponentially.

### Expected Behavior

String Messages do not need to be closed? For now this seems to be a workaround...

### Environment Information

7.9

### Steps To Reproduce

_No response_

### Configuration

```xml

```

### Input

_No response_

_No response_

_No response_

### Relevant Log Output

```console

```

### Anything else?

_No response_

tnleeuw:
Fix for ballooning PipeLineSession only.

Not yet cleaning up the `closeables`, that will be a separate PR.

After some consideration, I decided to put in just the same fix as is in 7.9 with only minor cosmetic changes, and a few small changes to the way it is tested.

Closes #8258

### DIFF
--- a/core/src/main/java/org/frankframework/core/PipeLineSession.java
+++ b/core/src/main/java/org/frankframework/core/PipeLineSession.java
@@ -143,9 +143,8 @@ public class PipeLineSession extends HashMap<String,Object> implements AutoClose
 		} else if (keysToCopy == null || "*".equals(keysToCopy)) { // if keys are not set explicitly ...
 			to.putAll(this);                                      // ... all keys will be copied
 		}
-		Set<AutoCloseable> closeablesInDestination = to.entrySet().stream()
-				.filter(entry -> shouldCloseSessionResource(entry.getKey(), entry.getValue()))
-				.map(Entry::getValue)
+		Set<AutoCloseable> closeablesInDestination = to.values().stream()
+				.filter(AutoCloseable.class::isInstance)
 				.map(AutoCloseable.class::cast)
 				.collect(Collectors.toSet());
 		if (to instanceof PipeLineSession toSession) {
@@ -169,8 +168,8 @@ public class PipeLineSession extends HashMap<String,Object> implements AutoClose
 	}
 
 	private static boolean shouldCloseSessionResource(final String key, final Object value) {
-		return value instanceof AutoCloseable &&
-			!key.startsWith(SYSTEM_MANAGED_RESOURCE_PREFIX);
+		return (value instanceof AutoCloseable && !key.startsWith(SYSTEM_MANAGED_RESOURCE_PREFIX)) &&
+				!(value instanceof Message message && message.isRequestOfType(String.class));
 	}
 
 	@Override
--- a/core/src/main/java/org/frankframework/core/PipeLineSession.java
+++ b/core/src/main/java/org/frankframework/core/PipeLineSession.java
@@ -17,6 +17,8 @@ package org.frankframework.core;
 
 import java.io.IOException;
 import java.time.Instant;
+import java.time.temporal.Temporal;
+import java.util.Date;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/frankframework/core/PipeLineSession.java
```diff
@@ -169,8 +168,8 @@ public Object put(String key, Object value) {
 	}
 
 	private static boolean shouldCloseSessionResource(final String key, final Object value) {
-		return value instanceof AutoCloseable &&
-			!key.startsWith(SYSTEM_MANAGED_RESOURCE_PREFIX);
+		return (value instanceof AutoCloseable && !key.startsWith(SYSTEM_MANAGED_RESOURCE_PREFIX)) &&
```
evandongen, line 171: Ik zou hier een aparte method van maken, `shouldNotClose(Object value)` oid, dat leest een stuk makkelijker weg

evandongen (changes requested):

core/src/main/java/org/frankframework/core/PipeLineSession.java
nielsm5: Moet deze nieuwe check ook niet in `scheduleCloseOnSessionExit(..)` ?

tnleeuw: Ja daar moet die eigenlijk ook in. Die had ik even over het hoofd gezien.

core/src/main/java/org/frankframework/core/PipeLineSession.java
```diff
@@ -169,8 +170,37 @@ public Object put(String key, Object value) {
...
+	}
+
+	/**
+	 * Check that the AutoCloseable value is of type {@link Message}, and if it is, that it does not contain a scalar or array value.
+	 * Scalar is defined as either {@link String}, {@link Number}, a {@link Date}, {@link Temporal} or {@link Boolean}.
+	 *
+	 * @param value AutoCloseable to check
+	 * @return {@code true} if {@code value} is not a {@link Message}, or if it is a Message with a request that is not a scalar or array type. Returns {@code false} otherwise.
+	 */
+	private static boolean isValueToBeClosed(AutoCloseable value) {
```
tnleeuw, line 186: I wonder if perhaps part of this should be moved into `Message` but until we need this check in multiple places, I decided against that.

core/src/main/java/org/frankframework/core/PipeLineSession.java
```diff
@@ -169,8 +170,37 @@ public Object put(String key, Object value) {
...
+		if (!(value instanceof Message message)) return true; // Should be closed, but is not a message
+		if (message.isNull()) return false; // Null message doesn't have to be closed
+		return !(message.isRequestOfType(String.class) ||
+				message.isRequestOfType(Number.class) ||
+				message.isRequestOfType(Date.class) ||
+				message.isRequestOfType(Temporal.class) ||
+				message.isRequestOfType(Boolean.class) ||
+				message.asObject().getClass().isArray()); // Arrays we have are mostly byte[] but I think all arrays should count, for simplicity.
```
tnleeuw, line 194: I don't think we should close `byte[]` either, just like we don't close `String` now, but there's no clean check for that.

The other types, I added them for completeness although I think that besides `Integer` and `Date` the others will not occur. `Integer` and `Date` could happen though, as far as I know.

evandongen (approved):

nielsm5 (approved):

PR #8333, opened 2025-01-31
Add a Larva test with PubSub domain

tnleeuw:
Backport the PubSub Larva test to 8.3.1 to verify that it also works with that version of the Frank!Framework.


pom.xml
```diff
@@ -19,18 +19,18 @@
 		<junit.excludedTags></junit.excludedTags>
 		<junit.includedTags></junit.includedTags>
 
-		<jackson.version>2.18.0</jackson.version>
-		<cxf.version>4.0.5</cxf.version>
-		<spring.version>6.1.13</spring.version>
+		<jackson.version>2.18.2</jackson.version>
+		<cxf.version>4.1.0</cxf.version>
+		<spring.version>6.2.2</spring.version>
```
tnleeuw, line 24: I don't know if all these dependency updates are needed but the Spring dependency update seemed to help for 9.0, and the other dependencies are a matter of "well why not, might as well update these".

Some Spring dependencies could however not be updated, they caused errors when updating, so I didn't spend too much time on figuring what parts of Spring were safe and updated only Spring Framework itself for 8.3.

evandongen (approved):

nielsm5 (approved):

PR #8335, opened 2025-01-31
Add a Larva test with PubSub domain

tnleeuw:
Backport of test to 9.0.1


pom.xml
```diff
@@ -21,10 +21,10 @@
 
 		<jackson.version>2.18.2</jackson.version>
 		<cxf.version>4.1.0</cxf.version>
-		<spring.version>6.2.1</spring.version>
+		<spring.version>6.2.2</spring.version>
```
tnleeuw, line 24: The updated dependencies seem to help making the test stable for all combinations, including WildFly + Artemis which was previously failing.

evandongen (approved):

nielsm5:
According to the JMS spec:

If you want your messages routed to…​ | Use this routing type …​
-- | --
A single queue within the matching address, in a point-to-point manner. | Anycast
Every queue within the matching address, in a publish-subscribe manner. | Multicast

Right now the default for topics seems to be Multicast, isn't that the desired setting?

nielsm5 (approved):

PR #8364, opened 2025-02-04
Try to make the PubSub Larva test more stable for all environments

tnleeuw:
Since the test doesn't seem to get stable for all environments, disable it by default. Can be manually enabled with `active.jmspubsub` property.


.gitattributes
```diff
@@ -27,6 +27,7 @@
 .classpath     text eol=lf
 .jsdtscope     text eol=lf
 org.eclipse.*  text eol=lf
+credentials.properties text eol=lf
```
nielsm5, line 30:
```suggestion
*.properties text eol=lf
```

.gitattributes
```diff
@@ -27,6 +27,7 @@
 .classpath     text eol=lf
 .jsdtscope     text eol=lf
 org.eclipse.*  text eol=lf
+*.properties text eol=lf
```
nielsm5, line 30:
```suggestion
org.eclipse.*  text eol=lf
*.properties   text eol=lf
```

PR #8378, opened 2025-02-07
Add whitelisting of HTTP params that are allowed to be copied into the session

nielsm5:
Not sure why #8371 cannot be used anymore? I dun-goof'd...


nielsm5 (commented):
Ik zag ook dat je hem gebruikt voor de multipart maar dit zit nog niet in de tests/doc 😏

tnleeuw:
I will create my own PR again for this branch so that it can actually be properly reviewed by you, Niels.

PR #8391, opened 2025-02-10
Introduce spring boot application test

evandongen:


test/src/test/java/org/frankframework/runner/IafTestInitializerTest.java
```diff
@@ -0,0 +1,23 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.io.IOException;
+
+import org.junit.jupiter.api.Test;
+import org.springframework.boot.SpringApplication;
+import org.springframework.context.ConfigurableApplicationContext;
+
+public class IafTestInitializerTest {
```
nielsm5, line 11: `@Tag("slow")`

PR #8545, opened 2025-03-06
Keep form field values in session

Matthbo:


philipsens:
The Cypress test that uses the Monaco Editor fails. It looks like it starts typing before the editor is really initialized or accepting of a value:
Test a Pipeline -- Test a Pipeline (failed)

philipsens:
> The Cypress test that uses the Monaco Editor fails. It looks like it starts typing before the editor is really initialized, accepting of a value or being overwritten by the empty value earlier submitted:

Resolved

philipsens (approved):

PR #8547, opened 2025-03-07
Improve Javadoc of pipes

thomaspj10:
Closes https://github.com/frankframework/frank-doc/issues/186

### DIFF
--- a/core/src/main/java/org/frankframework/pipes/FixedForwardPipe.java
+++ b/core/src/main/java/org/frankframework/pipes/FixedForwardPipe.java
@@ -127,19 +127,19 @@ public abstract class FixedForwardPipe extends AbstractPipe {
 	}
 
 	/**
-	 * If set, the processing continues directly at the forward of this pipe, without executing the pipe itself, if the input is empty
+	 * If {@code true}, the processing continues directly at the forward of this pipe, without executing the pipe itself, if the input is empty
 	 * @ff.default false
 	 */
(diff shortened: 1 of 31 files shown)


core/src/main/java/org/frankframework/pipes/FixedForwardPipe.java
```diff
@@ -127,19 +127,19 @@ public boolean skipPipe(Message input, PipeLineSession session) throws PipeRunEx
...
-	 * If set, the processing continues directly at the forward of this pipe, without executing the pipe itself, if the input is empty
+	 * If {@code true}, the processing continues directly at the forward of this pipe, without executing the pipe itself, if the input is empty
 	 * @ff.default false
 	 */
 	public void setSkipOnEmptyInput(boolean b) {
 		skipOnEmptyInput = b;
 	}
 
-	/** If set, this pipe is only executed when the value of parameter with name <code>ifParam</code> equals <code>ifValue</code> (otherwise this pipe is skipped) */
+	/** If {@code true}, this pipe is only executed when the value of parameter with name <code>ifParam</code> equals <code>ifValue</code> (otherwise this pipe is skipped) */
```
nielsm5, line 137:
```suggestion
	/** If set, this pipe is only executed when the value of parameter with name <code>ifParam</code> equals <code>ifValue</code> (otherwise this pipe is skipped) */
```

PR #8592, opened 2025-03-13
Remove asString from SendMessageJob

nielsm5:
No longer required since #8342


core/src/main/java/org/frankframework/scheduler/job/SendMessageJob.java
```diff
@@ -75,9 +73,10 @@ public void execute() throws JobExecutionException, TimeoutException {
 			localSender.start();
 			try (Message result = localSender.sendMessageOrThrow(toSendMessage, session)) {
 				// The result needs to be read, to make sure the StreamCaptureUtils sees this stream as read and will display the return value in Ladybug.
-				result.asString();
+				// This is part of the close method of captureReader and captureInputStream.
+				log.debug("SendMessageJob [{}] executed succesfully with result [{}]", this::getName, result::toString);
```
tnleeuw, line 77: We don't read the message anymore and the comment above still says the message needs to be read, for the StreamCaptureUtils. I'm not sure how the comment still applies, or if it is still relevant to keep it here if we now have a fix anyway in the StreamCaptureUtils?

evandongen (approved):

core/src/main/java/org/frankframework/scheduler/job/SendMessageJob.java
nielsm5: The comment is relevant, and the fix is in the `StreamCaptureUtils#captureReader` method.

I thought it would be handy to keep the comment so we'll know how it is/was supposed to work.

tnleeuw: And I think that in half a year or later we'll just be very confused by the comment... :-D

tnleeuw (approved):

PR #8613, opened 2025-03-17
Fix issue on health endpoints without authentication

evandongen:


security/src/main/java/org/frankframework/lifecycle/servlets/AbstractServletAuthenticator.java
```diff
@@ -193,6 +195,13 @@ public SecurityFilterChain configureHttpSecurity(HttpSecurity http) throws Excep
 			http.anonymous(AnonymousConfigurer::disable); // Disable the default anonymous filter and thus disallow all anonymous access
 		}
 
+		if (authorizeHealthEndpoints()) {
```
evandongen, line 198: This makes sure to only add the WebExpression when authentication is enabled.

console/backend/src/test/java/org/frankframework/console/configuration/SecurityChainConfigurerWithoutAuthenticationTest.java
```diff
@@ -0,0 +1,75 @@
...
+import org.springframework.test.web.servlet.setup.MockMvcBuilders;
+import org.springframework.web.context.WebApplicationContext;
+
+import org.frankframework.console.controllers.ServerStatistics;
+import org.frankframework.console.controllers.WebTestConfiguration;
+
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = { SecurityChainTestConfiguration.class, SecurityChainConfigurer.class, WebTestConfiguration.class, ServerStatistics.class })
+@WebAppConfiguration
+public class SecurityChainConfigurerWithoutAuthenticationTest {
```
evandongen, line 26: Extra unittest class to test the scenario _without_ authentication enabled.

security/src/main/java/org/frankframework/lifecycle/servlets/AbstractServletAuthenticator.java
```diff
@@ -193,6 +195,13 @@ public SecurityFilterChain configureHttpSecurity(HttpSecurity http) throws Excep
...
+			http.authorizeHttpRequests(authorize -> authorize
+					.requestMatchers(new AntPathRequestMatcher("/**/health"))
+					.access(new WebExpressionAuthorizationManager(EXPRESSION_IS_LOCALHOST_OR_AUTHENTICATED))
+			);
```
nielsm5, line 202: I think it would still be best if this code block is in the console SecurityChainConfigurer, if somebody creates an api/rest endpoint that ends with health it's now unsecure on localhost?

evandongen: Changed as discussed.

nielsm5 (approved):

tnleeuw (approved):

PR #8625, opened 2025-03-18
Improve flow modal buttons (better error feedback)

Matthbo:
Firefox can now open flows in a new tab, downloading or copying too big of a flow image still doesn't work. I give up on trying to put more energy into it.


philipsens:
Could this benefit from an e2e test?

nielsm5:
> Could this benefit from an e2e test?

Probably yes, but I think we've spent enough time on this for now.

PR #8689, opened 2025-04-01
Add inheritdoc to jms class structure

### LINKED ISSUE #2857: Frank!Doc documentation less than Javadoc documentation on schema and schemaLocation of XmlValidator
**Describe the issue**
The page https://frankdoc.frankframework.org/#!/Validators/XmlValidator shows the following info:

schema | the filename of the schema on the classpath. see doc on the method. (effectively the same as noNamespaceSchemaLocation) | schemaLocation | Pairs  of uri references (one for the namespace name, and one for a hint as to  the location of a schema document defining names for that namespace  name). see doc on the method.

While the Javadoc on the same methods contain more info which the Frank developer will not see:

	 * <p>
	 * The filename of the schema on the classpath. It is not possible to specify a
	 * namespace using this attribute.
	 * <p>
	 * An example value would be "xml/xsd/GetPartyDetail.xsd"
	 * </p>
	 * <p>
	 * The value of the schema attribute is only used if the schemaLocation
	 * attribute and the noNamespaceSchemaLocation are not set
	 * </p>

	 * <p>Pairs of URI references (one for the namespace name, and one for a
	 * hint as to the location of a schema document defining names for that
	 * namespace name).</p>
	 * <p> The syntax is the same as for schemaLocation attributes
	 * in instance documents: e.g, "http://www.example.com file%20name.xsd".</p>
	 * <p>The user can specify more than one XML Schema in the list.</p>
	 * <p><b>Note</b> that spaces are considered separators for this attributed.
	 * This means that, for example, spaces in filenames should be escaped to %20.
	 * </p>

Can all the available info be exposed to the Frank developers? Maybe the same is true for other attributes, pipes, sender, ...? Do we need somebody to go through all the available Javadoc and Frank!Doc?

**Reporter**
Jaco de Groot

evandongen:
closes #2857

Created https://github.com/frankframework/frank-doc/issues/333 and https://github.com/frankframework/frank-doc/issues/334 in the process.

### DIFF
--- a/nn-specials/src/main/java/org/frankframework/extensions/esb/EsbJmsTransactionalStorage.java
+++ b/nn-specials/src/main/java/org/frankframework/extensions/esb/EsbJmsTransactionalStorage.java
@@ -65,6 +65,8 @@ import org.frankframework.util.UUIDUtil;
  * </table>
  * </p>
  *
+ * {@inheritDoc}
+ *
  * @author Peter Leeuwenburgh
  */
 @Category(Category.Type.NN_SPECIAL)
--- a/nn-specials/src/main/java/org/frankframework/extensions/esb/EsbJmsTransactionalStorage.java
+++ b/nn-specials/src/main/java/org/frankframework/extensions/esb/EsbJmsTransactionalStorage.java
@@ -41,7 +41,7 @@ import org.frankframework.util.TransformerPool;
 import org.frankframework.util.UUIDUtil;
 
 /**
- * ESB (Enterprise Service Bus) extension of JmsTransactionalStorage.
+ * ESB (Enterprise Service Bus) extension of {@code JmsTransactionalStorage}.
  *
  * <p>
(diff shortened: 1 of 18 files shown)


nn-specials/src/main/java/org/frankframework/extensions/esb/EsbJmsTransactionalStorage.java
```diff
@@ -53,17 +53,8 @@
  * ESB.Infrastructure.US.Log.BusinessLog.2.AuditLog.1.Action</li>
  * </ul>
  * </p>
- * <p>
- * <b>Configuration </b><i>(where deviating from
```
evandongen, line 57: Zinloze table heb ik maar weggehaald

tnleeuw (approved):

nielsm5 (approved):

PR #8747, opened 2025-04-07
Update transitive cxf dependencies

evandongen:
Try again for all ladybug dependencies and update netty as well.

### DIFF
--- a/ladybug/common/pom.xml
+++ b/ladybug/common/pom.xml
@@ -12,19 +12,6 @@
 	<artifactId>frankframework-ladybug-common</artifactId>
 	<name>Frank!Framework Ladybug Commons</name>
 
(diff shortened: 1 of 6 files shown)


ladybug/common/pom.xml
```diff
@@ -12,6 +12,19 @@
...
+		<dependencies>
+			<dependency>
+				<groupId>org.apache.cxf</groupId>
+				<artifactId>cxf-bom</artifactId>
+				<version>4.0.6</version>
+				<type>pom</type>
+				<scope>import</scope>
+			</dependency>
+		</dependencies>
+	</dependencyManagement>
```
nielsm5, line 26: Won't it be easier to add it to the ladybug parent pom? That way you don't need to define it in the 3 ladybug modules.

nielsm5 (approved):

PR #8761, opened 2025-04-08
Fix ladybug SecurityFilterChain

nielsm5:
By removing the authenticator, I naively also removed the SecurityFilterChain.
See https://github.com/frankframework/frankframework/pull/8532

This caused all authentication to go through the default (Console's) SecurityFilter, which enables CSRF. The Ladybug does not support this feature, so by reverting the change from #8532 partially this should be fixed again.


ladybug/common/src/main/java/org/frankframework/ladybug/config/LadybugSecurityChainConfigurer.java
nielsm5, line 1: The change is that this file has been split up in 2, one for the ladybug-common and one remains in the debugger.

tnleeuw (approved):

PR #8768, opened 2025-04-09
Fix infinite stacktrace that leads to stackoverflow

tnleeuw:


core/src/main/java/org/frankframework/xml/SaxException.java
```diff
@@ -36,7 +36,7 @@ public SaxException(String message, Exception e) {
 		try {
 			initCause(e); // this fixes stacktrace under IBM JDK, does nothing under standard JDK
 		} catch (Exception e2) { // Jboss throws 'IllegalStateException: Can't overwrite cause'
-			addSuppressed(e2);
+			addSuppressed(e);
```
evandongen, line 39: Idem hier:
```suggestion
			addSuppressed(e); // Add the original exception as suppressed exception, or get a stackoverflow in IbisException#getMessages
```

evandongen (approved):

PR #8790, opened 2025-04-11
Fix some code inspection issues

evandongen:
When doing the Snyk updates recently, I was using the 'Problems' tab in IntelliJ - some interesting things came up after analysis. In this PR I've modified the following:
* use final and/or static where possible
* isEmpty instead of length > 0
* no c-style arrays
* don't use Optionals as a parameter

### DIFF
--- a/core/src/main/java/org/frankframework/pipes/ForEachChildElementPipe.java
+++ b/core/src/main/java/org/frankframework/pipes/ForEachChildElementPipe.java
@@ -159,7 +159,7 @@ public class ForEachChildElementPipe extends StringIteratorPipe implements IThre
 		private final ItemCallback callback;
 
 		private XmlWriter xmlWriter;
-		private Exception rootException=null;
+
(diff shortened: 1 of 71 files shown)


core/src/main/java/org/frankframework/pipes/ForEachChildElementPipe.java
```diff
@@ -215,8 +215,9 @@ public void endNode(String uri, String localName, String qName) throws SAXExcept
 
 		private void checkInterrupt() throws SAXException {
 			if (Thread.currentThread().isInterrupted()) {
-				rootException = new InterruptedException("Thread has been interrupted");
+				Exception rootException = new InterruptedException("Thread has been interrupted");
 				rootException.fillInStackTrace();
```
nielsm5, line 219: Dit kan weg toch? We doen er niets mee...

tnleeuw (approved):

PR #8817, opened 2025-04-15
Wire jobs via custom spring context

nielsm5:
One of the last steps to fully let Spring manage our lifecycle by simple sending events to the correct context. Enables CRaC, jobs that are managed via Spring, removes redundant code.

Next step would be to consolidate some methods that are now considered duplicates.


core/src/main/java/org/frankframework/scheduler/ScheduleManager.java
nielsm5, line 1: Deze file is verplaatst...

core/src/main/java/org/frankframework/scheduler/SchedulerHelper.java
```diff
@@ -184,13 +183,4 @@ public void deleteTrigger(String name, String group) throws SchedulerException {
 
 		getScheduler().unscheduleJob(key);
 	}
-
-	public void startScheduler() throws SchedulerException {
```
nielsm5, line 188: Is managed by Spring, this method is redundant...

core/src/main/java/org/frankframework/scheduler/SchedulerHelper.java
```diff
@@ -33,18 +32,18 @@
...
 import lombok.Setter;
+import lombok.extern.log4j.Log4j2;
 
 import org.frankframework.configuration.ConfigurationException;
 import org.frankframework.scheduler.job.IJob;
-import org.frankframework.util.LogUtil;
 
 /**
- * The SchedulerHelper encapsulates the quarz scheduler.
+ * The SchedulerHelper encapsulates the Quarz scheduler.
```
tnleeuw, line 41:
```suggestion
 * The SchedulerHelper encapsulates the Quartz scheduler.
```

PR #8857, opened 2025-04-21
Fix TransactionalStore upload action for resendReceiverMessage

nielsm5:


console/frontend/pom.xml
```diff
@@ -34,6 +34,11 @@
 			<artifactId>jakarta.servlet-api</artifactId>
 			<scope>provided</scope>
 		</dependency>
+
+		<dependency>
+			<groupId>org.junit.jupiter</groupId>
+			<artifactId>junit-jupiter-engine</artifactId>
+		</dependency>
```
nielsm5, line 41: Aparently needed because of surefire... It doesn't need to do anything but has to be on the classpath.

console/backend/src/main/java/org/frankframework/console/controllers/TransactionalStorage.java
```diff
@@ -162,7 +162,7 @@ public ResponseEntity<?> browseMessages(
 	@PutMapping(value = "/configurations/{configuration}/adapters/{adapterName}/receivers/{receiverName}/stores/Error/messages/{messageId}", produces = MediaType.APPLICATION_JSON_VALUE)
 	public ResponseEntity<?> resendReceiverMessage(TransactionStoragePathVariables path) {
 
-		RequestMessageBuilder builder = RequestMessageBuilder.create(BusTopic.MESSAGE_BROWSER, BusAction.STATUS);
+		RequestMessageBuilder builder = RequestMessageBuilder.create(BusTopic.MESSAGE_BROWSER, BusAction.UPLOAD);
```
nielsm5, line 165: Actual fix.

evandongen (approved):

PR #8858, opened 2025-04-21
Refactor InputOutputPipeProcessor

nielsm5:
This PR started off as me trying to add more test coverage and reduce the brain method in the `InputOutputPipeProcessor`.
I noticed some leaks in the leak log and added the values to the closeables list, however since they are String based messages, they are never closed.

Not sure what the way forward is here, do we trust the leak log, do we ignore String messages in there? Do we call close on Message but not set the value to null??


tnleeuw:
We don't know if the messages are always String messages, right? So we should close them.
But the leak-log should perhaps ignore string messages, if that is possible. They're never closed anymore, so they lead to false positives.

tnleeuw (approved):

nielsm5:
> We don't know if the messages are always String messages, right? So we should close them. But the leak-log should perhaps ignore string messages, if that is possible. They're never closed anymore, so they lead to false positives.

Some things, such as getInputFromFixedValue is always a `String`. The same goes for the on-empty-input replacement.
It's not really a big deal to call `closeOnClose` though 😄 .

PR #8885, opened 2025-04-25
Wait on IafTestInitializer to be healthy before running e2e tests

philipsens:
Use the `OutboundGateway` to gather health information and verify if the Framework has started (and loaded configurations).

After verification, the Test Container for Cypress will be started.

### DIFF
--- a/test/src/test/java/org/frankframework/runner/RunCypressE2eTest.java
+++ b/test/src/test/java/org/frankframework/runner/RunCypressE2eTest.java
@@ -16,9 +16,12 @@
 
 package org.frankframework.runner;
 
+import static org.awaitility.Awaitility.await;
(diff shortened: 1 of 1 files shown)


evandongen (approved):

philipsens:
In the GitHub Action:

> Error:  Failures: 
Error:    RunCypressE2eTest.lambda$createContainerFromSuite$2:139 AssertionError: Timed out retrying after 4000ms: expected 'Configuration [IAF_Util] LiquiBase applied [9] change(s) and added tag [FF_Util:8:Yannick van Leeuwen]Configuration [IAF_Util] startup in 2750 ms' to match /Application \[.+\] startup in \d+ ms/ ==> expected: <true> but was: <false>

I need to check if this is the correct health check.

PR #8961, opened 2025-05-12
Disable testtool echo2 by default

Matthbo:


commons/src/main/java/org/frankframework/condition/AppConstantCondition.java
```diff
@@ -0,0 +1,42 @@
...
+
+public class AppConstantCondition implements Condition {
+
+	@Override
+	public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
+		Map<String, Object> attributes = metadata.getAnnotationAttributes(ConditionalOnAppConstants.class.getName());
+		if (attributes == null)
+			return false;
+
+		if (!attributes.containsKey("name") || !attributes.containsKey("value") || !attributes.containsKey("defaultValue"))
```
tnleeuw, line 32: Will you always need to set `defaultValue` on the annotation?

tnleeuw (approved):

commons/src/main/java/org/frankframework/condition/AppConstantCondition.java
evandongen: +1

evandongen (approved):

commons/src/main/java/org/frankframework/condition/AppConstantCondition.java
Matthbo: Added a default to defaultValue in the annotation and remove the need for defaultValue to be set in the attributes map

tnleeuw (approved):

PR #8980, opened 2025-05-14
Make reading of scenario steps more flexible

tnleeuw:
Scenario step numbers can now skip steps, so you can comment out a single step without having to renumber all following steps.

Step numbers can also start with a '0' to align step numbers nicely in the file.

### DIFF
--- a/larva/src/test/java/org/frankframework/larva/ScenarioTest.java
+++ b/larva/src/test/java/org/frankframework/larva/ScenarioTest.java
@@ -42,10 +42,11 @@ class ScenarioTest {
 		List<String> steps = scenario.getSteps(config);
 
 		// Assert
-		assertEquals(3, steps.size());
+		assertEquals(4, steps.size());
 		assertEquals("step1.action.read", steps.get(0));
 		assertEquals("step2.action.write", steps.get(1));
 		assertEquals("step3.action.writeline", steps.get(2));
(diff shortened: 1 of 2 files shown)


philipsens:
Nice, this will make it a lot easier to write and test Larva scenarios!

larva/src/test/java/org/frankframework/larva/ScenarioTest.java
```diff
@@ -143,10 +145,11 @@ void getSteps5() throws IOException {
 		List<String> steps = scenario.getSteps(config);
 
 		// Assert
-		assertEquals(3, steps.size());
+		assertEquals(4, steps.size());
```
philipsens, line 148: Change comment that it is possible to skip

larva/src/test/java/org/frankframework/larva/ScenarioTest.java
```diff
@@ -86,12 +87,13 @@ void getSteps3() throws IOException {
 						step04.action.readline=inline
 				""";
 		Scenario scenario = createScenario(scenarioSteps);
+		config.setAllowReadlineSteps(false);
```
philipsens, line 90: Change the comment so it states that the steps can be started with 0

evandongen (approved):

PR #9000, opened 2025-05-16
Unix timestamp

vdegans:
* Adds the option retrieve a unix timestamp when using the {now} pattern.
* Adds the option to parse a unix timestamp to a date object within a param.


core/src/main/java/org/frankframework/parameters/AbstractParameter.java
```diff
@@ -648,7 +648,9 @@ private Object getValueForFormatting(ParameterValueList alreadyResolvedParameter
 				case "now":
 					if ("date".equalsIgnoreCase(formatType) || "time".equalsIgnoreCase(formatType)) {
 						substitutionValue = new Date();
-					} else{
+					} else if ("number".equalsIgnoreCase(formatType)) {
```
nielsm5:
```suggestion
					} else if ("millis".equalsIgnoreCase(formatType)) {
```
?? En niet vergeten dit te documenteren :)

nielsm5 (approved):

PR #9025, opened 2025-05-21
Fix MQTT resources

nielsm5:


messaging/src/main/java/org/frankframework/extensions/mqtt/MqttClientFactory.java
```diff
@@ -58,14 +59,12 @@ protected MqttClient map(MqttClientSettings data) {
 
 		String clientId = data.getClientId();
 		if (StringUtils.isEmpty(clientId)) {
-			clientId = AppConstants.getInstance().getProperty("transactionmanager.uid");
+			clientId = Misc.getHostname()+"-"+ UUIDUtil.createSimpleUUID();
```
nielsm5, line 62: Not always set, so it was an unreliable property...

messaging/src/main/java/org/frankframework/extensions/mqtt/MqttClientFactory.java
```diff
@@ -58,14 +59,12 @@ protected MqttClient map(MqttClientSettings data) {
...
 		}
 
-		try (MqttClient client = new MqttClient(data.getUrl(), clientId, getMqttDataStore(data.getPersistenceDirectory()))) {
-			client.connect(connectOptions);
-
-			return client;
-		}
+		MqttClient client = new MqttClient(data.getUrl(), clientId, getMqttDataStore(data.getPersistenceDirectory()));
+		client.connect(connectOptions);
+		return client;
```
nielsm5, line 67: Should not be closed!

core/src/main/java/org/frankframework/jdbc/datasource/ObjectFactory.java
```diff
@@ -87,7 +88,8 @@ protected O map(P object) {
 	 * If not cached yet, attempts to traverse all {@link IObjectLocator IObjectLocators} to do so.
 	 */
 	protected final O get(String name, Properties environment) {
-		return objects.computeIfAbsent(name, k -> compute(k, environment));
+		String nameWithResourcePrefix = StringUtils.prependIfMissing(name, resourcePrefix+"/");
```
nielsm5, line 91: Allow omitting the default resource prefix.

tnleeuw (approved):

PR #9074, opened 2025-05-29
Fix many small issues

nielsm5:
Also see https://github.com/wearefrank/ladybug/issues/480


pom.xml
```diff
@@ -23,7 +23,7 @@
 		<jackson.version>2.19.0</jackson.version>
 		<cxf.version>4.1.2</cxf.version>
 		<spring.version>6.2.7</spring.version>
-		<spring-security.version>6.5.0</spring-security.version>
+		<spring-security.version>6.4.5</spring-security.version>
```
tnleeuw, line 26: Toch weer downgrade naar 6.4.5 nodig?

PR #9107, opened 2025-06-07
Make HTTP timeout work based on the HttpSession (sharedResources)

nielsm5:
I've also added a scenario for the `CallConcurrentSharedHTTPResource` adapter, but I cant figure out a proper ignore rule..

### DIFF
--- a/core/src/main/java/org/frankframework/http/AbstractHttpSession.java
+++ b/core/src/main/java/org/frankframework/http/AbstractHttpSession.java
@@ -700,7 +700,7 @@ public abstract class AbstractHttpSession implements ConfigurableLifecycle, HasK
 
 		preAuthenticate(context);
 
-		final int hardTimeout; // Since the Apache HttpComponents library does not support hard timeouts we have to abord the request our selves.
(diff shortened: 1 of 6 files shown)


tnleeuw (approved):

core/src/main/java/org/frankframework/http/AbstractHttpSession.java
```diff
@@ -679,14 +681,48 @@ protected SSLConnectionSocketFactory getSSLConnectionSocketFactory() throws Conf
...
-
 		CloseableHttpClient client = getHttpClient();
+
 		HttpClientContext context = httpClientContext != null ? httpClientContext : getOrCreateHttpClientContext(client, session);
+
 		preAuthenticate(context);
-		log.trace("executing request using HttpClient [{}] and HttpContext [{}]", client::hashCode, () -> context);
-		return client.execute(targetHost, httpRequestBase, context);
+
+		final int hardTimeout; // Since the Apache HttpComponents library does not support hard timeouts we have to abord the request our selves.
```
evandongen:
```suggestion
		final int hardTimeout; // Since the Apache HttpComponents library does not support hard timeouts we have to abort the request our selves.
```

evandongen (approved):

PR #9110, opened 2025-06-08
Add method to find expired certificates

nielsm5:
The only thing I'm not quite happy with is the `SecurityItems.java` code.
Ideally all pipes are singletons in the AdapterContext, which could then do a lookup (via the beanfactory) for certain interfaces.
Alternatively one could argue that resources should be found via the AdapterContext as well, instead of keeping a copy of the thread in each bean.


evandongen:
> The only thing I'm not quite happy with is the `SecurityItems.java` code. Ideally all pipes are singletons in the AdapterContext, which could then do a lookup (via the beanfactory) for certain interfaces. Alternatively one could argue that resources should be found via the AdapterContext as well, instead of keeping a copy of the thread in each bean.
> 
> If this, at least for now, is the way forward, some minor cleanup is required, as well as making it possible to configure the duration via a property.

I've been looking at the code and I don't think we can easily simplify this. You can contain this in one stream though:

```
				String name = "%s/%s".formatted(config.getName(), adapter.getName());
				List<String> list = adapter.getPipeLine()
						.getPipes()
						.stream()
						.filter(this::hasKeyOrTrustStore)
						.map(FrankElement.class::cast)
						.map(this::extractSender)
						.map(this::extractKeyStore)
						.filter(Objects::nonNull)
						.map(this::getExpiringCertificatesList)
						.flatMap(List::stream)
						.toList();

				expiringCertificates.add(new KeyStoreDTO(name, list));
```

And `this::getExpiringCertificatesList` is the extracted inline method from before.

nielsm5:
So are we going to continue using this approach? Or are we doing to register all pipes in the AdapterContext? Perhaps not the best approach as we're trying to stabilize the code, which is why I thought this was the least impactful solution.

tnleeuw:
Pipes are not singletons. There can be multiple instances of each pipe within an adapter, each with different configuration.

I don't know if it changes much for your idea that pipes are not singletons?

nielsm5:
> Pipes are not singletons. There can be multiple instances of each pipe within an adapter, each with different configuration.
> 
> I don't know if it changes much for your idea that pipes are not singletons?

Yea, you can still make 'unique' singles singletons, just like how Adapters are now registered in a Configuration.

core/src/main/java/org/frankframework/encryption/PkiUtil.java
```diff
@@ -217,49 +245,82 @@ public static PrivateKey getPrivateKey(HasKeystore keystoreOwner, String purpose
...
+		return certificates;
+	}
+
+	private static boolean isDueToExpire(Certificate cert, Instant dateAfterWhichCertsAreExpired) {
+		if (cert instanceof X509Certificate x509Cert) {
+			Date notAfter = x509Cert.getNotAfter();
+			return dateAfterWhichCertsAreExpired.isBefore(notAfter.toInstant());
+		}
+
+		return false;
```
evandongen, line 321: Is dit nog een uitzonderingssituatie waar we iets mee moeten?

evandongen (approved):

core/src/main/java/org/frankframework/encryption/PkiUtil.java
nielsm5: Het lijkt mij niet.

tnleeuw (approved):

Issue #8577, opened 2025-03-12
Create LDAP ResourceFactory


nielsm5:
### Related Problem or Business Case

Allow LDAP connections to be provided via JDNI as well as resources.yml

### Describe the Feature

This will allow for the removal of the JndiBase, see #8215.

### Collaboration with External Companies

No, community resolution only

evandongen:
Since LDAP and JNDI are more intertwined than expected, we leave this as is for now.

PR #9120, opened 2025-06-10
deprecate providerUrl and jmsRealm

evandongen:
As discussed with Niels during development for https://github.com/frankframework/frankframework/issues/8577, we decided that was a bit too much for now and we should start smaller, by deprecating the jmsRealm and providerUrl for now.

### DIFF
--- a/core/src/main/java/org/frankframework/jndi/JndiBase.java
+++ b/core/src/main/java/org/frankframework/jndi/JndiBase.java
@@ -17,7 +17,6 @@ package org.frankframework.jndi;
 
 import java.io.IOException;
 import java.net.URL;
-import java.util.Iterator;
 import java.util.Properties;
 
 import javax.naming.Context;
@@ -33,6 +32,7 @@ import lombok.Getter;
 import lombok.Setter;
 
 import org.frankframework.configuration.ConfigurationException;
+import org.frankframework.configuration.ConfigurationWarning;
 import org.frankframework.core.HasApplicationContext;
 import org.frankframework.core.IConfigurable;
 import org.frankframework.core.NameAware;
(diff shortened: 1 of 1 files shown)


messaging/src/main/java/org/frankframework/jms/JMSFacade.java
```diff
@@ -84,7 +90,12 @@
  *
  * @author 	Gerrit van Brakel
  */
-public class JMSFacade extends JndiBase implements HasPhysicalDestination, IXAEnabled, Lifecycle {
+public class JMSFacade extends JndiBase implements ConfigurableLifecycle, FrankElement, NameAware, HasPhysicalDestination, IXAEnabled {
```
nielsm5, line 93: The JMSFacade now no longer uses the JndiBase for any setters other then the jndi specific ones.

nielsm5 (approved):

core/src/main/java/org/frankframework/jndi/JndiBase.java
```diff
@@ -130,9 +111,9 @@ public Properties getJndiEnv() throws NamingException {
 		}
 
 		if (log.isDebugEnabled()) {
-			for(Iterator it=jndiEnv.keySet().iterator(); it.hasNext();) {
-				String key=(String) it.next();
-				String value=jndiEnv.getProperty(key);
+			for (Object object : jndiEnv.keySet()) {
+				String key = (String) object;
```
tnleeuw, line 115:
```suggestion
			for (String key : jndiEnv.stringPropertyNames()) {
```

tnleeuw (approved):

core/src/main/java/org/frankframework/jndi/JndiBase.java
nielsm5: Tja, gaat toch weg 😄

PR #9133, opened 2025-06-11
Fix Console error page & logging page on framework startup error

Matthbo:


console/frontend/src/main/frontend/src/app/services/auth.service.ts
```diff
@@ -15,14 +15,32 @@ import {
...
+    },
+    {
+      name: 'getLogDirectory',
+      rel: 'logging',
+      description: 'view files/folders inside the log directory',
+      href: 'http://localhost:8080/iaf-test/iaf/api/logging',
+      type: 'GET',
+      roles: ['IbisObserver', 'IbisDataAdmin', 'IbisAdmin', 'IbisTester'],
+    },
+  ];
```
philipsens: Maybe it would be nicer to use a partial for the Link so it would not need all the (kind of incorrect?) metadata.

philipsens (approved):

PR #9144, opened 2025-06-15
Remove silly formatException class from MessageHandler

nielsm5:


core/src/main/java/org/frankframework/http/PushingListenerAdapter.java
```diff
@@ -109,13 +110,23 @@ public Message processRequest(Message rawMessage, PipeLineSession session) throw
...
 
+	// The ApplicationContext is practically always an Adapter except when the listener is created directly via the LarvaScenarioContext
+	private Message formatExceptionUsingErrorMessageFormatter(PipeLineSession session, Message inputMessage, Throwable t) {
+		if (applicationContext instanceof Adapter adapter) {
+			return adapter.formatErrorMessage(null, t, inputMessage, session, null);
+		}
+
+		log.warn("unformatted exception while processing input request [{}]", inputMessage, t);
+		return new Message(t.getMessage());
+	}
```
nielsm5, line 130: Although over here we can safely assume the `ApplicationContext` is always an `Adapter`, I've copied this block from the `JavaListener`.

core/src/main/java/org/frankframework/receivers/JavaListener.java
```diff
@@ -177,7 +178,7 @@ private Message processRequest(@Nonnull MessageWrapper<M> messageWrapper, @Nonnu
 					} catch (ListenerException e) {
 						// Message with error contains a String so does not need to be preserved.
 						// (Trying to preserve means dealing with extra IOException for which there is no reason here)
-						return handler.formatException(null, session, message, e);
+						return formatExceptionUsingErrorMessageFormatter(session, message, e);
```
nielsm5: I think this is quite strange, this listener should always just throw the exception, it's up to the `callee` to format it using the correct `ErrorMessageFormatter` if any.

core/src/main/java/org/frankframework/core/IMessageHandler.java
```diff
@@ -30,17 +30,15 @@ public interface IMessageHandler<M> {
...
 	 */
 	void processRawMessage(IListener<M> origin, RawMessageWrapper<M> message, PipeLineSession session, boolean duplicatesAlreadyChecked) throws ListenerException;
 
 	/**
-	 * Alternative to functions above, will NOT use {@link IListener#extractMessage}. Used by PushingListeners.
+	 * Alternative to functions above, will NOT use {@link IListener#extractMessage}.
+	 * TODO Remove pointless `RawMessage` wrapping and unwrapping? Method should have either a `message` or a `rawMessage`.
 	 */
-	Message processRequest(IListener<M> origin, RawMessageWrapper<M> rawMessage, Message message, PipeLineSession session) throws ListenerException;
+	Message processRequest(IPushingListener<M> origin, RawMessageWrapper<M> rawMessage, Message message, PipeLineSession session) throws ListenerException;
```
nielsm5: I agree this ain't perfect, but for now we are limiting the code-execution flow in implementing to call the correct 'entrypoint' for the receiver.

tnleeuw (approved):

tnleeuw:
Initially looks good except a stupid test that fails!

PR #9173, opened 2025-06-19
Mermaid flow zooming & panning

Matthbo:
Panning & zooming now works inside mermaid flow modals
Also made the 'New tab', 'Copy' & 'Download' buttons work with dynamic viewboxes

image

### DIFF
--- a/console/frontend/src/main/frontend/src/app/views/adapterstatistics/adapterstatistics.component.html
+++ b/console/frontend/src/main/frontend/src/app/views/adapterstatistics/adapterstatistics.component.html
@@ -114,7 +114,7 @@
                     </thead>
                     @for (receiver of stats.receivers; track receiver.name) {
                       <tbody>
-                        @for (threadsProcessing of receiver.processing; track threadsProcessing) {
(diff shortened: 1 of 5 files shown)


console/frontend/src/main/frontend/src/app/components/zoom-pan.directive.ts
```diff
@@ -0,0 +1,145 @@
...
+      x: fixedPointFloat(this.viewBox.x + dx),
+      y: fixedPointFloat(this.viewBox.y + dy),
+      w: this.viewBox.w,
+      h: this.viewBox.h,
+    };
+    this.applyViewBox(movedViewBox);
+    return movedViewBox;
+  }
+
+  applyViewBox(viewBox: ViewBox): void {
```
philipsens: I can see why the SVG should have a non-null assertion. The scale method is applied on the SVG itself. And the resize method checks if it is not null. Still, I think this could be a bit stricter. Making the method private would be better. But I think the assertion could also be inverted. `this.svg?.setAttribute` than it won't fail if it is misused.

Matthbo:
Stupid docker image tests failing :<

PR #9221, opened 2025-06-27
Modified bearer only authentication to expect roles in the token

evandongen:
This is needed for the authentication in the framework. Extended the integration test for this, to actually reach an endpoint which needs a role. 
I had some struggles with keycloak, since you need to stop keycloak to do a complete export. But there isn't a command to stop keycloak. I managed to export users, but those were missing the roles, which I actually need in the test. After some tweaking in the import file I managed to figure out the syntax to actually provide a user _with_ a role mapping.

### DIFF
--- a/security/src/main/java/org/frankframework/lifecycle/servlets/BearerOnlyAuthenticator.java
+++ b/security/src/main/java/org/frankframework/lifecycle/servlets/BearerOnlyAuthenticator.java
@@ -56,8 +56,6 @@ import lombok.Setter;
  */
 public class BearerOnlyAuthenticator extends AbstractServletAuthenticator {
 
-	private static final String DEFAULT_ROLE_PREFIX = "ROLE_";
-
 	@Setter
 	private String issuerUri;
 
--- a/security/src/main/java/org/frankframework/lifecycle/servlets/BearerOnlyAuthenticator.java
+++ b/security/src/main/java/org/frankframework/lifecycle/servlets/BearerOnlyAuthenticator.java
@@ -15,11 +15,21 @@
 */
 package org.frankframework.lifecycle.servlets;
 
+import java.util.Collection;
+import java.util.Map;
+import java.util.stream.Collectors;
+
 import org.apache.commons.lang3.StringUtils;
+import org.springframework.core.convert.converter.Converter;
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
+import org.springframework.security.core.GrantedAuthority;
+import org.springframework.security.core.authority.SimpleGrantedAuthority;
+import org.springframework.security.oauth2.jwt.Jwt;
 import org.springframework.security.oauth2.jwt.JwtDecoder;
 import org.springframework.security.oauth2.jwt.JwtDecoders;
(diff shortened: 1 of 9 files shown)


security/src/main/java/org/frankframework/lifecycle/servlets/BearerOnlyAuthenticator.java
```diff
@@ -34,36 +44,117 @@
...
+	@Setter
+	private String authoritiesClaimName;
+
 	@Override
 	public SecurityFilterChain configure(HttpSecurity http) throws Exception {
 		if (StringUtils.isAllBlank(issuerUri, jwkSetUri)) {
 			throw new IllegalArgumentException("Configuring issuerUri and/or jwkSetUri is mandatory to use BearerOnlyAuthenticator");
 		}
 
+		if (StringUtils.countMatches(issuerUri, ".") > 1) {
```
evandongen: Create a unit test for this

test/src/test/resources/test-realm.json
```diff
@@ -1,2508 +1,2468 @@
...
+      "disableableCredentialTypes" : [ ],
+      "requiredActions" : [ ],
+      "notBefore" : 0,
+      "access" : {
+        "manage" : true
+      },
+      "credentials": [
+        {"type": "password", "value": "testuser"}
+      ],
+      "realmRoles": [
```
evandongen, line 489: So this was what I needed to fix the test :-/

security/src/main/java/org/frankframework/lifecycle/servlets/BearerOnlyAuthenticator.java
```diff
@@ -34,36 +44,117 @@
...
+ * application.security.console.authentication.preferredUsernameClaimName=preferred_username
+ * application.security.console.authentication.authoritiesClaimName=realm_access.roles
+ * }</pre>
+ * </p>
  *
  * @author evandongen
  */
 public class BearerOnlyAuthenticator extends AbstractServletAuthenticator {
 
+	private static final String DEFAULT_ROLE_PREFIX = "ROLE_";
```
nielsm5: Deze staat ook in de noop authenticator. misschien dat die twee naar de abstracte class kunnen?

security/src/main/java/org/frankframework/lifecycle/servlets/BearerOnlyAuthenticator.java
```diff
@@ -34,36 +44,117 @@
...
 	@Setter
 	private String jwkSetUri;
 
+	/**
+	 * The claim name in the JWT token that contains the preferred username of the user.
+	 * Defaults to "sub", which is the standard claim for subject identifier. But, when using Keycloak, it is common to use "preferred_username" instead.
+	 * @see "JwtAuthenticationConverter#principalClaimName"
+	 */
+	@Setter
+	private String preferredUsernameClaimName;
```
nielsm5: In de OAuth2Authenticator heet dit ding `userNameAttributeName`. Ik twijfel een beetje of dat nou een duidelijke naam is maar misschien is het wel handig om ze gelijk te houden?

PR #9280, opened 2025-07-09
Fix time-dependent tests

tnleeuw:
There are many more places where `Instant.now()` could be replaced with call to TimeProvider, and same for `Date.now()` which isn't used yet.

For now however this fixes some broken tests.

The name "TimeProvider" was the best I could come up with for now but I'm open to better suggestions!

### DIFF
--- a/commons/src/main/java/org/frankframework/util/TimeProvider.java
+++ b/commons/src/main/java/org/frankframework/util/TimeProvider.java
@@ -1,0 +1,40 @@
+/*
+   Copyright 2025 WeAreFrank!
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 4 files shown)


evandongen (approved):

commons/src/main/java/org/frankframework/util/TimeProvider.java
```diff
@@ -37,4 +46,45 @@ public static Instant now() {
...
+	public static void setClock(@Nonnull Clock clock) {
+		TimeProvider.clock = clock;
+	}
+
+	/**
+	 * Get the {@link Clock} currently being used. This clock can be used as a base-clock to create other specialized clocks for testing, which is the
+	 * reason to make it accessible.
+	 */
+	public static Clock getClock() {
+		return TimeProvider.clock;
```
evandongen, line 71: Waarom is dit niet `this.clock`?

tnleeuw: Het is een static field in een static class, dus er is geen `this`

PR #9297, opened 2025-07-14
Fix containsKey in PropertyResolver

nielsm5:


core/src/test/java/org/frankframework/pipes/ReplacerPipeTest.java
```diff
@@ -208,4 +212,35 @@ public void testMarkSupportedIsFalse() throws Exception {
...
+
+			pipeLineProcessor.setPipeProcessor(inputOutputPipeProcessor);
+
+			pipeline.setPipeLineProcessor(pipeLineProcessor);
+
+			pipe.setGetInputFromFixedValue("^{prefix.?{variable}.suffix}");
+			pipe.setFind("^");
+			pipe.setReplace("$");
+			pipe.setSubstituteVars(true);
+			pipe.addParameter(ParameterBuilder.create("variable", "value"));
```
nielsm5, line 236: Added an extra test for this magnificent beast...

commons/src/test/java/org/frankframework/util/PropertyLoaderTest.java
```diff
@@ -46,6 +46,51 @@ public void testMultiCycles() {
...
+
+	@Test
+	public void testContainsKeyInApplicationProperties() {
+		assertTrue(constants.containsKey("key4"));
+	}
+
+	@Test
+	public void testContainsKeyInSystemProperties() {
+		try {
+			System.setProperty("dummy.property", "value");
```
tnleeuw, line 82: I think that this is a better proof that the `containsKey` check also includes the environment:
```suggestion
			assertFalse(constants.containsKey("dummy.property"));
			System.setProperty("dummy.property", "value");
```

PR #9335, opened 2025-07-21
Add errorhandling to server-time service

Matthbo:
Fixes Console being very slow because of timezone error when converting time/date info to server timezone

### DIFF
--- a/console/frontend/src/main/frontend/src/app/services/server-time.service.ts
+++ b/console/frontend/src/main/frontend/src/app/services/server-time.service.ts
@@ -16,12 +16,7 @@ export class ServerTimeService {
   private serverTimezoneOffset: number = 0;
 
   getIntialTime(): string {
-    try {
-      return formatDate(this.baseTime, this.dateFormat, this.locale, this.timezone);
-    } catch (error) {
-      console.error("Couldn't format initial time with a valid timezone", error);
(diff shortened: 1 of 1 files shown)


console/frontend/src/main/frontend/src/app/services/server-time.service.ts
```diff
@@ -16,7 +16,12 @@ export class ServerTimeService {
   private serverTimezoneOffset: number = 0;
 
   getIntialTime(): string {
-    return formatDate(this.baseTime, this.dateFormat, this.locale, this.timezone);
+    try {
+      return formatDate(this.baseTime, this.dateFormat, this.locale, this.timezone);
+    } catch (error) {
+      console.error("Couldn't format initial time with a valid timezone", error);
+      return formatDate(this.baseTime, this.dateFormat, this.locale);
+    }
```
tnleeuw: The invalid timezone we have now seen is `Z` which is UTC. What is `this.locale` in this context? Browser locale?

Or can we perhaps assume that when timezone cannot be parsed, we default it to UTC?

Matthbo: Locale is indeed browser locale.
I could make it be UTC as fallback, maybe also nice to put at check as soon as the timezone gets set instead of trying it everytime

nielsm5 (approved):

Issue #9212, opened 2025-06-25
Always preserve InputStream/Reader messages when created


tnleeuw:
To improve stability and ease-of-use, messages that are created with an InputStream or Reader as request-object should always be preserved to either memory or disk. This ensures that they can always be re-read without requiring checks throughout the entire code-base.

PR #9337, opened 2025-07-21
Don't close messages with session

tnleeuw:


core/src/main/java/org/frankframework/stream/Message.java
```diff
@@ -201,12 +187,18 @@ protected Message(@Nonnull InputStream request, @Nonnull MessageContext context,
...
+			copyFromTemporaryMessage(message);
 		}
-		messageNotClosedAction = new MessageNotClosedAction();
-		CleanerProvider.register(this, messageNotClosedAction);
+	}
+
+	private void copyFromTemporaryMessage(Message message) {
+		this.request = message.request;
+		message.request = null; // Prevent potential temp files being closed
+		// Copy all keys except the name, so we do not overwrite the original name (if given) with a potential temporary-file name.
```
nielsm5, line 197: When would this happen? When it's _not_ a temporary file this (indirect) constructor isn't be used? Which means it would always be a temporary message, and you should be fine copying over the entire context?

core/src/test/java/org/frankframework/http/cxf/SoapProviderTest.java
```diff
@@ -371,12 +371,8 @@ public Message processRequest(Message message, PipeLineSession pipelineSession)
 				try {
 					MatchUtils.assertXmlEquals(getFile("VrijeBerichten_PipelineRequest.xml").asString(), message.asString());
 
-					// Ensure the message is registered on the PipelineSession.
-					pipelineResult.closeOnCloseOf(pipelineSession);
```
nielsm5, line 375: This test verifies that after the pipeline has processed the message (and used to be closed) it would still be able to read the response. Now that `scheduleCloseOnSessionExit` and `closeOnCloseOf` no longer exist I sincerely doubt that this will be an issue.

I don't think we should remove the test(s), as they are still useful, but the names might be a bit confusing.

nielsm5 (commented):
This makes me so happy. A lot of complexity gone, the code is more readable and less errorprone!

core/src/main/java/org/frankframework/stream/Message.java
tnleeuw: I was thinking that this case could happen when the name comes from an attachment in an email or multipart HTTP request. If that is then overflowing to disk when we create a message, the original name (of the attachment) could be overwritten with the name of the temporary file on disk.

I did not yet see this happen, but was concerned it could.

I can add a unit test to see if I can replicate such a case with either a large HTTP request or e-mail.

core/src/test/java/org/frankframework/http/cxf/SoapProviderTest.java
tnleeuw: I'll check these tests, if I can do something to improve what they actually want to test.

tnleeuw:
> This makes me so happy. A lot of complexity gone, the code is more readable and less errorprone!

Larva tests were complaining about 1 unclosed SerializableFileReference being "leaked", both before and after these changes.
So no regressions in that either!

core/src/main/java/org/frankframework/stream/Message.java
nielsm5: Ah I might understand what you mean. Would be great to test this...

core/src/test/java/org/frankframework/http/cxf/SoapProviderTest.java
nielsm5: Perhaps for now if you could just summerize (in this pr) which tests might have had an impact on Message#close, we can always decide what to do with them later on.

core/src/main/java/org/frankframework/stream/Message.java
tnleeuw: Test added

nielsm5:
see #9212

PR #9358, opened 2025-07-23
Console use Signals for state management

Matthbo:
This PR replaces RxJS subject/observable based state management with 
Angular's Signals wherever possible in all services, components need to have their own state management analysed on a case by case bases

### DIFF
--- a/console/frontend/src/main/frontend/.eslintrc.js
+++ b/console/frontend/src/main/frontend/.eslintrc.js
@@ -26,7 +26,12 @@ module.exports = {
         'unicorn/prefer-ternary': 'warn',
         'unicorn/no-null': 'off',
         'unicorn/prefer-dom-node-text-content': 'warn',
-        'unicorn/consistent-function-scoping': 'off',
+        'unicorn/consistent-function-scoping': [
+          'error',
+          {
+            checkArrowFunctions: 'false',
+          },
+        ],
       },
     },
     {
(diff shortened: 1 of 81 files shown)


console/frontend/src/main/frontend/src/app/http-interceptors/auth-interceptor.ts
```diff
@@ -1,15 +1,11 @@
...
-import { AppService } from '../app.service';
 
 @Injectable()
 export class AuthInterceptor implements HttpInterceptor {
-  constructor(
-    private authService: AuthService,
-    private appService: AppService,
-  ) {}
+  // private authService: AuthService = inject(AuthService);
+  // private appService: AppService = inject(AppService);
```
philipsens: I think something as simple as this should not have to be kept

philipsens (changes requested):
Nice refactor! ✅ Readability

For the `// eslint-disable-next-line unicorn/consistent-function-scoping` that is showing on a lot of lines:
It is probably just because we are using named and unnamed functions inconsistently. The scope of an unnamed arrow function, however, is fine.

So I'd suggest setting this option in the ESLint config: `"checkArrowFunctions": false`

console/frontend/src/main/frontend/.eslintrc.js
```diff
@@ -26,6 +26,7 @@ module.exports = {
         'unicorn/prefer-ternary': 'warn',
         'unicorn/no-null': 'off',
         'unicorn/prefer-dom-node-text-content': 'warn',
+        'unicorn/consistent-function-scoping': 'off',
```
philipsens: I meant like this:

```suggestion
        'unicorn/consistent-function-scoping': [
          'error',
          {
            checkArrowFunctions: 'false',
          },
        ],
```

Matthbo: Oooh I didn't realise you could do it like that so I resorted to disabling the whole rule until we change to the new eslint config

PR #9365, opened 2025-07-24
Reintroduce frank-config-layout as Mermaid replacement

Matthbo:


console/frontend/src/main/frontend/package.json
```diff
@@ -86,5 +87,6 @@
     "mochawesome": "^7.1.3",
     "prettier": "^3.6.2",
     "typescript": "~5.7.3"
-  }
+  },
+  "packageManager": "pnpm@10.13.1+sha512.37ebf1a5c7a30d5fabe0c5df44ee8da4c965ca0c5af3dbab28c3a1681b70a256218d05c81c9c0dcf767ef6b8551eb5b960042b9ed4300c59242336377e01cfad"
```
philipsens, line 91: I think this also works, and then there are fewer changes to this line.

```suggestion
  "packageManager": "pnpm@10"
```

Matthbo: That would be nice, I have however noticed corepack thinks pnpm pre-releases are also able to be latest versions so I had it set to the latest stable instead.
Maybe it doesn't matter much but it's something to consider

philipsens: Don't they follow the semver spec?

Well, maybe it's fine like this. We'll see in the future.

Matthbo: They called the latest pre-release `10.14-0` or something like that, very strange

PR #9406, opened 2025-07-31
Remove a few more unneeded invocations of PipeLineSession#unscheduleCloseOnSessionExit

tnleeuw:


nielsm5 (approved):
Just to clarify for future references, we don't store the message in the PLS closeables, so there is no need to unschedule them anymore.

PR #9464, opened 2025-08-13
Split PkiUtil so it can be used in plugins

MatthijsSmets:
Moved generic PkiUtil methods to commons
Moved KeystoreType to commons

Fixed unrelated sonarqube issues because files were touched

This is in preparation for the Cloud Agent plugin which can depend only on commons and management gateway, not the core module


commons/src/test/java/org/frankframework/encryption/CommonsPkiUtilTest.java
```diff
@@ -0,0 +1,210 @@
...
+	private InputStream toInputStream(KeyStore ks) throws Exception {
+		ByteArrayOutputStream bos = new ByteArrayOutputStream();
+		ks.store(bos, DUMMY_PW.toCharArray());
+		return new ByteArrayInputStream(bos.toByteArray());
+	}
+
+	@Test
+	void testCreateKeyStoreSuccess() throws Exception {
+		URL url = mock(URL.class);
+		when(url.openStream()).thenReturn(toInputStream(keyStore));
```
nielsm5, line 97: Zeer creatief dit :D
Maar je had ook zelf een cert kunnen maken en die in de test-resources kunnen zetten :)

PR #9559, opened 2025-08-30
Cleanup credentialprovider

nielsm5:
- Rename interface to `CredentialProvider` instead of everything being a `CredentialFactory`
- Refactor `Credentials` class a bit
- Remove suppliers
- Providers should not need to know defaults, it's the job of the `CredentialFactory` to deal with this
- Use uniform `username` and `password` fields
- Move some code to their more appropriate places
- Tests are a bit silly still but at least more uniform
- Ensure all providers are traversed when calling getConfiguredAliasses()

### DIFF
--- a/credentialProvider/src/main/java/org/frankframework/credentialprovider/CredentialFactory.java
+++ b/credentialProvider/src/main/java/org/frankframework/credentialprovider/CredentialFactory.java
@@ -198,9 +198,13 @@ public class CredentialFactory {
 	public static Collection<String> getConfiguredAliases() throws Exception {
 		Collection<String> aliases = new LinkedHashSet<>();
 		for (ICredentialProvider factory : getInstance().delegates) {
-			Collection<String> configuredAliases = factory.getConfiguredAliases();
-			if (configuredAliases != null) {
-				aliases.addAll(configuredAliases);
+			try {
+				Collection<String> configuredAliases = factory.getConfiguredAliases();
+				if (configuredAliases != null) {
+					aliases.addAll(configuredAliases);
+				}
+			} catch (Exception e) {
+				log.warning("unable to find configured aliases in factory ["+factory+"]");
 			}
 		}
 		return aliases;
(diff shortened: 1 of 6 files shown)


credentialProvider/src/main/java/org/frankframework/credentialprovider/CredentialFactory.java
```diff
@@ -103,54 +110,101 @@ private void tryFactory(String factoryClassName) {
...
+			return null;
+		}
+	}
+
+	/**
+	 * Entrypoint
+	 *
+	 * Attempts to find the credential for the specified alias.
+	 *
+	 * When non is found, uses the default (provided) fallback user/pass combination.
```
evandongen, line 166:
```suggestion
	 * When none is found, uses the default (provided) fallback user/pass combination.
```

credentialProvider/src/main/java/org/frankframework/credentialprovider/CredentialFactory.java
```diff
@@ -103,54 +110,101 @@ private void tryFactory(String factoryClassName) {
...
 
-		for (ICredentialFactory factory : credentialFactoryDelegates) {
+		for (ICredentialProvider factory : credentialFactoryDelegates) {
 			try {
-				ICredentials result = factory.getCredentials(alias, defaultUsernameSupplier, defaultPasswordSupplier);
+				ICredentials result = factory.getCredentials(alias);
 
-				// check if the alias is the same as the one we are looking for - will throw if not
-				result.getUsername();
+				// Check if the alias is the same as the one we are looking for - will throw if not
```
evandongen, line 182: Ik snap dat je dit comment niet hebt geschreven, maar ik zie zo gauw niet waar dit hieronder 'checks if the alias is the same'

evandongen (approved):
Mooi spul!

tnleeuw (approved):

PR #9565, opened 2025-09-01
Cleanup and initial version to use custom alias field selectors

nielsm5:
Cleanup of the FileSystemCredentials, and add more tests for custom field selectors. 

There is a disabled test which sonar complains about. But that's still being worked on.


credentialProvider/src/main/java/org/frankframework/credentialprovider/delinea/DelineaCredentials.java
```diff
@@ -16,44 +16,28 @@
 package org.frankframework.credentialprovider.delinea;
 
 import org.frankframework.credentialprovider.CredentialAlias;
-import org.frankframework.credentialprovider.ICredentials;
 
-public class DelineaCredentials implements ICredentials {
+public class DelineaCredentials extends org.frankframework.credentialprovider.Secret {
```
evandongen, line 20: Wellicht dit ook nog even renamen naar DelineaSecret?

PR #9606, opened 2025-09-12
Make an explicit difference between global and local AppConstants values

tnleeuw:
The "global" nature of setting app-constant properties is why tests influence each other when modifying AppConstants despite clearing instances.

Now make an explicit difference between setting global and local AppConstants.


core/src/main/java/org/frankframework/configuration/OverwritePropertySourcePostProcessor.java
```diff
@@ -24,6 +24,8 @@
  * Overwrite a property available to the Ibis configuration and the Spring
  * configuration. When the property isn't present it will be added.
  *
+ * TODO: Is this class still used or can it be removed?
```
evandongen: As far as I can see, it's not

core/src/main/java/org/frankframework/util/AppConstants.java
```diff
@@ -30,12 +30,23 @@
...
+ * <p>
+ *     There is a global instance that can be retrieved with {@link AppConstants#getInstance()}, and an
+ *     instance per configuration that can be retrieved with {@link AppConstants#getInstance(ClassLoader)}
+ *     using the configuration classloader. Each method will the instance if it does not yet exist.
+ * </p>
+ * <p>
+ *     To change a value across all instances, use {@link AppConstants#setGlobalProperty(String, String)}. Changing
+ *     a value for only a single instance with {@link AppConstants#setProperty(String, String)} is in general recommended
+ *     only for tests.
+ * </p>
```
evandongen, line 49: 👌

core/src/main/java/org/frankframework/configuration/OverwritePropertySourcePostProcessor.java
tnleeuw: I also cannot find any references to it anywhere, but didn't dare to remove it just yet without Niels having a look. 
If it's unused, we can remove it in another PR.

nielsm5: Tja, I think at this point it's safe to remove...

Issue #9503, opened 2025-08-18
HttpSender throwing 'Missing required fields: ClientId, clientSecret' error while valid 'clientAlias' is provided


MLenterman:
### Current Behavior

Looks like authAlias is not being resolved. Using clientId and clientSecret like below also generates the same error and the attribute values are empty strings in "Loaded Configuration".
```xml
clientId="${kik-inzage-api/username}"
clientSecret="${kik-inzage-api/password}"
```
Hardcoding the values for clientId and clientSecret attributes also produces the exact same error, even though I can see the values being there in "Loaded Configuration".

### Expected Behavior

Ability to use an authAlias to provide client credentials

### Environment Information

**FF! 9.3.0-20250815.095041**: Kik-Inzage 1.0.3
Running on **e1b1969996ec** using **Apache Tomcat/10.1.44**
Java Version: **OpenJDK Runtime Environment (21.0.8+9-LTS)**
Heap size: **201.2 MiB**, total JVM memory: **328.0 MiB**
Free memory: **126.8 MiB**, max memory: **3.8 GiB**
Free disk space: **964.0 GiB**, total disk space: **1006.9 GiB**
Up since: **2025-08-18 14:54:14**, timezone: **Etc/UTC**

### Configuration

```xml
<SenderPipe
    name="EigendomsinformatieGetEigendomsinformatieKoi-sender"
    getInputFromFixedValue="&lt;dummy/&gt;">
    <HttpSender
        name="EigendomsinformatieGetEigendomsinformatieKoi-httpsender"
        methodType="GET"
        parametersToSkipWhenEmpty="*"
        oauthAuthenticationMethod="CLIENT_CREDENTIALS_BASIC_AUTH"
        clientAlias="kik-inzage-api"
        tokenEndpoint="${application.connections.kik-inzage-api.tokenEndpoint}"/>
    <Param name="url" pattern="${application.connections.kik-inzage-api.rootUrl}/eigendomsinformatie/kadastraalobjectidentificatie/{kadastraalobjectidentificatie}" />
    <Param name="formaat" sessionKey="formaat"/>
    <Param name="klantreferentie" sessionKey="klantreferentie"/>
    <Param name="gebruikeridentificatie" sessionKey="gebruikeridentificatie"/>
    <Param name="hyperlinkopproduct" sessionKey="hyperlinkopproduct"/>
    <Param name="inkoopnummer" sessionKey="inkoopnummer"/>
[...]

PR #9626, opened 2025-09-16
refactor: make HttpSender use shared resource properties via getters

vdegans:
This refactor updates the `AbstractHttpSession` to use getters when accessing credentials, token endpoints, and authentication methods.

By switching to getters, the `AbstractHttpSender` can now override these methods to delegate to a referenced shared resource. This ensures that when a sharedResourceRef is configured, the sender uses the live authentication properties from the shared resource instead of its own local state.

Fixes:
* #9503

### DIFF
--- a/core/src/main/java/org/frankframework/http/AbstractHttpSender.java
+++ b/core/src/main/java/org/frankframework/http/AbstractHttpSender.java
@@ -228,6 +228,8 @@ public abstract class AbstractHttpSender extends AbstractHttpSession implements 
 		if (StringUtils.isNotBlank(sharedResourceRef)) {
 			try {
 				HttpSession session = getSharedResource(sharedResourceRef);
+				log.debug("Using shared HttpSession [{}]", sharedResourceRef);
+				
 				setHttpClient(session.getHttpClient());
 				setHttpContext(session.getDefaultHttpClientContext());
 			} catch (Exception e) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/frankframework/http/AbstractHttpSender.java
```diff
@@ -474,6 +476,32 @@ public String getPhysicalDestinationName() {
 		return getUrl();
 	}
 
+	private HttpSession getSharedSession() {
+		try {
+			return StringUtils.isNotBlank(sharedResourceRef) ? getSharedResource(sharedResourceRef) : null;
```
nielsm5: In SharedHttpSenderTest:

```java
		ApplicationContext applicationContext = mock(ApplicationContext.class);
		when(applicationContext.containsBean("dummy")).thenReturn(true);
		when(applicationContext.getBean("dummy", any(SharedResource.class))).thenReturn(context);
		sender.setApplicationContext(applicationContext);
```

core/src/main/java/org/frankframework/http/AbstractHttpSender.java
```diff
@@ -474,6 +476,32 @@ public String getPhysicalDestinationName() {
...
+		} catch (NullPointerException e) {
+			return null;
+		}
+	}
+
+	@Override
+	public CredentialFactory getCredentials() {
```
nielsm5, line 506: Comments

core/src/main/java/org/frankframework/http/AbstractHttpSender.java
```diff
@@ -474,6 +476,32 @@ public String getPhysicalDestinationName() {
...
+		HttpSession sharedSession = getSharedSession();
+		return sharedSession != null ? sharedSession.getCredentials() : super.getCredentials();
+	}
+
+	@Override
+	public String getTokenEndpoint() {
```
nielsm5, line 512: When you attempt to set this on a sharedResource, throw a ConfiguraitonWarning in configure.

nielsm5 (approved):

PR #9629, opened 2025-09-16
Throw exception on move/rename when target is the same as source

evandongen:


filesystem/src/main/java/org/frankframework/filesystem/LocalFileSystem.java
```diff
@@ -311,8 +311,14 @@ public Path moveFile(Path f, String destinationFolder, boolean createFolder, Map
 				throw new FileSystemException("Cannot create folder ["+ destinationFolder +"]", e);
 			}
 		}
+
 		try {
-			Path result = Files.move(f, toFile(destinationFolder, getName(f)));
+			Path target = toFile(destinationFolder, getName(f));
+			if (exists(target)) {
+				throw new FileSystemException("Cannot move file ["+ f +"] to ["+ destinationFolder+"], source and destination are the same");
```
tnleeuw: Error message klopt niet met de conditie: source en destination hoeven niet hetzelfde te zijn. Destination bestaat al.

filesystem/src/main/java/org/frankframework/filesystem/smb/Samba1FileSystem.java
```diff
@@ -257,7 +257,12 @@ public SmbFile moveFile(SmbFile f, String destinationFolder, boolean createFolde
 		}
 
 		SmbFile dest = toFile(destinationFolder, f.getName());
+
 		try {
+			if (f.equals(dest)) {
```
tnleeuw: Moeten we niet algemeen kijken of `dest` al bestaat?

tnleeuw (approved):

nielsm5 (changes requested):
In the `FileSystemUtils.moveFile(..)` is already a check if the file exists. That should not be part of the FileSystem it self but the utility around it.

Why is that flow not working? I'd expect the `FileSystemUtils.checkSource` to take care of this.

PR #9631, opened 2025-09-16
Fix credential alias with-hyphens

nielsm5:


credentialProvider/src/main/java/org/frankframework/credentialprovider/CredentialAlias.java
```diff
@@ -98,8 +98,8 @@ private static String extractAlias(@Nullable final String rawAlias) {
 		}
 
 		String cleanAliasName = StringUtil.split(aliasName, "{").get(0);
-		if (!cleanAliasName.matches("[a-zA-Z0-9.]+")) {
-			throw new IllegalArgumentException("alias must only consist of letters and numbers");
+		if (!cleanAliasName.matches("[a-zA-Z0-9._-]+")) {
```
nielsm5, line 101: Actual issue (hyphens were not allowed).

credentialProvider/src/main/java/org/frankframework/credentialprovider/CredentialFactory.java
```diff
@@ -173,6 +174,11 @@ private static ICredentials readSecretFields(ISecret secret, CredentialAlias ali
 
 		try {
 			username = substitute(alias.getUsernameField(), secret);
+		} catch (NoSuchElementException | IOException e) {
+			log.log(Level.FINE, e, () -> "username field not found in alias ["+alias+"]");
+		}
+
+		try {
```
nielsm5, line 181: Other issue, you could not retrieve password only fields.

tnleeuw (approved):

credentialProvider/src/main/java/org/frankframework/credentialprovider/CredentialAlias.java
evandongen: Nee joh

PR #9674, opened 2025-09-23
Single Line Fixed Width Batch File Support

gvanbrakel:
I came across Fixed Width files without newlines, but I still liked to parse them ....
Unfortunately I was not able to install the full framework in Eclipse, therefor I did not run iaf-test.

### DIFF
--- a/batch/src/main/java/org/frankframework/batch/AbstractRecordHandler.java
+++ b/batch/src/main/java/org/frankframework/batch/AbstractRecordHandler.java
@@ -1,5 +1,5 @@
 /*
-   Copyright 2013, 2020 Nationale-Nederlanden, 2020-2021, 2023 WeAreFrank!
+   Copyright 2013, 2020 Nationale-Nederlanden, 2020-2021, 2023, 2025 WeAreFrank!
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
@@ -59,7 +59,7 @@ public abstract class AbstractRecordHandler implements IRecordHandler, IWithPara
 	private final List<Integer> recordIdentifyingFields = new ArrayList<>();
 
 	protected @Getter @Nonnull ParameterList paramList = new ParameterList();
-	
+
 	private @Getter int recordLength=-1;
 
 	@Override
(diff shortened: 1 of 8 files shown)


nielsm5 (approved):

tnleeuw:
Does "single line" mean that there is a single record and the file is not terminated with a newline character?

Or does it mean that the file is binary and there are multiple records, to be read entirely positionally and not terminated by a newline after each record?

tnleeuw:
PS: @gvanbrakel I don't know what issue you had with installing the framework in Eclipse; you can run it with Docker after building Docker images but also, running `mvn verify` will now run a basic version of the Larva tests in IAF-Test without needing Docker or any external software running.

tnleeuw:
> > Does "single line" mean that there is a single record and the file is not terminated with a newline character?
> > Or does it mean that the file is binary and there are multiple records, to be read entirely positionally and not terminated by a newline after each record?
> 
> The file contains multiple text records of a fixed width, that are not separated or terminated by a newline character

The name "singleLine" implies for me a single record. I'm not sure what will be a better name but I feel that "single line" isn't quite right and can be confusing.

tnleeuw:
> > PS: @gvanbrakel I don't know what issue you had with installing the framework in Eclipse; you can run it with Docker after building Docker images but also, running `mvn verify` will now run a basic version of the Larva tests in IAF-Test without needing Docker or any external software running.
> 
> Problem in Eclipse is java.lang.ClassNotFoundException: org.frankframework.util.AppConstants, when starting Tomcat, even when only iaf-example is added as a module

That's odd. Seems like some sort of dependency-management problem in Eclipse but it's the first time I hear of such a problem from any of the developers using Eclipse.
Have all maven artefacts been built & installed? Perhaps Eclipse has a problem finding some of the dependent classes that should be in the frankframework/commons module for instance? (Such as PropertyLoader, the parent-class of AppConstants).

gvanbrakel:
> > > Does "single line" mean that there is a single record and the file is not terminated with a newline character?
> > > Or does it mean that the file is binary and there are multiple records, to be read entirely positionally and not terminated by a newline after each record?
> > 
> > 
> > The file contains multiple text records of a fixed width, that are not separated or terminated by a newline character
> 
> The name "singleLine" implies for me a single record. I'm not sure what will be a better name but I feel that "single line" isn't quite right and can be confusing.

I agree on that it might be confusing. Alternatives I considered were 'unseparated' or 'noNewline'. I did not like those because of them being negative.

tnleeuw (approved):

PR #9787, opened 2025-10-17
Use expression to compute `instance.name.lc` rather than Java code

tnleeuw:


core/src/main/resources/AppConstants.properties
```diff
@@ -51,7 +51,7 @@ application.version=${frankframework-core.version}
 instance.name=Frank
 
 ## [Generated] Lowercase of ${instance.name}
-#instance.name.lc
+instance.name.lc=${=instance.name.toLowerCase()}
```
evandongen, line 54: Dit is echt zoveel beter!

evandongen (approved):

nielsm5 (approved):

PR #9805, opened 2025-10-21
Fix the CredentialProvider and Kubernetes module

nielsm5:
There were three issues.
- The CredentialProvider was minimizing the jar and shading too many classes.
Not all of them were required, but due to Spring's initialization sequence this causes stacktraces

- The kubernetes client was also minimizing the jar, leaving out essential classes. Since I'm unsure which are required I've remove the minimization option, the jar has grown 1 mb. (but not actually works!)

- The Maven Copyright plugin is extremely slow for some reason, adding over 10 minutes to my Maven builds. I've added it to the `ci` profile.

### DIFF
--- a/credentialProvider/pom.xml
+++ b/credentialProvider/pom.xml
@@ -142,6 +142,16 @@
 						</goals>
 						<phase>package</phase>
 						<configuration>
+							<filters>
+								<filter>
+									<artifact>org.springframework:spring-web</artifact>
+									<excludes>
+										<!-- Remove unused serviceloader classes to prevent ClassNotFoundExceptions. -->
+										<exclude>META-INF/web-fragment.xml</exclude>
+										<exclude>META-INF/services/jakarta.servlet.ServletContainerInitializer</exclude>
+									</excludes>
+								</filter>
+							</filters>
 							<createDependencyReducedPom>false</createDependencyReducedPom>
 							<minimizeJar>true</minimizeJar>
 							<shadedArtifactAttached>true</shadedArtifactAttached>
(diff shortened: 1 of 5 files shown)


credentialProvider/pom.xml
```diff
@@ -142,6 +142,16 @@
 						</goals>
 						<phase>package</phase>
 						<configuration>
+							<filters>
+								<filter>
+									<artifact>org.springframework:spring-web</artifact>
+									<excludes>
+										<!-- Remove unused serviceloader classes to prevent ClassNotFoundExceptions. -->
+										<exclude>META-INF/web-fragment.xml</exclude>
+										<exclude>META-INF/services/jakarta.servlet.ServletContainerInitializer</exclude>
```
nielsm5, line 151: We don't need to start Spring here so I've removed the service-loader

docker/Tomcat/webapp/src/scripts/catalinaAdditional.properties
```diff
@@ -31,7 +31,7 @@ FileViewer.permission.rules=${credentialFactory.filesystem.root:-/dev/null} * de
 org.apache.tomcat.util.digester.PROPERTY_SOURCE=org.apache.tomcat.util.digester.EnvironmentPropertySource,org.frankframework.credentialprovider.CredentialProvidingPropertySource
 
 # Configures the `Frank!Framework CredentialManager`, to load credentials from the `credentials.properties` file.
-credentialFactory.class=nl.nn.credentialprovider.PropertyFileCredentialFactory
+credentialFactory.class=org.frankframework.credentialprovider.KubernetesCredentialFactory,org.frankframework.credentialprovider.PropertyFileCredentialFactory
 credentialFactory.map.properties=/opt/frank/secrets/credentials.properties
```
nielsm5, line 35: Unsure if this is a bad choice here, but I thought why not.

kubernetes/pom.xml
```diff
@@ -14,18 +14,20 @@
...
 		</dependency>
 
+		<!-- Provided scoped dependencies -->
 		<dependency>
-			<groupId>io.fabric8</groupId>
-			<artifactId>kubernetes-client</artifactId>
-			<version>6.14.0</version>
+			<groupId>org.frankframework</groupId>
+			<artifactId>frankframework-commons</artifactId>
+			<scope>provided</scope>
```
nielsm5, line 30: This prevents it from being added to the shaded jar.

philipsens (approved):

pom.xml
```diff
@@ -1731,6 +1719,26 @@
 				</plugins>
 			</reporting>
 		</profile>
+		<profile>
+			<id>ci</id>
+			<build>
+				<plugins>
+					<!-- Niels: The Copyright plugin adds 13 minutes to the build. -->
```
tnleeuw, line 1726: Bij mij lokaal ook 😢

PR #9845, opened 2025-11-04
Add a cypress test to test keycloak endpoint from jenkins

MatthijsSmets:
This is a test to confirm keycloak can be reached from a container run by Jenkins.

### DIFF
--- a/.github/workflows/full-maven-build.yml
+++ b/.github/workflows/full-maven-build.yml
@@ -77,6 +77,14 @@ jobs:
         distribution: 'temurin'
         java-version: ${{ env.JDK_VERSION }}
 
+    - name: Create cypress.env.json from secret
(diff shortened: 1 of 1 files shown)


.github/workflows/full-maven-build.yml
```diff
@@ -77,6 +77,14 @@ jobs:
         distribution: 'temurin'
         java-version: ${{ env.JDK_VERSION }}
 
+    - name: Create cypress.env.json from secret
+      uses: mobiledevops/secret-to-file-action@v1
+      with:
+        base64-encoded-secret: ${{ secrets.CI_CYPRESS_ENV }}
+        filename: "cypress.env.json"
+        is-executable: true
```
philipsens, line 85: This does not matter @copilot

Issue #9824, opened 2025-10-27
Add lenient and strict mode to application Health endpoint


nielsm5:
### Related Problem or Business Case

There are currently 3 health endpoints, one for the application (`getFrankHealth`), a more finegrained one for a configuration (`getConfigurationHealth`) and a specific one for each adapter (`getAdapterHealth`).

When an adapter is stopped, the configuration and adapter endpoints will return an 503 error. The application endpoint is unaffected by this, however when an adapter is in state ERROR, it will also return a 503.

### Describe the Feature

It would be nice if you could add a query parameter to the application `getFrankHealth` endpoint, so when a configuration error occurs it does not return a 503 but 200 OK.

This way people can still use the endpoint to validate if the application starts up, even though the configurations have not.

### Collaboration with External Companies

No, community resolution only

### Anything else?

It looks like the endpoint has already been implemented correclty, to only look at configuration exceptions.

PR #9879, opened 2025-11-11
Add validation that table name and all configured fields are valid

### LINKED ISSUE #4184: All available database connections are used when commentField attribute on JdbcTableListener is configured with a non-existing column

**Describe the issue**
During a deployment of a new application version one of our Liquibase changes (add unique constraint to column) failed due to duplicate values in the target table causing a following Liquibase change to fail aswell, so our COMMENTS column which is configured as `commentField="COMMENTS"` never got added to the target table.

All adapters came up as green but the application was very slow, the scheduler was unable to start jobs and kept giving the following error: `(ConnectionWaitTimeoutException): (ConnectionWaitTimeoutException) J2CA1010E: Connection not available; timed out waiting for 30 seconds`.

Our DBA investigated and said that on the database side he saw 40 active sessions, 20 coming from each node we have on our acceptance environment, which is exactly the maximum size of our configured connection pool.\

After some digging around in the log files we saw the following log line:

`2022-12-14 15:28:00,592 ERROR [ListenTask[Receiver [ProcessMail]]-20621] null receivers.Receiver - Receiver [ProcessMail] error occured
nl.nn.adapterframework.core.ListenerException: [nl.nn.adapterframework.jdbc.JdbcTableListener] [ProcessMail] exception executing statement [UPDATE EXCHANGE_MAILID_TO_GUID SET STATUS='0',COMMENTS=? WHERE STATUS!='0' AND GUID=?]: (SQLSyntaxErrorException) SQLState [42000], errorCode [904]: (OracleDatabaseException) ORA-00904: "COMMENTS": ongeldige ID
~[ibis-adapterframework-core-7.7.jar:7.7]`

which made us check the presence of the COMMENTS column, once this column has been added to the target table, the amount of active sessions dropped to a normal amount of active sessions and everything was running as expected again.

In this situation I'd expect that the JdbcTableListener would check the presence of configured columns and give a configuration warning if absent before starting up.

**Reporter**
Laurens

**To Reproduce**
1. JdbcTableListener with commentField="UNKNOWN_COLUMN"
2. Start IBIS
3. See that all connections are consumed

**Environment**
```
FF! 7.7: XXXX
Running on XXXX.insim.biz using IBM WebSphere Application Server/9.0.5.13
Java Version: Java(TM) SE Runtime Environment (8.0.7.15 - pxa6480sr7fp15-20220818_01(SR7 FP15))
Heap size: 2529M, total JVM memory: 4096M
Free disk space: 11GB, total disk space: 12GB
Up since: 2022-12-13 15:55:15 (1d)
```

**Additional Environment**
 - DBMS: [e.g. Oracle, MSSQL, MariaDB, MySql, PostgreSQL] Oracle

**Additional Context**

tnleeuw:
## Changes
Add validation that table name and all configured fields are valid.

<details>
<summary>
</summary>
  
### Issues

- [x] Closes #4184

### Backports

- [x] N/A

### Tests

- [x] E2E/Integration tests added/updated (N/A)

### DIFF
--- a/core/src/main/java/org/frankframework/jdbc/JdbcTableListener.java
+++ b/core/src/main/java/org/frankframework/jdbc/JdbcTableListener.java
@@ -40,6 +40,7 @@ import org.frankframework.core.IProvidesMessageBrowsers;
 import org.frankframework.core.ITransactionalStorage;
 import org.frankframework.core.ListenerException;
 import org.frankframework.core.ProcessState;
+import org.frankframework.dbms.DbmsException;
 import org.frankframework.dbms.JdbcException;
 import org.frankframework.lifecycle.LifecycleException;
 import org.frankframework.receivers.RawMessageWrapper;
@@ -115,7 +116,7 @@ public class JdbcTableListener<M> extends JdbcListener<M> implements IProvidesMe
 		if (!isConnectionsArePooled()) {
 			try {
 				validateQueryFields(connection);
-			} catch (SQLException e) {
+			} catch (SQLException | DbmsException e) {
 				throw new LifecycleException("Error requesting database metadata", e);
 			}
 		} else {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/frankframework/jdbc/JdbcTableListener.java
```diff
@@ -95,6 +105,82 @@ public void configure() throws ConfigurationException {
 		}
 	}
 
+	@Override
+	public void start() {
+		super.start();
+		// Validate this in start not configure, because a broken configuration could otherwise prevent the startup of the entire application.
```
evandongen, line 111: Is dat niet juist de bedoeling van configure()?

tnleeuw: To prevent starting an adapter, yes.

But adding this validation to `configure()` now means the whole application becomes non-functional if the validation fails, and not just the adapter in which the misconfigured listener is placed.

That is probably a bug but something that should be addressed outside this PR so for now I've put this validation here.

nielsm5: I think that might be because this is a listener, is that correct?
If so could you add this to ticket https://github.com/frankframework/frankframework/issues/9824 ?

tnleeuw: I think if the `configure()` of a pipe throws an exception that would also have the same effect, but this is what triggered me to investigate why startup appears to break when a `configure()` fails.

I'll add this to #9824 .

PR #10137, opened 2025-12-11
Change parameter types conversion

nielsm5:
#10144 should be merged first, which adds extra tests and fixes several parameter `asValue().toString()` issues.

### DIFF
--- a/core/src/main/java/org/frankframework/parameters/Parameter.java
+++ b/core/src/main/java/org/frankframework/parameters/Parameter.java
@@ -69,7 +69,7 @@ public class Parameter extends AbstractParameter<Message> {
 		this.type = type;
 
 		if (type == ParameterType.XML) {
-			ConfigurationWarnings.add(this, log, "use attribute outputType with value [XML] instead");
+			ConfigurationWarnings.add(this, log, "use attribute xpathResult with value [XML] instead");
 			setXpathResult(OutputType.XML);
 		}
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/frankframework/parameters/Parameter.java
```diff
@@ -57,9 +67,24 @@ public String toString() {
 	/** The target data type of the parameter, related to the database or XSLT stylesheet to which the parameter is applied. */
 	public void setType(ParameterType type) {
 		this.type = type;
+
+		if (type == ParameterType.XML) {
+			ConfigurationWarnings.add(this, log, "use attribute outputType with value [XML] instead");
```
tnleeuw: Is dat niet attribute `xpathResultType`?

core/src/main/java/org/frankframework/parameters/ParameterType.java
```diff
@@ -24,62 +24,64 @@ public enum ParameterType {
 
 	/** Renders the contents of the first node (in combination with xslt or xpath). Please note that
 	 * if there are child nodes, only the contents are returned, use <code>XML</code> if the xml tags are required */
-	STRING,
+	STRING(true),
 
 	/** Renders an xml-nodeset as an xml-string (in combination with xslt or xpath). This will include the xml tags */
-	XML,
+	@ConfigurationWarning("ParameterType XML is deprecated, use StringParameter with outputType=XML instead")
```
tnleeuw: Was het niet `xpathResultType=XML`?

PR #10153, opened 2025-12-13
Create test that uses scenario.resolveProperties

nielsm5:
Setting `scenario.resolveProperties=false` to true will break this test.

Why?


test/src/test/testtool/XsltProviderListener/scenario07.properties
```diff
@@ -0,0 +1,20 @@
+scenario.description = Zaakbrug XsltProviderListenerTest 3
+
+scenario.resolveProperties=false
```
tnleeuw, line 4:
```suggestion
# Because the input-file for this scenario uses XSLT properties with same syntax as Larva scenario properties this property should be set to false and that way the scenario works as intended.
scenario.resolveProperties=false
```

tnleeuw (approved):

PR #10154, opened 2025-12-13
Unify stepParameters logic

nielsm5:
Hopefully at one point we can introduce more of a step-context type to be passed.
This at least makes both the read and write actions take the parameters for that given step.

### DIFF
--- a/larva/src/main/java/org/frankframework/larva/ScenarioRunner.java
+++ b/larva/src/main/java/org/frankframework/larva/ScenarioRunner.java
@@ -332,7 +332,7 @@ public class ScenarioRunner {
 	private int executeActionReadStep(Scenario scenario, Step step, LarvaScenarioAction scenarioAction, String fileName, Message expected) {
 		String actionName = step.getActionTarget();
 		try {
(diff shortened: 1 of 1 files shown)


larva/src/main/java/org/frankframework/larva/ScenarioRunner.java
```diff
@@ -340,7 +332,7 @@ private int executeActionWriteStep(Scenario scenario, Step step, LarvaScenarioAc
 	private int executeActionReadStep(Scenario scenario, Step step, LarvaScenarioAction scenarioAction, String fileName, Message expected) {
 		String actionName = step.getActionTarget();
 		try {
-			Message message = scenarioAction.executeRead(scenario.getProperties()); // cannot close this message because of FrankSender (JSON scenario02)
+			Message message = scenarioAction.executeRead(step.getStepParameters()); // cannot close this message because of FrankSender (JSON scenario02)
```
tnleeuw:
```suggestion
			Message message = scenarioAction.executeRead(step.getStepParameters());
```

evandongen (approved):

PR #10237, opened 2026-01-08
Add xslt test with different encoding

nielsm5:
Add test to see what happens if we brute-force an ISO-8859-1 xml file.

### DIFF
--- a/core/src/test/java/org/frankframework/xslt/XsltPipeTest.java
+++ b/core/src/test/java/org/frankframework/xslt/XsltPipeTest.java
@@ -3,6 +3,10 @@ package org.frankframework.xslt;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
 
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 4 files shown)


core/src/test/resources/Xslt/ISO-8859-1/output-encoding.xsl
```diff
@@ -0,0 +1,11 @@
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="2.0">
+	<xsl:output method="xml" indent="no" omit-xml-declaration="yes" encoding="ISO-8859-1" />
```
nielsm5, line 2: This `encoding` doesn't seem to be doing much.

core/src/test/java/org/frankframework/xslt/XsltPipeTest.java
```diff
@@ -182,4 +192,29 @@ void test3934WithDomDoc() throws Exception {
...
+		URL url = ClassLoaderUtils.getResourceURL("/Xslt/ISO-8859-1/iso-8859-1.xml");
+		Message message = new UrlMessage(url, StandardCharsets.ISO_8859_1.displayName());
+		URL resultUrl = ClassLoaderUtils.getResourceURL("/Xslt/ISO-8859-1/utf-8.xml");
+		assertNotNull(resultUrl);
+		String expected = new UrlMessage(resultUrl).asString();
+
+		// Act
+		PipeRunResult result = doPipe(message);
+		assertNotNull(result);
+		MessageUtils.computeDecodingCharset(result.getResult()); // Everything breaks if this method is not called!
```
nielsm5, line 211: So this only seems to work if we compute the (new) charset after, else the framework will interpet it as `UTF-8`.

core/src/test/resources/Xslt/ISO-8859-1/output-encoding.xsl
tnleeuw: Isn't ISO-8859-1 the default encoding for all XML?

nielsm5: No apparently it's UTF-8 🤔

evandongen (approved):

core/src/test/java/org/frankframework/xslt/XsltPipeTest.java
tnleeuw: If we need to call `computeDecodingCharset()` in the test to get the correct value, doesn't that mean we're not doing it in the place where we _should_ be doing it?

PR #10378, opened 2026-02-02
Minor cleanups and use claims from userinfo endpoint

nielsm5:
Refactored a bit of code and made it possible to 'extend' (?) or rather 'replace' the current claimset with the UserInfo claimset.
Unsure if this should be optional, or if we can say 'if you call user-info to fetch the user-info, use set user-info'.

### DIFF
--- a/security/src/main/java/org/frankframework/lifecycle/servlets/BearerOnlyAuthenticator.java
+++ b/security/src/main/java/org/frankframework/lifecycle/servlets/BearerOnlyAuthenticator.java
@@ -142,7 +142,7 @@ public class BearerOnlyAuthenticator extends AbstractServletAuthenticator {
 		String principalClaimValue = jwt.getClaimAsString(userNameAttributeName);
 		AbstractAuthenticationToken token = new JwtAuthenticationToken(jwt, authorities, principalClaimValue);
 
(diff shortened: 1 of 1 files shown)


security/src/main/java/org/frankframework/lifecycle/servlets/BearerOnlyAuthenticator.java
```diff
@@ -118,98 +114,75 @@ public SecurityFilterChain configure(HttpSecurity http) throws Exception {
...
+			jwt = updateJwtWithUserInfoUri(jwt);
 		}
 
+		Collection<GrantedAuthority> authorities = new HashSet<>(getGrantedAuthorities(jwt));
+		authorities.add(FactorGrantedAuthority.fromAuthority(FactorGrantedAuthority.BEARER_AUTHORITY));
+
+		String principalClaimValue = jwt.getClaimAsString(userNameAttributeName);
+		AbstractAuthenticationToken token = new JwtAuthenticationToken(jwt, authorities, principalClaimValue);
+
+		// If Authorities are set, ensure the user has the required roles
```
tnleeuw:
```suggestion
		// If Authorities are set, the user is authenticated if the user has at least one of the required roles
```

Issue #10426, opened 2026-02-12
Out-of-bounds exception when parsing defaults in multi-expression strings


vdegans:
### Current Behavior

Currently, when resolving a string with multiple variable expressions, an **IndexOutOfBoundsException** can occur if:

1. The first variable in the string **does not have a default value**, and  
2. A subsequent variable **does have a default value**.

#### Example:

```properties
# Properties
key1=value1

String input = "${key1}${key2:-default}";
```

### Expected Behavior

The first variable without a default is resolved to its value (or an empty string if missing), and subsequent variables with defaults are correctly resolved using their own default values, so the entire string evaluates without throwing an exception.

For the example: ${key1}${key2:-default} should result in value1default.

### Environment Information

FF 10.0.0

### Steps To Reproduce

1. Create a resources.yaml file with the following content:
```YAML
jdbc:
  - name: haalcentraal-connector
    type: "org.h2.jdbcx.JdbcDataSource"
    url: "jdbc:h2:/opt/frank/h2/haalcentraal-connector"

  - name: haalcentraal-connector-postgresql
    type: org.postgresql.xa.PGXADataSource
    url: jdbc:postgresql://${jdbc.hostname}:${jdbc.port:-5432}/${jdbc.database}?tcpKeepAlive=${jdbc.tcpKeepAlive:-true}${jdbc.adittionalUrlParameters}
    authalias: jdbc

  - name: ladybug
    type: org.postgresql.ds.PGSimpleDataSource
    url: jdbc:postgresql://${jdbc.hostname}:${jdbc.port:-5432}/${jdbc.database}?tcpKeepAlive=${jdbc.tcpKeepAlive:-true}${jdbc.adittionalUrlParameters}
    authalias: jdbc
```
2. Start the application
3. Observe that an `IndexOutOfBoundsException` is thrown, and the application will keep retrying to start.

### What database are you using?

PostgreSQL

### Relevant Log Output

```console
java.lang.StringIndexOutOfBoundsException: Range [68, 32) out of bounds for length 74
```

Issue #10425, opened 2026-02-12
Support resolving variables inside default values of other variables


vdegans:
Currently, variables contained within the default value of another variable are not resolved. The default is treated as a literal string rather than being recursively processed.

This leads to incorrect behavior when defaults themselves contain expressions.

Example of the current situation:
`${key:-${defaultKey:-value1}` would resolve to `${defaultKey:-value1}` when no properties are set.

Expected behavior
`${key:-${defaultKey:-value1}` would resolve to `value1` when no properties are set.

PR #10427, opened 2026-02-12
Prevent runtime errors by safely resolving variables with and without defaults

vdegans:
# Changes
This PR updates StringResolver to properly handle default values when resolving property expressions. Previously, expressions without defaults could trigger IndexOutOfBoundsException if a later expression in the same string included a default value.

Added logic to safely resolve variables without defaults while correctly processing later expressions that do have defaults. Nested defaults and consecutive variables are now evaluated without exceptions.

Fixed cases where missing variables previously caused the parser to incorrectly attempt to extract a default value from the wrong scope.

Example:

Before: ${jdbc.hostname}${jdbc.port:-5432} → could throw IndexOutOfBoundsException if jdbc.hostname is missing.
After: ${jdbc.hostname}${jdbc.port:-5432} → resolves to 5432 (empty string for missing hostname) without exceptions.

# Linked issues
#10426 
#10425 

# Documentation

- [ ]  FF! Doc updated (user-facing behavior/config)
- [ ]  FF! Manual updated (if applicable)
- [ ]  Javadoc updated/generated (developer-facing APIs)

# Tests

- [x]  Unit tests added/updated to cover scenarios with missing variables and defaults

# Breaking changes
Default values are now always evaluated when present.
This can change behavior if your system previously relied on unresolved defaults remaining literal or unprocessed.

### DIFF
--- a/commons/src/main/java/org/frankframework/util/StringResolver.java
+++ b/commons/src/main/java/org/frankframework/util/StringResolver.java
@@ -26,6 +26,7 @@ import java.util.Set;
 
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.logging.log4j.LogManager;
 import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
 
@@ -264,6 +265,10 @@ public class StringResolver {
 		} else {
 			if (ctx.providedDefaultValue != null) { // use default value of property if missing actual
 				String resolvedDefault = substVars(ctx.providedDefaultValue, props1, props2, ctx.resolveWithPropertyName);
+
+				org.apache.logging.log4j.Logger logger = LogManager.getLogger(StringResolver.class.getName());
+				logger.trace("No value found for key [{}], using default value [{}]", key, resolvedDefault);
+
 				sb.append(resolvedDefault);
 			}
 		}


tnleeuw (approved):

commons/src/main/java/org/frankframework/util/StringResolver.java
```diff
@@ -260,7 +263,8 @@ private static void appendReplacement(StringBuilder sb, String key, Map<?, ?> pr
 			}
 		} else {
 			if (ctx.providedDefaultValue != null) { // use default value of property if missing actual
-				sb.append(ctx.providedDefaultValue);
+				String resolvedDefault = substVars(ctx.providedDefaultValue, props1, props2, ctx.resolveWithPropertyName);
+				sb.append(resolvedDefault);
```
nielsm5, line 272: leuk, je kan dit ook direct doen, of een log regel toevoegen :)

commons/src/test/java/org/frankframework/util/StringResolverTest.java
```diff
@@ -243,6 +243,102 @@ public void resolveWithDefault() {
...
+		properties.put("b", "2");
+
+		String result = StringResolver.substVars("${a}${b}", properties);
+		assertEquals("12", result);
+	}
+
+	@Test
+	public void resolveNestedDefaultWithColonInside() {
+		String result = StringResolver.substVars("${a:-${b:-default:value}}", properties);
+		assertEquals("default:value", result);
```
nielsm5, line 310: 👍🏻

PR #10428, opened 2026-02-12
Add TIP to doc about PullingJmsListener element

nielsm5:


tnleeuw (changes requested):
I think it would be good to also add a note in RELEASES.MD where we mention the change, in the 9.0 release?

Also, the build has an error now because Frank!Doc is unhappy. I think the `{@inheritClassDoc}` should go above the `@ff.tip`.

PR #10443, opened 2026-02-16
Load json schema in configure method to detect errors early

evandongen:
So we can throw a ConfigurationException instead of a runtime exception when starting the pipe.


core/src/main/java/org/frankframework/pipes/JsonValidator.java
```diff
@@ -62,18 +62,15 @@ public class JsonValidator extends AbstractValidator {
...
-	}
 
-	@Override
-	public void start() {
 		try {
-			super.start();
 			jsonSchema = getJsonSchema();
-		} catch (IOException e) {
-			throw new LifecycleException("unable to start validator", e);
+		} catch (JsonSchemaException | IOException e) {
```
evandongen, line 72: JsonSchemaException does not extend IOException so added here as well. That's the thrown exception if the syntax is not correct.

PR #10455, opened 2026-02-18
Add keycloak compose file

Matthbo:
Adds a keycloak compose file with a realm import.
The "ff-test" realm has a client called "ff-console" which has client roles setup to work with the framework role-mapping file.

Start the docker compose setup by going to the docker directory and running `docker compose -f tomcat.yml -f keycloak.yml up`, go to `localhost:9080` and login with `frank@framework.org` and `frank` and create a new user. Now you can login to the ff console with said user.

See https://github.com/frankframework/ci-images/pull/64 also

### DIFF
--- a/docker/keycloak.yml
+++ b/docker/keycloak.yml
@@ -34,11 +34,6 @@ services:
           - keycloak.localtest.me
     ports:
       - "9080:9080"
-    post_start:
-      - command: "chown -R keycloak /opt/keycloak/data/h2"
-        user: root
-      - command: "chown -R keycloak /opt/keycloak/data/transaction-logs"
-        user: root
 
 networks:
   keycloak: {}


philipsens (commented):
We also have the https://github.com/frankframework/ci-images repo where we include a lot of config files and environment variables for different images. There are also compose files in that repo for running the images separately. 
Might it be an idea to create an image there, that automates the realm import and sets env-variables, and have it referenced here with a smaller compose file?

docker/keycloak.yml
```diff
@@ -0,0 +1,47 @@
...
+      keycloak:
+        aliases:
+          - keycloak.localtest.me
+    ports:
+      - "9080:9080"
+    post_start:
+      - command: "chown -R keycloak /opt/keycloak/data/h2"
+        user: root
+      - command: "chown -R keycloak /opt/keycloak/data/transaction-logs"
+        user: root
```
philipsens: Isn't it possible to do in the Dockerfile instead?

Matthbo: I've kept it in this compose file as the volume bindings are also in here, unless we always want these set no matter if the volume bindings are defined

philipsens: We'd only be creating and chowning the directories

Matthbo: https://github.com/frankframework/ci-images/pull/65 has been made, the post commands have been removed here

PR #10471, opened 2026-02-23
Update CRL Documentation

nielsm5:


core/src/main/java/org/frankframework/pipes/CrlPipe.java
```diff
@@ -33,27 +33,15 @@
...
- * 		<forward name="success" path="Process each Line" />
- * 	</pipe>
- * 	<pipe
- * 		name="Read issuer"
- * 		className="org.frankframework.pipes.FixedResultPipe"
- * 		filename="dir/issuer.cer"
- * 		storeResultInSessionKey="issuer">
- * 		<forward name="success" path="Read CRL" />
- * 	</pipe>
  * 	<pipe
```
tnleeuw, line 45: Why did you remove this example of how to read the data, without replacing it with a new (working) example?

nielsm5: The input needs to be a crt or crl file, the pipe `"Read input CSV file"` doesn't add much imo...
Before the input (previous pipe) mattered, not every pipe would supply an inputstream. Since v7.8 everything is a message and this no longer applies.

PR #10499, opened 2026-03-05
Add adapterAware interface to allow the Adapter from being injected

nielsm5:


core/src/main/java/org/frankframework/pipes/AbstractPipe.java
```diff
@@ -351,12 +354,13 @@ public void throwEvent(String event, Message message) {
 		}
 	}
 
+	/**
+	 * Please use {@link ConfigurationAware} or AdapterAware where needed so this can be removed from the base-class.
+	 */
 	@Override
+	@Deprecated(forRemoval = true)
 	public Adapter getAdapter() {
```
nielsm5, line 362: The goal is to get rid of this field altogether. Events should be published and traverse without the need for this.

PR #10540, opened 2026-03-16
Change EventThrowing interface

nielsm5:
Ensure each Receiver has a name.
Make it so Monitoring does not require the entire Adapter but just the AdapterName.

(I see I need to fix a few tests by ensuring Receiver has a name.)

### DIFF
--- a/core/src/main/java/org/frankframework/core/Adapter.java
+++ b/core/src/main/java/org/frankframework/core/Adapter.java
@@ -892,7 +892,7 @@ public class Adapter extends GenericApplicationContext implements ManagableLifec
 	public void addReceiver(Receiver<?> receiver) {
 		if (StringUtils.isBlank(receiver.getName())) {
 			log.warn("receiver does not have a name, generating implicit one.");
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/frankframework/core/Adapter.java
```diff
@@ -890,6 +890,11 @@ private static String getAdditionalLogging(final PipeLineSession pipeLineSession
 	public void addReceiver(Receiver<?> receiver) {
 		receivers.add(receiver);
 
+		if (StringUtils.isBlank(receiver.getName())) {
+			log.warn("receiver does not have a name, generating implicit one.");
+			receiver.setName("Receiver [%d]".formatted(receivers.size()));
```
evandongen: Als je nu meerdere receivers hebt zonder naam, krijgen ze dan niet allebei dezelfde naam?

Matthbo: This is run in the AddReceiver method, it is safe to say the size will be increased each time this has run

PR #10613, opened 2026-03-25
Update CI to split sonar from build

Matthbo:


.github/workflows/full-maven-build.yml
```diff
@@ -178,3 +178,13 @@ jobs:
         path: |
           ~/.sonar/cache
         key: sonar-cache-${{ github.base_ref }}
+
+    # Check the Quality Gate status.
+    - name: SonarQube Quality Gate check
```
nielsm5, line 183: At least this adds a quality gate...

Matthbo: Wait the name is what does it??

nielsm5: Nono, the `sonarqube-quality-gate-action` step. If you think it's ok, we can merge it? 😄

nielsm5 (approved):

.github/workflows/full-maven-build.yml
Matthbo: Oooh got it, weird they dont show this step on their website when I wanted to add a new github CI workflow. They only gave what I made at first

PR #10716, opened 2026-04-13
Introduce parameters for cache key and value, and deprecating XPath methods

### LINKED ISSUE #9873: Allow setting Cache key using a Param

The options to set the key of a cached item are very limited for an `EhCache` or `Cache` in older versions.
I would like to base the key of a combination of sessionKeys, I could do this using the stylesheet option but this feels a bit too much for a simple string concat.

### Describe the Feature

Ideally I could use a parameter to set the cache key, such as:

```
<Cache name="my-cache">
    <Parameter name="key" pattern="{Item}{User}/>
</Cache>
```

No, community resolution only

### Anything else?

_No response_

evandongen:


evandongen:
Morgen even kijken wat unit en larva testen doen.

test/src/main/configurations/MainConfig/ConfigurationCache.xml
```diff
@@ -83,35 +83,50 @@
...
+						<Param name="value" xpathExpression="*/value[text()]" xpathResult="XML"/>
+					</cache>
 					<sender className="org.frankframework.senders.EchoSender" />
 				</sender>
 				<forward name="success" path="READY"/>
 			</pipe>
-			<pipe name="XPath-storeEmtpy" >
+			<pipe name="XPath-storeEmpty" >
 				<sender className="org.frankframework.senders.SenderWrapper" >
-					<cache keyXPath="*/key" valueXPath="*/value[text()]" cacheEmptyKeys="true" cacheEmptyValues="true" />
```
nielsm5, line 114: Could you keep the old notation (in a separate file perhaps) so we keep integration-testing the now deprecated approach as well?

core/src/main/java/org/frankframework/cache/AbstractCacheAdapter.java
```diff
@@ -99,36 +140,88 @@ public String transformKey(String input, PipeLineSession session) {
...
+	 * exist or an error occurs when trying to determine the parameter value.
+	 */
+	private String getParameter(String parameterName, Message input, PipeLineSession session) {
+		if (getParameterList().hasParameter(parameterName)) {
+			try {
+				ParameterValueList parameterValueList = getParameterList().getValues(input, session);
+				ParameterValue parameterValue = parameterValueList.findParameterValue(parameterName);
+
+				if (parameterValue != null) {
+					return parameterValue.asStringValue();
```
nielsm5: I would let it return a message? In key you can use asString when when its the value you can use it as-is.

core/src/main/java/org/frankframework/cache/AbstractCacheAdapter.java
```diff
@@ -82,12 +101,34 @@ public void configure() throws ConfigurationException {
...
 	@Override
 	public String transformKey(String input, PipeLineSession session) {
-		if (StringUtils.isNotEmpty(getKeyInputSessionKey()) && session!=null) {
-			input=(String)session.get(getKeyInputSessionKey());
+		// Tries to get the parameter with name 'key', or else falls back to the deprecated way of determining the key (using keyXPath or keyStyleSheet)
+		if (getParameterList().hasParameter(PARAM_KEY)) {
+			// To behave like the original flow, this obscure logic needs to remain
+			String resolvedKey = getParameter(PARAM_KEY, new Message(input), session);
+
+			if (StringUtils.isNotBlank(resolvedKey)) {
```
nielsm5: And if this returns a message, `if( !Message.isEmpty() ) { ... `

nielsm5 (changes requested):
A few minor remarks but looks good overall :)

test/src/main/configurations/MainConfig/ConfigurationCache.xml
evandongen: Good one, I will!

evandongen:
We should not merge this into master before 10.1 release.

nielsm5 (approved):

PR #10738, opened 2026-04-24
Named secret provider to have aliases in secret

### LINKED ISSUE #10805: Add KubernetesNamedSecretProvider as an alternative Kubernetes credential provider
The existing `KubernetesCredentialFactory` resolves credentials by treating each Kubernetes secret's name as the auth alias — one secret per alias. We should also support an alternate solution where a single Kubernetes secret holds credentials for multiple aliases using a key-naming convention (like: `myalias.username`, `myalias.password`) as this is what will be used in the Frank!Cloud.
Add a new `KubernetesNamedSecretProvider` that resolves credentials from a fixed set of named Kubernetes secrets configured via `credentialFactory.kubernetes.secretNames`. Within each secret, aliases are identified by key prefix: a key like `myalias.username` means the alias is `myalias` and the field is `username`.

MatthijsSmets:
## Changes
This PR adds a new Kubernetes secret provider, which will be used primarily in the Frank!Cloud. 
Instead of an authalias corresponding to one kubernetes secret and the framework discovering secrets itself.
With this provider we set a property (`credentialFactory.kubernetes.secretNames`) to tell the framework the names of the secrets (comma separated), and each secret will have keys that are prefixed with the authalias instead.

So before:
KubernetesSecret:
```yaml
name: authAlias1
data:
	username=value
	password=value
```
And a second secret with another alias:
```yaml
name: authAlias2
data:
	username=value
	password=value
```

Now:
KubernetesSecret
```yaml
name: secretName
data:
	authAlias1.username=value
	authAlias1.password=value
	authAlias2.username=value
	authAlias2.password=value
```
And we set the property: `credentialFactory.kubernetes.secretNames` with value: `secretName`


kubernetes/src/main/java/org/frankframework/credentialprovider/KubernetesSecret.java
```diff
@@ -22,10 +22,10 @@
 import org.jspecify.annotations.Nullable;
 
 import io.fabric8.kubernetes.api.model.Secret;
-import lombok.extern.java.Log;
+import lombok.extern.log4j.Log4j2;
```
nielsm5: We use java util logging in the prod code, and only log4j2 during tests.

kubernetes/src/main/java/org/frankframework/credentialprovider/KubernetesCredentialFactory.java
```diff
@@ -71,132 +64,46 @@
...
-	static final String K8_MASTER_URL = "credentialFactory.kubernetes.masterUrl";
-	private static final String K8_NAMESPACE_PROPERTY = "credentialFactory.kubernetes.namespace";
-
-	private static final int CACHE_DURATION_MILLIS = 60_000;
-	public static final String DEFAULT_NAMESPACE = "default";
-
-	protected String namespace;
-	private KubernetesClient client;
+@Log4j2
+public class KubernetesCredentialFactory extends BaseKubernetesCredentialProvider {
```
nielsm5:
```suggestion
public class KubernetesCredentialFactory extends AbstractKubernetesCredentialProvider {
```

kubernetes/src/main/java/org/frankframework/credentialprovider/KubernetesNamedSecretProvider.java
```diff
@@ -0,0 +1,145 @@
...
+		for (String configuredSecretName : configuredSecretNames) {
+			Secret secret = secretCache.computeIfAbsentOrExpired(configuredSecretName, this::getConfiguredSecretByName);
+			if (secret == null || secret.getData() == null || secret.getData().isEmpty()) {
+				continue;
+			}
+
+			boolean hasAnyKeyForAlias = secret.getData().keySet().stream().anyMatch(k -> k.startsWith(prefix));
+			if (hasAnyKeyForAlias) {
+				return new AuthAliasPrefixKubernetesSecret(alias, secret, prefix);
+			}
```
nielsm5, line 89: Ik weet niet of dit nou handig is of dat je beter de geëxtraheerde secrets kan cachen, maar voor maar even hier mee beginnen.

nielsm5 (changes requested):

kubernetes/src/main/java/org/frankframework/credentialprovider/KubernetesSecret.java
MatthijsSmets: Good to know, updated it everywhere

nielsm5 (approved):

nielsm5:
Merge the master :)

PR #10761, opened 2026-04-29
Fix frankdoc version to avoid build failure

evandongen:


Matthbo:
The frontend fails to build do to my generate sitemap script, removing the "prebuild" script from `package.json` should fix this and make the frontend be able to build

```diff
-  "prebuild": "node ./scripts/generate-sitemap.mjs"
```

evandongen:
> The frontend fails to build do to my generate sitemap script, removing the "prebuild" script from `package.json` should fix this and make the frontend be able to build
> 
> ```diff
> -  "prebuild": "node ./scripts/generate-sitemap.mjs"
> ```

I don't think this has anything to do with the sitemap script.

Matthbo:
> > The frontend fails to build do to my generate sitemap script, removing the "prebuild" script from `package.json` should fix this and make the frontend be able to build
> > ```diff
> > -  "prebuild": "node ./scripts/generate-sitemap.mjs"
> > ```
> 
> I don't think this has anything to do with the sitemap script.

For me the parent of the doc wouldn't build because the frontend module had an error which had to do with that script, I saw the same in one of the CI builds here too hence why I mentioned it
I did think this was the doc repo however, not the framework one

nielsm5 (approved):

PR #10784, opened 2026-05-02
Remove hostname from MessageUtils.generateMessageID

### LINKED ISSUE #10756: CID generated consists of too many characters
### Current Behavior

The CID generated by this adapter exceeds the 100 characters assigned to DB field it's meant to be inserted in. This is causing issues where the adapter is no longer receptive of subsequent calls. 
CID looks like this: fallback-message-id-application-main-worker-hybrid-67dbd7b74b-7tckj-64408e34-223f90c1_19dbfee7f62_-57cc

### Expected Behavior

Expected behaviour is that the CID does not cross the 100 char threshold.

### Environment Information

**FF! 9.4.0-b20260323173333**: application-main-worker 
Running on **application-main-worker-hybrid-67dbd7b74b-7tckj** using **Apache Tomcat/11.0.15**
Java Version: **OpenJDK Runtime Environment (21.0.9+10-LTS)**
Heap size: **1013.1 MiB**, total JVM memory: **1.1 GiB**
Free memory: **156.9 MiB**, max memory: **2.0 GiB**
Free disk space: **22.6 GiB**, total disk space: **29.9 GiB**
Up since: **2026-04-28 09:49:09**, timezone: **Etc/UTC**
Console version: **9.4.0-SNAPSHOT**

### Steps To Reproduce

_No response_

### Configuration

```xml
<pipe name="Insert into DATABASE for Updating"
        className="org.frankframework.pipes.SenderPipe">
        <sender className="org.frankframework.jdbc.FixedQuerySender"
          timeout="10"
          query="INSERT INTO DATABASE VALUES(?, CURRENT_TIMESTAMP, 'Oogabooga', 0, ?)" />
        <param name="messageid" sessionKey="cid" />
        <param name="message" sessionKey="VariousArtists" />
        <forward name="success" path="nextPipe" />
      </pipe>
```

### Input

_No response_

PostgreSQL

Microsoft Edge

### Relevant Log Output

```console
org.frankframework.core.SenderException: got exception executing query [INSERT INTO DATABASE VALUES(?, CURRENT_TIMESTAMP, 'Oogabooga', 0, ?)]: (PSQLException) ERROR: value too long for type character varying(100)
```

### Anything else?

_No response_

nielsm5:
## Changes

I'm unsure why we included the hostname, but I don't think it's useful? The ID is unique enough as it is...

<details>
<summary>
</summary>
  
### Title

### Issues

### Backports

### Documentation

- [ ] FF! Doc updated (user-facing behavior/config)
- [ ] FF! Manual updated (if applicable)

### Tests

### Breaking changes

</details>

### DIFF
--- a/core/src/main/java/org/frankframework/core/Adapter.java
+++ b/core/src/main/java/org/frankframework/core/Adapter.java
@@ -83,7 +83,6 @@ import org.frankframework.stream.MessageContext;
 import org.frankframework.util.AppConstants;
 import org.frankframework.util.DateFormatUtils;
 import org.frankframework.util.LogUtil;
-import org.frankframework.util.MessageUtils;
 import org.frankframework.util.Misc;
 import org.frankframework.util.RunState;
 import org.frankframework.util.RunStateManager;
@@ -759,12 +758,7 @@ public class Adapter extends GenericApplicationContext implements ManagableLifec
 	 * @param pipeLineSession {@link PipeLineSession} session in which message is to be processed
 	 * @return The {@link PipeLineResult} from processing the message, or indicating what error occurred.
 	 */
-	public PipeLineResult processMessageDirect(String messageId, Message message, PipeLineSession pipeLineSession) {
-		if (StringUtils.isEmpty(messageId)) {
-			messageId = MessageUtils.generateFallbackMessageId();
-			log.info("messageId not set, creating synthetic id [{}]", messageId);
-			pipeLineSession.put(PipeLineSession.MESSAGE_ID_KEY, messageId);
-		}
+	public PipeLineResult processMessageDirect(@NonNull String messageId, @NonNull Message message, PipeLineSession pipeLineSession) {
 		try (final CloseableThreadContext.Instance ignored = LogUtil.getThreadContext(this, messageId, pipeLineSession);
 			IbisMaskingLayout.HideRegexContext ignored2 = IbisMaskingLayout.pushToThreadLocalReplace(composedHideRegexPattern)
 		) {
--- a/core/src/main/java/org/frankframework/senders/FrankSender.java
+++ b/core/src/main/java/org/frankframework/senders/FrankSender.java
@@ -472,17 +472,13 @@ public class FrankSender extends AbstractSenderWithParameters implements HasPhys
 		return dm;
 	}
 
+	/**
+	 * The MessageID is never null as it's set in {@link #setupChildSession(PipeLineSession, ParameterValueList, PipeLineSession)}.
+	 */
 	private ServiceClient getAdapterServiceClient(String target) throws SenderException {
 		Adapter adapter = findAdapter(target);
 		return (message, session) -> {
-			String messageId = session.getMessageId();
-			if (StringUtils.isEmpty(messageId)) {
-				// TODO figure out if this can actually happen
-				messageId = MessageUtils.generateFallbackMessageId();
-				log.info("messageId not set, creating synthetic id [{}]", messageId);
-			}
-
-			PipeLineResult plr = adapter.processMessageDirect(messageId, message, session);
+			PipeLineResult plr = adapter.processMessageDirect(session.getMessageId(), message, session);
 			session.setExitState(plr);
 			// TODO: The code here seems rather dubious actually. Check if we really want to keep this special behaviour on having the text "ListenerException" in our error message.
 			if (plr.getState() == PipeLine.ExitState.ERROR) {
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/frankframework/core/Adapter.java
```diff
@@ -759,12 +758,7 @@ public void logToMessageLogWithMessageContentsOrSize(Level level, String logMess
 	 * @param pipeLineSession {@link PipeLineSession} session in which message is to be processed
 	 * @return The {@link PipeLineResult} from processing the message, or indicating what error occurred.
 	 */
-	public PipeLineResult processMessageDirect(String messageId, Message message, PipeLineSession pipeLineSession) {
-		if (StringUtils.isEmpty(messageId)) {
```
nielsm5, line 763: This couldn't happen except when being called in the FrankSender.
I believe that class always sets the MessageId but just incase I've added the fallback there for now.

core/src/main/java/org/frankframework/senders/FrankSender.java
```diff
@@ -475,7 +475,14 @@ private DispatcherManager getDispatcherManager(boolean dllDispatch) throws Sende
 	private ServiceClient getAdapterServiceClient(String target) throws SenderException {
 		Adapter adapter = findAdapter(target);
 		return (message, session) -> {
-			PipeLineResult plr = adapter.processMessageDirect(session.getMessageId(), message, session);
+			String messageId = session.getMessageId();
+			if (StringUtils.isEmpty(messageId)) {
+				// TODO figure out if this can actually happen
+				messageId = MessageUtils.generateFallbackMessageId();
```
tnleeuw: The new messageId should also be put into the session.

core/src/main/java/org/frankframework/senders/FrankSender.java
```diff
@@ -475,7 +475,14 @@ private DispatcherManager getDispatcherManager(boolean dllDispatch) throws Sende
...
+				log.info("messageId not set, creating synthetic id [{}]", messageId);
+			}
+
+			PipeLineResult plr = adapter.processMessageDirect(messageId, message, session);
```
tnleeuw:
```suggestion
			if (StringUtils.isEmpty(session.getMessageId())) {
				// TODO figure out if this can actually happen
				session.put(PipeLineSession.MESSAGE_ID_KEY, MessageUtils.generateFallbackMessageId());
				log.info("messageId not set, creating synthetic id [{}]", session::getMessageId);
			}

			PipeLineResult plr = adapter.processMessageDirect(session.getMessageId(), message, session);
```

tnleeuw (changes requested):

core/src/main/java/org/frankframework/senders/FrankSender.java
tnleeuw: Since the same `messageId` should be in the method parameter and also in the `session`, perhaps we should remove the `messageId` parameter and instead always get it from the `session` inside the method `Adapter#processMessageWithExceptions(...)`?

And when it's not set in the session, throw an `IllegalArgumentException`.

(If we do that, probably better to do so in another PR though)

core/src/main/java/org/frankframework/core/Adapter.java
tnleeuw: Running unit tests and Larva tests, I could not trigger this code that is removed here.

I think we should keep the fallback code in case we want to backport these changes, because in older versions of the framework there might still be cases that trigger this code.

That actually also makes me hesitate about this part of the change _if_ we need to backport it to anything older than 10.0, we might need this code here in `Adapter` for safety.

core/src/main/java/org/frankframework/senders/FrankSender.java
nielsm5: I really like the idea! It seems backwards now to have this 'seemingly useless' method parameter.
I think it would also be a good idea to prevent updating `mid`, `cid`, `originalMessage` and SystemManaged-Resources (SYSTEM_MANAGED_RESOURCE_PREFIX).

Definitly something for another pr. I wanted to keep this as simple as possible for the backport.

tnleeuw: I ran tests on 9.4 and the removed code from Adapter is also not triggered in that version, so safe to remove.

In 9.0 this check is not yet present, I added it as a bit of paranoia / safety in May 2025. A safety precaution that we apparently didn't need!

So perhaps we also do not need to add it in into the FrankSender. (But if we do, then we should do it correctly by updating the session).

core/src/main/java/org/frankframework/core/Adapter.java
nielsm5: Ah, yea I've gone through the code a bit more thogoughly, and in FrankSender#setupChildSession it always generates a MessageId, making this obsolete.

tnleeuw (approved):

PR #10816, opened 2026-05-07
Run but skip steps in required maven build if no files match

philipsens:
## Changes

I want the check to always succeed, since it is required.
With this check it will always run, but just skip everything if there are no relevant files to build/test.

This will also help with the auto merging for Kodiak.

<details>
<summary>
</summary>
  
### Title

### Issues

### Backports

### Documentation

- [ ] FF! Doc updated (user-facing behavior/config)
- [ ] FF! Manual updated (if applicable)

### Tests

### Breaking changes

</details>

### DIFF
--- a/.github/workflows/full-maven-build.yml
+++ b/.github/workflows/full-maven-build.yml
@@ -19,37 +19,49 @@ on:
       - master
       - '[0-9]+.[0-9]+-release'
       - 'release/[0-9]+.[0-9]+'
-    paths-ignore:
-      - "**.md"
-      - "**.yml"
-      - "**.txt"
-      - "**.png"
-      - "**.toml"
-      - ".github/**"
-      - "!.github/workflows/*.yml"
-      - "docker/**"
   pull_request:
     branches:
       - master
       - '[0-9]+.[0-9]+-release'
       - 'release/[0-9]+.[0-9]+'
-    paths-ignore:
-      - "**.md"
-      - "**.yml"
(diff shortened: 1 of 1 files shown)


.github/workflows/full-maven-build.yml
```diff
@@ -19,44 +19,57 @@ on:
...
+              - '!**/*.png'
+              - '!**/*.toml'
+              - '!docker/**'
+              - '!.github/**'
+              - '.github/workflows/*.yml'
+
   build-maven-artifacts:
     name: Build and Test Maven Artifacts on JDK
     runs-on: ubuntu-latest
+    needs: check-changes
```
nielsm5, line 56:
```suggestion
    needs: check-changes
    if: needs.check-changes.outputs.relevant != 'true'
```

[control-jobs-with-conditions ?](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-jobs-with-conditions)

philipsens: If you don't start the job it is never going to succeed. This is one of the downsides of how GitHub actions work. At first this was all one job, but I decided to make it a separate one, for reusability and so it is easier to see that that job itself passes.

PS why the `!=`? I reversed the file list so it matched all other files instead, so the "relevant" files. This is a better way to math than to say: ignore whenever this file is touched.

evandongen (approved):

PR #10832, opened 2026-05-11
Deprecate SimpleJdbcListener in favor of JdbcTableListener

### LINKED ISSUE #10800: Deprecate the SimpleJdbcListener
This class is broken and seeing how incredibly silly it is we need to get rid of it...
Alternatives such as the JdbcTableListener should suffice, especially now that they are non-blocking.

https://github.com/frankframework/frankframework/blob/cc244ae3fad5e1b862302c58bffd427c4c75cffb/core/src/main/java/org/frankframework/jdbc/SimpleJdbcListener.java#L46

evandongen:


core/src/main/java/org/frankframework/jdbc/SimpleJdbcListener.java
```diff
@@ -43,6 +44,8 @@
  *
  * @author  Peter Leeuwenburgh
  */
+@Deprecated(forRemoval = true, since = "10.2")
+@ConfigurationWarning("Please use JdbcTableListener to perform a count query")
```
nielsm5: Perhaps you can also add something like:

```suggestion
@ConfigurationWarning("Please use JdbcTableListener (also non-locking) to perform a count query or let it directly return messages.")
```

nielsm5 (approved):

PR #10838, opened 2026-05-12
Add test for CorePipeLineProcessor to increase coverage before we merge the refactoring

tnleeuw:


core/src/test/java/org/frankframework/processors/CorePipeLineProcessorTest.java
```diff
@@ -0,0 +1,307 @@
...
+		// Assert
+		assertFalse(pipeLineResult.isSuccessful(), "Expected failure Pipe result");
+		assertEquals(expected, pipeLineResult.getResult().asString());
+		assertFalse(session.containsKey("s1"), "Did not expect pipe 1 to be executed");
+		assertFalse(session.containsKey("s2"), "Did not expect pipe 2 to be executed");
+	}
+
+	@ParameterizedTest
+	@CsvSource({
+		"fail-wrap-output, wrapping-success[fail-wrap-output], true, true, false",
```
tnleeuw, line 232: The output `wrapping-success[fail-wrap-output]` is the wrapped input. It is not the failed wrapping. The failure-forward of the output wrapper is ignored and the input is echoed.

This is not the result that I had expected originally, but this is purely a description of what the current implementation does without trying to "fix" anything in it.

I expect that the refactor-PR will change this output by following the failure-forward of the output wrapper.

core/src/test/java/org/frankframework/processors/CorePipeLineProcessorTest.java
```diff
@@ -0,0 +1,307 @@
...
+		"fail-validator-output, wrapping-success[wrapping-success[wrapping-success[fail-validator-output]]], false, true, false",
+		"fail-validator-output, wrapping-success[wrapping-success[wrapping-success[fail-validator-output]]], false, true, true",
+		"fail-wrap-output, wrapping-success[fail-wrap-output], true, false, false",
```
tnleeuw, line 235: I don't know why the output-validator failing directly to the pipeline-exit instead of an error-handling pipe, results in the pipeline message that fails the validation being wrapped an additional 2 times.

But that's just the way the current implementation works... This might change with my refactor-PR.

core/src/test/java/org/frankframework/processors/CorePipeLineProcessorTest.java
```diff
@@ -0,0 +1,307 @@
...
+		@Override
+		public @NonNull PipeRunResult doPipe(@NonNull Message message, @NonNull PipeLineSession session) throws PipeRunException {
+			try {
+				String data = message.asString();
+				if (data != null && data.contains(failOnValue)) {
+					Message result = new Message("wrapping-failed[" + data + "]");
+					return new PipeRunResult(findForward("failure"), result);
+				}
+				Message result = new Message("wrapping-success[" + data + "]");
+				return new PipeRunResult(findForward("success"), result);
```
tnleeuw, line 277: In hindsight, the testcase-output would perhaps be easier to understand if it included the name of the wrapper-pipe that executes the wrapping.

I might still push such a change, or do that in a separate PR.

core/src/test/java/org/frankframework/processors/CorePipeLineProcessorTest.java
```diff
@@ -0,0 +1,307 @@
...
+		err3.setName("err3");
+		err3.addForward(new PipeForward("success", "error"));
+		err3.setGetInputFromFixedValue("err3");
+
+		EchoPipe err4 = new EchoPipe();
+		err4.setName("err4");
+		err4.addForward(new PipeForward("success", "error"));
+		err4.setGetInputFromFixedValue("err4");
+
+		return new IPipe[]{p1, p2, err1, err2, err3, err4};
```
nielsm5, line 109: Niet dat het fout is ofzo maar waarom niet list.of ?

nielsm5 (approved):

PR #10912, opened 2026-05-26
Status page improvements & configuration tab list behaviour changes

### LINKED ISSUE #10836: Console adapter status page tweaks
"Adapter status" contains a lot more info and controls for both application, configurations and adapters, and should be called "Status" instead.

On the All tab all mentions of "configuration" should be changed to "application" and the messages (previous "configuration messages") should be collapsed, for all other configuration tabs it should always show an open "configuration messages".

Instead of having IAF_Util be hidden at all times specifically on the Status page it should listen to a new appconstant property called "application.console.status.hide" which should be interpreted as list of configurations to be hidden.

Matthbo:
Configurations will only be hidden when in the list of app constant `application.console.status.hide`
Changed "Adapter status" to "Status" & mentions of "Configuration" while on the All tab to "Application"


philipsens (approved):
I know that it was discussed to be kind of a hidden property. But I want to go to a state where everything of the frankframework is documented. I think ideally we want to document this one too. Even if it is just something short like: 
`## Hides configurations from status page. Can be useful for demo environments. Comma separated list of configuration names.`

PR #10958, opened 2026-06-03
Fix Config Layout Flow XSLT

Matthbo:


philipsens (approved):

philipsens:
It was added somewhere else, so this was double.

PR #11028, opened 2026-06-15
Fix and add tests for IbisStoreSummary

### LINKED ISSUE #11033: Ibisstore Summary broken

nielsm5:
## Changes

Fix IbisStoreSummary code
Add IbisStoreSummary tests

See #11007 for more info 😄 .

<details>
<summary>
</summary>
  
### Title

### Issues

### Backports

### Documentation

### Tests

### Breaking changes

</details>

### DIFF
--- a/core/src/test/java/org/frankframework/management/bus/endpoints/TestIbisstoreSummary.java
+++ b/core/src/test/java/org/frankframework/management/bus/endpoints/TestIbisstoreSummary.java
@@ -1,0 +1,51 @@
+package org.frankframework.management.bus.endpoints;
+
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.spy;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+
+import org.frankframework.core.Adapter;
+import org.frankframework.management.bus.message.StringMessage;
+import org.frankframework.receivers.Receiver;
+import org.frankframework.testutil.MatchUtils;
+import org.frankframework.testutil.TestConfiguration;
+import org.frankframework.testutil.TestFileUtils;
+import org.frankframework.testutil.junit.DatabaseTestEnvironment;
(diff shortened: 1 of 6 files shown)


tnleeuw (approved):

core/src/test/java/org/frankframework/management/bus/endpoints/TestIbisstoreSummary.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Unlike the other test classes in this package, this one doesn't use the message-bus.
+ * Here we test the underlying database class.
+ */
+public class TestIbisstoreSummary {
+
+	@TxManagerTest
+	@WithLiquibase
+	@WithLiquibase(file = "/Migrator/PrefillErrorStore.xml")
```
evandongen, line 29:
```suggestion
	@WithLiquibase(file = "/Migrator/PrefillErrorStore.xml")
```

evandongen (approved):

core/src/test/java/org/frankframework/management/bus/endpoints/TestIbisstoreSummary.java
nielsm5: First one runs the iaf_databasechangelog (thus creates the table)
The 2nd one fills it with records

PR #11039, opened 2026-06-16
Optionally show host and application in Ladybug table

mhdirkse:
## Changes

Please await https://github.com/wearefrank/ladybug/pull/833 before merging. Show columns host and application in the debug tab of ladybug when the Frank!Framework passes a host and an application to Ladybug. Passing these values can be done using setters in Ladybug's TestTool class. When these setters are not called, the Ladybug table will omit both the host and the application even when host and application are supplied as metadata names.

### DIFF
--- a/ladybug/debugger/src/main/resources/springIbisTestToolApi.xml
+++ b/ladybug/debugger/src/main/resources/springIbisTestToolApi.xml
@@ -52,6 +52,10 @@
 				<value>namespace</value>
 				<value>status</value>
 				<value>numberOfCheckpoints</value>
+				<!-- These will not appear in the debug table when host and application are not set
+				     on Ladybug's TestTool class -->
+				<value>host</value>
+				<value>application</value>
 			</list>
 		</property>
 	</bean>
--- a/ladybug/debugger/src/main/resources/springIbisTestToolIgnoreReport.xml
+++ b/ladybug/debugger/src/main/resources/springIbisTestToolIgnoreReport.xml
@@ -38,6 +38,8 @@
 				<value>numberOfCheckpoints</value>
 				<value>estimatedMemoryUsage</value>
 				<value>storageSize</value>
+				<value>host</value>
+				<value>application</value>
 			</list>
 		</property>
 		<property name="metadataFilter">


evandongen (commented):
Martijn, ik heb dit net nog even doorgenomen omdat dit voor mij ook nieuw terrein is. Maar, zoals ik dit nu zie, zijn deze bestanden bedoeld voor specifieke configuraties, terwijl je dit volgens mij juist generiek wilt aanpassen.

Edit: dit moet waarschijnlijk in https://github.com/wearefrank/ladybug/blob/d7b918aa196572efaff22419bf8275c4ab5f0c55/ladybug-common/src/main/java/org/wearefrank/ladybug/Config.java#L226 aangepast worden, dan kan je daar gelijk obv een environment variabele ofzo kijken of ze moeten worden getoond.

mhdirkse:
You are right. I am adding the host and the application in the metadataNames as you are suggesting. See https://github.com/wearefrank/ladybug/pull/833/changes#diff-157d19c063de65dc2a147c525c500ee336f989e166c8f5a12c7f2b209e93ac28. I intended this link to point to Config.java in my Ladybug PR but it takes you to another file than I intended.

mhdirkse:
When metadata names are not set in Spring configuration files, then the list from Config.java is taken as the default. Views that list the metadata names do not take this default. In this PR I update some of these views to include the host and the application.

evandongen:
> When metadata names are not set in Spring configuration files, then the list from Config.java is taken as the default. Views that list the metadata names do not take this default. In this PR I update some of these views to include the host and the application.

If that's the default, that's the right place to modify this, isn't it? Why don't the views that actually use the metadataNames don't use the default values? I read these xml changes as overrides for the default metadataNames, but the list contains the same values ..

mhdirkse:
The files I edited use metadataNames that differ from the defaults. Look at property `extraMetadataFieldExtractors` of bean `metadataExtractor`. The elements defined there define extra metadata names that are really used by the views. When we add metadata names then we have to repeat all existing metadata names. I have to repeat host and application because of this.

There are also views that have metadata names that are more different from the default names, for example Tibet 2. I did not add host and application to those.

Matthbo:
Is this PR still relevant?

mhdirkse:
No, it is not relevant anymore. We did the following pull requests instead: https://github.com/frankframework/frankframework/pull/11309, https://github.com/frankframework/frankframework/pull/11316, https://github.com/frankframework/frankframework/pull/11323 and https://github.com/frankframework/frankframework/pull/11495.

PR #11053, opened 2026-06-18
Add support for correlationId parameter in MessageStoreSender

### LINKED ISSUE #10895: CorrelationID* attributes missing in 9.4.2
### Current Behavior

Attribute correlationIDStyleSheet, correlationIDXPath, correlationIDSessionKey  are not exist in version 9.4. Not able to set CorrelationId value for messagelog. As shown in the picture, you can see the CorrelationId. Previously, it was possible to assign values in that section using xsl. For example, that column was used to display DocumentID_ created by the xsl file, but now it shows the FileID as the value.

Image

Image

### Expected Behavior

Image

In version 7.9, we had the correlationIDStyleSheet feature, which allowed us to set the CorrelationId value for the messagelog. I expected this feature to be available in version 9.4 as well.
It would also be beneficial if we had correlationIDXPath and correlationIDSessionKey available for messagelog, similar to what was provided in version 7.9.

Image

### Environment Information

Current version 9.4.2
Old version 7.9.7-20251126.055735

### Steps To Reproduce

_No response_

### Configuration

```xml

```

### Input

_No response_

_No response_

_No response_

### Relevant Log Output

```console

```

### Anything else?

_No response_

evandongen:
## Changes
In the past, the `MessageSendingPipe` had a lot attributes to dynamically determine the correlationId. This has been removed and while updating a legacy application, another use-case popped up. The correlationId might also be needed when using a MessageStore. 

This pull request adds support for param-based providing a correlationId, which enables you to use all the available options to determine a parameter value, inclusing using an Xpath or a stylesheet.

<details>
<summary>
</summary>
  
### Title

### Issues

### Backports

### Documentation

- [x] FF! Reference updated (user-facing behaviour/config)
- [n/a] FF! Docs updated (if applicable)

### Tests

### Breaking changes

</details>


test/src/main/configurations/MainConfig/ConfigurationMessageStoreSenderAndListener.xml
```diff
@@ -38,47 +38,92 @@
...
+				<Exit name="EXIT" state="SUCCESS" code="200"/>
+				<Exit name="EXCEPTION" state="ERROR"/>
 			</Exits>
 
-			<ExceptionPipe name="error" />
+			<ExceptionPipe name="error"/>
 		</Pipeline>
 	</Adapter>
+
+	<Adapter name="writeIbisstoreWithCidXsl">
```
evandongen, line 85: I have used these two adapters with test a pipeline and jdbc database viewer to confirm that they both work. Any ideas on how to check that correlation id resolving works in a larva test?

tnleeuw: How about the following:
1. Send a message in XML or JSON that has a field with known value
2. Extract that as CID in the MessageStoreSender
3. In the scenario itself, wait for a database record in the IBISSTORE wit that CID.

There are several scenarios that wait for records with expected values to be written in the message log / IBISSTORE.

core/src/main/java/org/frankframework/jdbc/MessageStoreSender.java
```diff
@@ -129,11 +125,26 @@ public void addParameter(IParameter p) {
...
 			messageToStore = messageWrapper;
 		}
-		return new SenderResult(storeMessage(messageId, correlationID, TimeProvider.nowAsDate(), null, null, messageToStore));
+		return new SenderResult(storeMessage(messageId, correlationId, TimeProvider.nowAsDate(), null, null, messageToStore));
+	}
+
+	/**
+	 * If the given parameterName is not in the parameterList, fall back to the default value. If it is, resolve the value
+	 */
+	private String resolveParameter(Message message, PipeLineSession session, String parameterName, String defaultValue) throws SenderException {
```
tnleeuw: Calling this method twice, you might now resolve all parameter values twice. Given the work that this does, that is rather inefficient.

tnleeuw (approved):

nielsm5 (approved):

core/src/main/java/org/frankframework/jdbc/MessageStoreSender.java
nielsm5: Ja `paramList.getValues(message, session)` kan soms best zwaar zijn... Vooral omdat er meerdere params meegegeven kunnen worden (los van mid/cid).

test/src/main/configurations/MainConfig/ConfigurationMessageStoreSenderAndListener.xml
evandongen: Thank you!

core/src/main/java/org/frankframework/jdbc/MessageStoreSender.java
evandongen: Ok, i'll make sure to do that only once

tnleeuw (approved):

core/src/main/java/org/frankframework/jdbc/MessageStoreSender.java
```diff
@@ -129,11 +129,35 @@ public void addParameter(IParameter p) {
...
+			MessageWrapper<Serializable> messageWrapper = new MessageWrapper<>(message, messageId, correlationId);
 			messageWrapper.getContext().putAll(sessionValuesToStore);
 			messageToStore = messageWrapper;
 		}
-		return new SenderResult(storeMessage(messageId, correlationID, TimeProvider.nowAsDate(), null, null, messageToStore));
+		return new SenderResult(storeMessage(messageId, correlationId, TimeProvider.nowAsDate(), null, null, messageToStore));
+	}
+
+	/**
+	 * Resolves the parameter values if relevant parameters are present, returns null otherwise.
```
tnleeuw: Please specify in the JavaDoc the names of these "relevant parameters".

core/src/main/java/org/frankframework/jdbc/MessageStoreSender.java
```diff
@@ -129,11 +129,35 @@ public void addParameter(IParameter p) {
...
+	 */
+	private @Nullable ParameterValueList resolveParameterValues(Message message, PipeLineSession session) throws SenderException {
+		if (!paramList.hasParameter(PARAM_MESSAGEID) && !paramList.hasParameter(PARAM_CORRELATION_ID)) {
+			return null;
```
tnleeuw: Instead of returning a NULL, you can return `new ParameterValueList()` and query the empty list.

core/src/main/java/org/frankframework/jdbc/MessageStoreSender.java
```diff
@@ -129,11 +129,35 @@ public void addParameter(IParameter p) {
...
+		} catch (ParameterException e) {
+			throw new SenderException("Could not resolve parameters", e);
+		}
+	}
+
+	/**
+	 * If the given parameterName is not in the parameterList, fall back to the default value. If it is, resolve the value
+	 */
+	private String resolveParameter(String parameterName, String defaultValue, @Nullable ParameterValueList parameterValues) {
+		if (parameterValues == null || !paramList.hasParameter(parameterName)) {
```
tnleeuw: If `parameterValues` is never null you can just check if the named parameter exists in `parameterValues`.

PR #11069, opened 2026-06-22
Merge logging api endpoints into one controller

Matthbo:
Small refactor to make finding these endpoints a little easier


console/backend/src/main/java/org/frankframework/console/controllers/Logging.java
MatthijsSmets, line 55: Is there a reason why all other endpoints in this controller start with server, and this one starts with logging?

Matthbo: This one is only for viewing the logs, the ones starting with server configure how the framework handles logging which for some reason is called server before I had a say in this

MatthijsSmets: That explains it, would it be good idea to move these new endpoints to their own controller then?

Matthbo: Other than the prefix for the api url, they do connect to the same parts of the framework and fall under the same BusTopic.
Im pretty sure the only reason they were in different controllers to begin with was because of a pre angular (I think also made using Echo2?) console, which was rebuilt into Angular 1 before I had to upgrade it

MatthijsSmets (approved):

nielsm5 (approved):

PR #11090, opened 2026-06-23
Add response headers from session keys

### LINKED ISSUE #10521: Add ability to set response headers in the ApiListener

I want to redirect consumers of my API to a pre-signed S3 link so I dont have to stream large files over the API gateway to the consumer.

### Describe the Feature

A way of setting the Location response header.

No, community resolution only

### Anything else?

_No response_

tnleeuw:
## Changes
Closes #10521

<details>
<summary>
</summary>
  
### Title

### Issues

### Backports

### Documentation

### Tests

### Breaking changes

</details>


Laurens-makel:
What about situations where the happy path would produce a binary with the `Content-Disposition` header set but the error paths produce a JSON message without any `Content-Disposition` header?

core/src/main/java/org/frankframework/http/rest/ApiListenerServlet.java
```diff
@@ -780,6 +780,12 @@ private String extractHeaderParamsAsXml(HttpServletRequest request, ApiListener
 	 */
 	private static boolean writeToResponseStream(ApiListener listener, HttpServletResponse response, Message result, PipeLineSession session) throws IOException {
 		response.resetBuffer();
+		for (String headerName : listener.getResponseHeaderSessionKeySet()) {
+			String headerValue = session.getString(headerName);
+			if (StringUtils.isNotBlank(headerName)) {
```
Laurens-makel: Why serialize to `String` if you are potentially going to discard the `headerValue` anyways?

tnleeuw: > Why serialize to `String` if you are potentially going to discard the `headerValue` anyways?

Good point, thanks!

Typo, I meant to check if `headerValue` would be blank. The `headerName` is assumed to be not blank at this point in code, as the string-splitting util doesn't return blank values.

tnleeuw:
> What about situations where the happy path would produce a binary with the `Content-Disposition` header set but the error paths produce a JSON message without any `Content-Disposition` header?

Nothing actually changed in that code path, you just see this in your diff because of some whitespace-changes.

In my understanding, the way this should work is that the error-path doesn't assign a value to the content-disposition session key and therefore no content-disposition header is set on the response.

core/src/main/java/org/frankframework/http/rest/ApiListener.java
```diff
@@ -462,6 +464,17 @@ public List<String> getAuthenticationRoleList() {
...
 	}
 
+	/**
+	 * An optional list of session keys to be used to set response-headers. Session key names should be separated with a {@literal ,} or {@literal ;}.
+	 */
+	public void setResponseHeaderSessionKeys(@Nullable String responseHeaderSessionKeys) {
+		this.responseHeaderSessionKeys = new HashSet<>(StringUtil.split(responseHeaderSessionKeys, ",;"));
+	}
+
+	public @NonNull Set<@NonNull String> getResponseHeaderSessionKeySet() {
```
evandongen, line 480:
```suggestion
	public @NonNull Set<@NonNull String> getResponseHeaderSessionKeys() {
```

?

evandongen (approved):

nielsm5 (approved):

core/src/main/java/org/frankframework/http/rest/ApiListener.java
tnleeuw: Volgens mij breekt dat de Frank!Doc die verwacht dat getters en setters met zelfde naam eenzelfde type hebben.

Daarom heb ik die naam niet gebruikt.

evandongen: Misschien kan je dat er nog even bijzetten dan. Dit lijkt vooral een typo

evandongen (approved):

PR #11134, opened 2026-06-30
Charset issues in XSLT processing

### LINKED ISSUE #11119: XsltPipe does not honor encoding attribute in .xslt
### Current Behavior

I want to have the result of a Xslt transformation written to a ISO-8859-1 encoded file.
To do so, I have **<xsl:output method="text" encoding="ISO-8859-1"/>** in my .xslt
Next pipe is a LocalFileSystemPipe with action=write and charset-ISO-8859-1
The result file is UTF-8 encoded.
Ladybug shows the result of the XsltPipe is a byte[], encoded with UTF-8

### Expected Behavior

The result of the XsltPipe has the encoding specified in the output-tag
or
The LocalFileSystemPipe converts the stream to the requested encoding.

I managed to work around the problem by using a to-string conversion:
<FixedResultPipe name="convert" returnString="txt" replaceFixedParams="true"> 
    <Param name="txt"/>
</FixedResultPipe>

### Environment Information

**FF! 10.2.0-20260625.042324**: MAN-OEM-Simulator 
Running on **MacBook-Pro-3.fritz.box** using **Apache Tomcat/10.1.43**
Java Version: **OpenJDK Runtime Environment (21.0.8+9-LTS)**
Heap size: **177.8 MiB**, total JVM memory: **422.0 MiB**
Free memory: **244.2 MiB**, max memory: **1.0 GiB**
Free disk space: **624.8 GiB**, total disk space: **1.8 TiB**
Up since: **2026-06-25 13:30:32**, timezone: **Europe/Amsterdam**

### Steps To Reproduce

_No response_

### Configuration

```xml

```

### Input

_No response_

_No response_

_No response_

### Relevant Log Output

```console

```

### Anything else?

_No response_

tnleeuw:
## Changes

Improve handling of non-default charsets in XSLT stylesheet processing with XML or TEXT outputs.

Part of fix for #11119 

<details>
<summary>
</summary>
  
### Title

### Issues

### Backports

### Documentation

### Tests

### Breaking changes

</details>


core/src/test/resources/Xslt/ISO-8859-1/iso-8859-1.xml
tnleeuw, line 1: File changed to allow more strict (binary) comparison between actual and expected XSLT output, to make sure that output is in the expected charset.

core/src/test/resources/Xslt/ISO-8859-1/iso-8859-1.txt
tnleeuw, line 1: Github doesn't know to render this file in ISO-8859-1 so it looks like garbage in the browser...

core/src/test/java/org/frankframework/util/MessageUtilsTest.java
```diff
@@ -210,7 +212,7 @@ public void testJsonMessage() throws IOException {
 		Message json = new Message(JSON_TEST_INPUT);
 		MimeType mimeType = MessageUtils.computeMimeType(json);
 		assertNotNull(mimeType);
-		assertEquals("application/json", mimeType.toString());
```
nielsm5, line 213: Json mag geen charset hebben, en moet altijd UTF_8 zijn.

core/src/main/java/org/frankframework/senders/XsltSender.java
```diff
@@ -215,17 +217,17 @@ protected ContentHandler createHandler(Message input, PipeLineSession session, T
...
 			}
-			if (indentXml==null) {
+			if (indentXml == null) {
 				indentXml = DEFAULT_INDENT;
 				if (log.isTraceEnabled()) log.trace("Default indentXml [{}]", indentXml);
 			}
 
 			if (handler == null || disableOutputEscaping) {
-				XmlWriter xmlWriter = messageBuilder.asXmlWriter();
+				XmlWriter xmlWriter = messageBuilder.asXmlWriter(poolToUse.getOutputEncoding());
```
nielsm5, line 230: Ik vind het heel frappant hoor. De `XmlStreamWriter.builder().setOutputStream(asOutputStream()).get();` hoort deze detactie te doen. Daar zit een hele `detectEncoding` methode in om de charset te detecteren.

core/src/test/java/org/frankframework/util/MessageUtilsTest.java
tnleeuw: Als de charset gezet is op de Message dan komt er nu blijkbaar toch altijd UTF-8 achter.
Ik zal eens kijken hoe dat komt, en wat ik daaraan kan doen.

core/src/main/java/org/frankframework/senders/XsltSender.java
tnleeuw: Dit was wel waarmee ik de output charset gefixt had... 🤷

evandongen: Maar is dit dan wel nodig?

tnleeuw: Blijkbaar wel want anders wordt de text-output altijd UTF-8 ipv de gespecificeerde charset.

core/src/main/java/org/frankframework/stream/MessageBuilder.java
```diff
@@ -56,26 +60,34 @@ public MessageBuilder() throws IOException {
...
+		charset = StandardCharsets.UTF_8;
 		return new OutputStreamWriter(outputStream, StandardCharsets.UTF_8);
 	}
 
-	public XmlWriter asXmlWriter() {
+	public @NonNull XmlWriter asXmlWriter() {
+		return asXmlWriter(null);
+	}
+
+	public @NonNull XmlWriter asXmlWriter(@Nullable Charset charset) {
```
nielsm5, line 84: This change is not used, and I feel like the `XmlStreamWriter` should determine the charset. Otherwise we might get very strange scenario's where the Xml header contains a and the actual encoding is b?

nielsm5 (approved):

core/src/main/java/org/frankframework/stream/MessageBuilder.java
tnleeuw: The change is used in the XsltSender, which needs this because apparently the XmlStreamWriter was not detecting the encoding.

The XsltSender gets the correct encoding from the XSLT config map so will always apply the correct value, but I see your concern that other code may set here an incorrect value.

I think that there is not a big problem when the output is actually XML, but there is a big problem when the output-method of the XSLT was TEXT and thus the output written to the XmlStreamWriter is text, not XML.

I'll do some more checking to see if there are other options to fix this.

tnleeuw: I've looked at the API for the XmlStreamWriter and do not see any option for it to just follow whatever encoding it is passed on.

It has an output-encoding, and will recode whatever is written to it, to that output-encoding.

If you do not set the output-encoding, it is by default UTF-8 unless you write an XML with an encoding-header. In which case that becomes the output-encoding.

For the rest it will only need to detect encoding of the data written to it, so that it can do proper recoding to the output-encoding.

I hope my explanation makes sense -- and I see no other way to make the XmlStreamWriter follow the encoding set in an XSLT that has output method `text`.

PR #11136, opened 2026-07-01
Small log improvements

nielsm5:
## Changes

<details>
<summary>
</summary>
  
### Title

### Issues

### Backports

### Documentation

### Tests

### Breaking changes

</details>

### DIFF
--- a/core/src/main/java/org/frankframework/stream/Message.java
+++ b/core/src/main/java/org/frankframework/stream/Message.java
@@ -678,10 +678,15 @@ public class Message implements Serializable {
 		toStringPrefix(result);
 		result.append(getObjectId());
 
-		if (request != null) {
-			result.append(": [").append(request).append("]");
+		if (LOG.isDebugEnabled()) {
+			if (request != null) {
+				result.append(" content [").append(request).append("]");
+			} else {
+				result.append(" no-content");
+			}
 		}
 
+		result.append(" size [").append(size()).append("]");
 
 		return result.toString();
 	}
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/frankframework/stream/Message.java
```diff
@@ -678,10 +678,15 @@ public String toString() {
 		toStringPrefix(result);
 		result.append(getObjectId());
 
-		if (request != null) {
-			result.append(": [").append(request).append("]");
+		if (LOG.isDebugEnabled()) {
```
nielsm5, line 681: @tnleeuw we don't often log messages outside of DEBUG/INFO. I think the contents may not be relevant on INFO level but the message hashcode and size do. Especially with large volumes of data.. What are your thoughts?

tnleeuw: I would agree but we need to be careful with our Larva tests that they don't break when the log-level changes.

If running the Larva tests locally with log-level DEBUG and on server with log-level INFO or WARN, they shouldn't break on local to get them working on the server.

nielsm5:
The diff is:

`<field name="TVARCHAR">exitState [ERROR], result [Message[6f8d0546:String] size [23]]</field>`

`<field name="TVARCHAR">exitState [ERROR], result [Message[HASHCODE:String]: [&lt;root forward="error"/&gt;]]</field>`

For most, this change wont do anything as locally you typically use log level DEBUG. And ideally you don't see messages like this in the error storage 😞

core/src/main/java/org/frankframework/stream/Message.java
tnleeuw: Is hashcode relevant, actually?
Type, size, object-id - check

But hashcode, what can you use it for? It's just a number and changes with every small change in data / representation.

Anyway, I'm fine leaving that in.

tnleeuw (approved):

core/src/main/java/org/frankframework/stream/Message.java
nielsm5: To correlate messages?

tnleeuw: Isn't that what message-id or object-id can be used for?

PR #11138, opened 2026-07-01
Provide a JdbcTableListener without a message store

### LINKED ISSUE #10961: Provide a `JdbcTableListener` without a message store

The JdbcTableListener maintains its own MessageLog, which cannot currently be disabled. In our situation, we already have a dedicated error-handling flow that includes its own MessageLog. Having multiple MessageLogs for the same process leads to confusion and makes it harder to determine the primary source of logging information.

### Describe the Feature

The JdbcTableListener has its own MessageLog by default. There should be an option to disable this, ideally through a configurable attribute on the listener.

No, community resolution only

### Anything else?

_No response_

Configuration 7.9.7:
```xml
<adapter name="HandleReassignedDocumentsOut" active="${go.live}">
        <receiver
            name="HandleReassignedDocumentsOut">
            <listener className="nl.nn.adapterframework.jdbc.SimpleJdbcListener"
                selectQuery="SELECT COUNT(*) FROM FILESTORE WHERE TYPE='XML' AND STATUS IN('to_zip', 'warn_scan') AND ORIGIN IN ('NN','EM')"
                 />
        </receiver>
```

9.4.x
```xml
<adapter name="HandleReassignedDocumentsOut" active="${go.live}">
        <receiver
            name="HandleReassignedDocumentsOut" onError="close">
            <JdbcTableListener
                name="HandleReassignedDocumentsOut"
                tableName="FILESTORE"
                statusField="STATUS"
                statusValueAvailable="to_zip"
                statusValueInProcess="in_process"
                statusValueProcessed="zipped"
                statusValueError="error_scan"
                keyField="FILEID"
                correlationIdField="DOCID"
                messageField="ORIGIN"
                selectCondition="TYPE='XML' AND ORIGIN IN ('NN','EM')"
            />
        </receiver>
```

evandongen:
## Changes
JdbcListener needs to be used directly. This wasn't possible with the selectQuery setter being protected.

<details>
<summary>
</summary>
  
### Title

### Issues

### Backports

- [x] Backport PRs created (if needed) and linked

### Documentation

- [n/a] FF! Reference updated (user-facing behaviour/config)
- [n/a] FF! Docs updated (if applicable)

### Tests

### Breaking changes

</details>

### DIFF
--- a/core/src/main/java/org/frankframework/jdbc/JdbcListener.java
+++ b/core/src/main/java/org/frankframework/jdbc/JdbcListener.java
@@ -450,7 +450,7 @@ public class JdbcListener<M> extends JdbcFacade implements IPeekableListener<M>,
 		return updateStatusQueries.get(state);
 	}
 
-	protected void setSelectQuery(String string) {
+	public void setSelectQuery(String string) {
 		selectQuery = string;
 	}
 
--- a/core/src/main/java/org/frankframework/jdbc/JdbcListener.java
+++ b/core/src/main/java/org/frankframework/jdbc/JdbcListener.java
@@ -63,7 +63,8 @@ import org.frankframework.util.MessageUtils;
 import org.frankframework.util.StringUtil;
 
 /**
- * JdbcListener base class.
+ * JdbcListener base class. Please note that you should not use this directly but use the JdbcTableListener or MessageStoreListener. Message browsing,
+ * and error and status updates are not performed and have to be done manually.
  *
  * @param <M> MessageWrapper or key. Key is also used as messageId
  *
--- a/core/src/main/java/org/frankframework/jdbc/AbstractJdbcListener.java
+++ b/core/src/main/java/org/frankframework/jdbc/AbstractJdbcListener.java
@@ -1,5 +1,5 @@
 /*
-   Copyright 2026 WeAreFrank!
+   Copyright 2013, 2016, 2018-2020 Nationale-Nederlanden, 2020-2025 WeAreFrank!
 
    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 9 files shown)


evandongen:
@nielsm5 , can you confirm this is the right solution, or do you see another possibility? If not, should we forward port this?

evandongen:
Made this a draft for now, since we most likely want to provide this as a custom code class

core/src/main/java/org/frankframework/jdbc/JdbcListener.java
nielsm5:
```suggestion
public abstract class AbstractJdbcListener<M> extends JdbcFacade implements IPeekableListener<M>, IHasProcessState<M>, ReceiverAware<M> {
```

tnleeuw: @nielsm5 Indeed, yesterday Erik and myself also concluded that this class should really be considered an abstract class.

Unfortunately that does leave us still with the question: what is the best listener for the requirements that NN has for this specific adapter that Sena is trying to convert from 7.x to 9.4.

nielsm5:
For the 9.4 indeed a solid solution. The 'forward-port' should have the JdbcListener extends AbstractJdbcListener or create a new class to prevent duplication there. For now, this does not break any existing JdbcListener implementations while adding/solving the problem(s) in the JdbcListener it self 🚀 .

(Though I'm unsure exactly what should extend what without diving into the code. I believe an `AbstractProcessStateJdbcListener extends JdbcListener` could also work. )

nielsm5 (approved):

tnleeuw (approved):

PR #11153, opened 2026-07-02
Document property ibistesttool.maxStorageDays in the Frank!Doc

### LINKED ISSUE #11156: Add ibistesttool.maxStorageDays property to reference documentation
https://github.com/frankframework/frankframework/blob/2ee17f82ccc5e063475b707101c89ba09fa1c048/core/src/main/resources/AppConstants.properties#L591-L593

See:
https://github.com/frankframework/frankframework/blob/2ee17f82ccc5e063475b707101c89ba09fa1c048/ladybug/debugger/src/main/java/org/frankframework/ladybug/DeploymentSpecificsBeanPostProcessor.java#L45

mhdirkse:
## Changes

Document property `ibistesttool.maxStorageDays` in the Frank!Doc. The Frank!Framework already passes this property to Ladybug. Ladybug uses it for the maximum retention time in days of Ladybug reports specifically in the database storage.

This should work because of the following code in DeploymentSpecificsBeanPostProcessor.java, method `postProcessBeforeInitialization()`:

```
			String maxStorageDays = appConstants.getProperty("ibistesttool.maxStorageDays");
			if (maxStorageDays != null) {
				long maxStorageDaysLong = OptionConverter.toFileSize(maxStorageDays, -1);
				databaseStorage.setMaxStorageDays(maxStorageDaysLong);
			}
```

### DIFF
--- a/core/src/main/resources/AppConstants.properties
+++ b/core/src/main/resources/AppConstants.properties
@@ -592,6 +592,9 @@ ibistesttool.reportTransformation=TestTool/xsl/Default.xsl
 ## Maximum database size. When necessary the oldest report(s) will be deleted when adding a new one (default is 100MB)
 #ibistesttool.maxStorageSize=100MB
 
+## Maximum retention time of Ladybug reports in the database storage
+#ibistesttool.maxStorageDays
+
 ## Maximum file size for debug storage (.tts) files (default is 1MB)
 #ibistesttool.maxFileSize=1MB
 
@@ -630,17 +633,6 @@ testtool.echo2.enabled=false
 #validators.disabled
 
 #ibistesttool.defaultView
-
-#ibistesttool.freeSpaceMinimum
-
-#ibistesttool.maxStorageSize
-
-#ibistesttool.maxBackupIndex
-
-#ibistesttool.maxFileSize
-
-## Maximum retention time of Ladybug reports in the database storage
-#ibistesttool.maxStorageDays
 
 ####
 #### ManageDatabase Internal Utility Adapter
--- a/core/src/main/resources/AppConstants.properties
+++ b/core/src/main/resources/AppConstants.properties
@@ -639,6 +639,9 @@ testtool.echo2.enabled=false
 
 #ibistesttool.maxFileSize
 
+## Maximum retention time of Ladybug reports in the database storage
+#ibistesttool.maxStorageDays
+
 ####
 #### ManageDatabase Internal Utility Adapter
 
--- a/core/src/main/resources/AppConstants.properties
+++ b/core/src/main/resources/AppConstants.properties
@@ -590,10 +590,10 @@ ibistesttool.maxMessageLength=1000000
 ibistesttool.reportTransformation=TestTool/xsl/Default.xsl
 
 ## Maximum database size. When necessary the oldest report(s) will be deleted when adding a new one (default is 100MB)
-#ibistesttool.maxStorageSize=100MB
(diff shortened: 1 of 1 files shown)


evandongen (approved):
~This might need to be backported if this also applies to the 9.4 changes you're working on.~ 

Nevermind, we don't backport documentation changes.

core/src/main/resources/AppConstants.properties
```diff
@@ -639,6 +639,9 @@ testtool.echo2.enabled=false
 
 #ibistesttool.maxFileSize
 
+## Maximum retention time of Ladybug reports in the database storage
```
jacodg: I would put the property next to ibistesttool.maxStorageSize as they are closeley related

core/src/main/resources/AppConstants.properties
jacodg: Hmm, I see this property is also on line 593 and at that location it is documented. Strictly not part of this PR but can you check for duplicates?

core/src/main/resources/AppConstants.properties
```diff
@@ -592,6 +592,9 @@ ibistesttool.reportTransformation=TestTool/xsl/Default.xsl
 ## Maximum database size. When necessary the oldest report(s) will be deleted when adding a new one (default is 100MB)
 #ibistesttool.maxStorageSize=100MB
 
+## Maximum retention time of Ladybug reports in the database storage
+#ibistesttool.maxStorageDays
```
nielsm5: Is there a default value? Or perhaps recommended value?
And do we want to enable this feature by default? Or leave it disabled?

core/src/main/resources/AppConstants.properties
nielsm5: Yea, strange.. also noticed it in the Reference doc...

core/src/main/resources/AppConstants.properties
mhdirkse: As far as I know the default value is -1. This means no limit is applied to the number of days a report is kept. We already have a default limit on the amount of data of the reports in the DatabaseStorage.

nielsm5: And what is the default for that value? And woudn't it be sensible to have both?

The pound symbol disables the property, so since both properties are 'commented out' they have no effect. This means the ladybug is in full control and end-users won't know.

After a quick search in the wearefrank/ladybug project, I found the default is indeed 100. I think a sensible default for report age, regardless of the database size is 1 month. If you agree, could you un-comment the two properties and set the default values?

(Also it might be a nice feature to be able to be able to keep failed reports longer then successful ones)

mhdirkse: I uncommented the two properties and they have defaults now.

core/src/main/resources/AppConstants.properties
```diff
@@ -590,7 +590,10 @@ ibistesttool.maxMessageLength=1000000
 ibistesttool.reportTransformation=TestTool/xsl/Default.xsl
 
 ## Maximum database size. When necessary the oldest report(s) will be deleted when adding a new one (default is 100MB)
-#ibistesttool.maxStorageSize=100MB
+ibistesttool.maxStorageSize=100MB
+
+## Maximum retention time of Ladybug reports in the database storage
+ibistesttool.maxStorageDays=31
```
evandongen: Mag 100 dagen zijn volgens Jaco

nielsm5 (approved):

PR #11157, opened 2026-07-03
Message#asInputStream(encoding) should not ignore Message charset when available

### LINKED ISSUE #11119: XsltPipe does not honor encoding attribute in .xslt

tnleeuw:
## Changes
Closes 2nd part of issue in #11119

<details>
<summary>
</summary>
  
### Title

### Issues

### Backports

### Documentation

### Tests

### Breaking changes

</details>

### DIFF
--- a/core/src/main/java/org/frankframework/stream/Message.java
+++ b/core/src/main/java/org/frankframework/stream/Message.java
@@ -46,7 +46,6 @@ import org.apache.commons.io.input.ReaderInputStream;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.jetbrains.annotations.UnknownNullability;
 import org.jspecify.annotations.NullMarked;
 import org.jspecify.annotations.Nullable;
 import org.w3c.dom.Node;
@@ -683,7 +682,7 @@ public class Message implements Serializable {
 		return result;
 	}
 
-	private @UnknownNullability String nodeToString(Node node) throws IOException {
+	private String nodeToString(Node node) throws IOException {
 		try {
 			LOG.debug("returning Node {} as String", this::getObjectId);
 			return XmlUtils.nodeToString(node);
--- a/core/src/main/java/org/frankframework/stream/Message.java
+++ b/core/src/main/java/org/frankframework/stream/Message.java
@@ -232,13 +232,11 @@ public class Message implements Serializable {
 		return (String) context.get(MessageContext.METADATA_CHARSET);
 	}
 
-	public Message withCharset(@Nullable String charset) {
+	public void setCharset(@Nullable String charset) {
 		context.withCharset(charset);
-		return this;
 	}
-	public Message withCharset(@Nullable Charset charset) {
+	public void setCharset(@Nullable Charset charset) {
 		context.withCharset(charset);
-		return this;
 	}
 
 	/**
@@ -399,6 +397,7 @@ public class Message implements Serializable {
 	 * Return a {@link Reader} backed by the data in this message. {@link Reader#markSupported()} is guaranteed to be true for the returned stream.
 	 */
 	public Reader asReader() throws IOException {
+		//noinspection java:S1874
 		return asReader(null);
 	}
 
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/frankframework/stream/Message.java
```diff
@@ -231,6 +232,15 @@ public String getCharset() {
 		return (String) context.get(MessageContext.METADATA_CHARSET);
 	}
 
+	public Message withCharset(@Nullable String charset) {
```
nielsm5: MessageContext is more like a builder pattern. I think here we can/should use `setCharset(...)`
```suggestion
	public Message setCharset(@Nullable String charset) {
```

core/src/main/java/org/frankframework/stream/Message.java
```diff
@@ -435,50 +449,81 @@ public Reader asReader(@Nullable String defaultDecodingCharset) throws IOExcepti
...
+		if (request instanceof Enum<?> enumValue) {
+			return new ByteArrayInputStream(enumValue.name().getBytes(encodingCharset));
+		}
+		if (request instanceof Node node) {
+			LOG.debug("returning Node {} as InputStream", this::getObjectId);
+			return new ByteArrayInputStream(nodeToString(node).getBytes(encodingCharset));
+		}
+
+		if (StreamUtil.AUTO_DETECT_CHARSET.equals(getCharset())) {
+			MessageUtils.computeDecodingCharset(this); // This can update the charset, or remove it if no charset could be detected
```
nielsm5:
```suggestion
			MessageUtils.computeDecodingCharset(this); // This can update the charset, or remove the `auto-detect` flag if no charset could be detected.
```

core/src/main/java/org/frankframework/stream/Message.java
```diff
@@ -435,50 +449,81 @@ public Reader asReader(@Nullable String defaultDecodingCharset) throws IOExcepti
...
+		}
+
+		if (StringUtils.isEmpty(getCharset()) || encodingCharset.equals(getCharset())) {
+			// Binary data with no charset applied, cannot recode to return as-is, OR
+			// Requested charset is same as stored charset
+			return asInputStream();
+		}
+
+		// Now we know that we have binary data with a charset that is different from the one requested, so get a Reader for the data and an input stream that recodes the data.
+		return ReaderInputStream.builder().setCharset(encodingCharset).setReader(asReader()).get();
```
nielsm5: This asReader feels a bit ... error prone. Both methods have calls to eachother.

We already use `StreamUtil.getCharsetDetectingInputStreamReader(inputStream, readerCharset);` there, isn't it better to delegate the charset to that method instead and only use `ReaderInputStream` here?

nielsm5 (changes requested):

core/src/main/java/org/frankframework/stream/Message.java
tnleeuw: The recursion will not actually happen because `asRead()` will call the `asInputStream()` without charset, but I understand your reservations.
I think that it should also work this way, and be more reliable.

evandongen (approved):

nielsm5 (approved):

PR #11176, opened 2026-07-06
Fix flaky test - ensure soapwrapper byte size is correct

nielsm5:


evandongen (approved):

evandongen:
Oh, toch niet: 
`Error:    SoapWrapperTest.validateEncryptedErrorSoap1_1:417 expected: <Given final block not properly padded. Such issues can arise if a bad key is used during decryption.> but was: <Error parsing the inputstream>`

tnleeuw:
It seems that the test still works only sometimes. The flakiness hasn't been fixed, apparently.

nielsm5:
I have merge conflicts because of the master merge...

PR #11187, opened 2026-07-07
Update eslint config & fix linting errors

Matthbo:
Fixes linting in Console frontend
Also solves all linting errors and most warnings


philipsens (approved):
Most things really improved! Especially the early returns are nice.

What was the reason for moving the unicorn config to the top?

console/frontend/src/main/frontend/src/main.ts
```diff
@@ -39,30 +38,28 @@ function onReady(): void {
...
-        bodyElement.classList.add('body-small');
-      } else {
-        bodyElement.classList.remove('body-small');
-      }
+      bodyElement.classList.toggle('body-small', bodyElement.clientWidth < 769);
     });
   }
   bodyElement.addEventListener('scroll', function (this: HTMLElement) {
+    // eslint-disable-next-line unicorn/no-this-outside-of-class
     const scroll2top = this.querySelector<HTMLElement>('.scroll-to-top');
```
philipsens, line 57: I'm wondering what it wanted to do here.

Matthbo:
> Most things really improved! Especially the early returns are nice.
> 
> What was the reason for moving the unicorn config to the top?

Now all the file type specific rules are at the bottom and the global rules are at the top, looks cleaner to me than the other way around :smile:

console/frontend/src/main/frontend/src/main.ts
Matthbo: No clue, they probably didnt think old js style code still used this as a reference for the elements

philipsens:
> > Most things really improved! Especially the early returns are nice.
> > What was the reason for moving the unicorn config to the top?
> 
> Now all the file type specific rules are at the bottom and the global rules are at the top, looks cleaner to me than the other way around 😄

Ah yes, that's why I put it at the bottom. I liked how some rules could specifically be set for certain file types, but unicorn was not configurable so I placed it at the bottom. I can't remember if the ordering was important for any other reason.

Does unicorn target other files too?

Matthbo:
> > > Most things really improved! Especially the early returns are nice.
> > > What was the reason for moving the unicorn config to the top?
> > 
> > 
> > Now all the file type specific rules are at the bottom and the global rules are at the top, looks cleaner to me than the other way around 😄
> 
> Ah yes, that's why I put it at the bottom. I liked how some rules could specifically be set for certain file types, but unicorn was not configurable so I placed it at the bottom. I can't remember if the ordering was important for any other reason.
> 
> Does unicorn target other files too?

The whole eslint config should now target everything in `./src`, but practically if I run `pnpm lint` it targets .ts, .scss & .html files

Matthbo:
Oh also some unicorn rules didnt work and even broke eslint for .html files so I had to turn them off globally and enable them for .ts files only, turning them off for .html files only didn't work

PR #11261, opened 2026-07-14
Add combobox cypress test

MatthijsSmets:


console/frontend/src/main/frontend/cypress/e2e/general/combobox.cy.ts
```diff
@@ -0,0 +1,84 @@
+describe('Combobox', () => {
+  const combobox = '[data-cy=test-pipeline__config__input]';
+  const comboboxInput = `${combobox} input`;
+  const options = '.combobox__options';
+  const optionLabels = '.combobox__list-item__label';
+  const selectedOption = '.combobox__list-item--selected';
+  const clearButton = '.combobox__list-item--clear';
+
+  beforeEach(() => {
+    cy.intercept('GET', '**/api/adapters*').as('getAdapters');
```
philipsens, line 10: There are two requests to /adapters. One is extended and one isn't.

I'm not sure if the correct one is used.

console/frontend/src/main/frontend/cypress/e2e/general/combobox.cy.ts
```diff
@@ -0,0 +1,84 @@
...
+    cy.visit('/#/test-pipeline');
+    cy.wait('@getAdapters', { timeout: 30_000 });
```
philipsens, line 12: Why wait so long?

console/frontend/src/main/frontend/cypress/e2e/general/combobox.cy.ts
```diff
@@ -0,0 +1,84 @@
...
+    cy.get(comboboxInput).should('not.be.disabled');
```
philipsens, line 13: This should've waited for `configurationsLoaded`, that kind of worked like an intercept but more natively in the code.

philipsens:
The CI on Jenkins is skipped, keep that in mind.

console/frontend/src/main/frontend/cypress/e2e/general/combobox.cy.ts
MatthijsSmets: It does not actually wait that long, it. just gives the wait more time in case it takes longer than the default 20 seconds.

console/frontend/src/main/frontend/cypress/e2e/general/combobox.cy.ts
MatthijsSmets: Its fine to leave this here for now

PR #11269, opened 2026-07-14
Fix exception traces in STDOUT

### LINKED ISSUE #11262: The `VerifyDatabaseConnectionBean` does not log the cause of a startup problem

When there's a problem starting up the application, a FATAL log entry is logged with an accompanying stacktrace, resulting in the following log lines:

```
2026-07-14 11:03:27,700 FATAL [main] {} lifecycle.VerifyDatabaseConnectionBean - NOTE: when using JMS the framework requires a database to be present, else you can influence this by setting property [jdbc.required=false]
2026-07-14 11:05:44,322 WARN  [main] {} support.ClassPathXmlApplicationContext - Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'VerifyDatabaseConnectionBean' defined in class path resource [SpringApplicationContext.xml]: no default datasource [jdbc/frankapplication] found or configured
```

In my case, this gives you the completely wrong direction, since this has nothing to do with JMS and just the H2 dependency is missing.

### Describe the Feature

It seems we correctly throw the IllegalStateException with the Exception causing the issue:

Image

I'd like:
* It would be nice if the cause is logged
* The line about JMS is not always true, can this be changed?

No, community resolution only

### Anything else?

_No response_

nielsm5:
## Changes

image

Do we also want to reverse the order of traces in our normal logging?

<details>
<summary>
</summary>
  
### Title

### Issues

### Backports

### Documentation

### Tests

### Breaking changes

</details>

### DIFF
--- a/core/src/main/java/org/frankframework/lifecycle/VerifyDatabaseConnectionBean.java
+++ b/core/src/main/java/org/frankframework/lifecycle/VerifyDatabaseConnectionBean.java
@@ -119,7 +119,9 @@ public class VerifyDatabaseConnectionBean implements ApplicationContextAware, In
 			log.info("found default datasource [{}]", dataSource);
 			return dataSource;
 		} catch (Exception e) {
-			log.fatal("NOTE: when using JMS the framework requires a database to be present, else you can influence this by setting property [jdbc.required=false]");
+			// Some exceptions may occur when JMS is enabled but no DataSource has been provided.
+			// To 'always' give people a hint, this informative note is logged as warning.
+			log.warn("NOTE: when using JMS the framework requires a database to be present, else you can influence this by setting property [jdbc.required=false]");
 			throw new IllegalStateException("no default datasource ["+defaultDatasource+"] found or configured", e);
 		}
 	}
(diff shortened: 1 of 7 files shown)


mhdirkse:
This is nice when you are not really interested in the exception. This is the case when I see exceptions about the credentials factory because I usually do not work on authorization. How about exceptions you want to investigate? Do you have to go to the logfiles to see them?

nielsm5:
> This is nice when you are not really interested in the exception. This is the case when I see exceptions about the credentials factory because I usually do not work on authorization. How about exceptions you want to investigate? Do you have to go to the logfiles to see them?

Yes, this is for the system out, not the log files themselves.

core/src/main/java/org/frankframework/lifecycle/VerifyDatabaseConnectionBean.java
```diff
@@ -120,7 +120,7 @@ private void checkTxManagerAndDataSource(PlatformTransactionManager transactionM
 			return dataSource;
 		} catch (Exception e) {
 			log.fatal("NOTE: when using JMS the framework requires a database to be present, else you can influence this by setting property [jdbc.required=false]");
```
evandongen: deze log.fatal is niet altijd waar

evandongen (approved):

tnleeuw (approved):

PR #11309, opened 2026-07-17
Fix circular bean creation with Ladybug

mhdirkse:
## Changes

Fix starting the Frank!Framework after bumping Ladybug. Ladybug's View class now needs a reference to class TestTool while class TestTool also needs references to all View objects. Ladybug handles this circular dependency as follows. Class View has a setter for the TestTool object but this property is not autowired. Instead class TestTool iterates over all the View objects and initializes their references to TestTool.

This solution was mitigated by the Spring configuration files for Ladybug in the Frank!Framework. In file `ladybug/common/src/main/resources/springTestToolCommon.xml` for example, the View beans had attribute `autowire="byName"`. This attribute forced autowiring even though no @Autowire annotations were present in Ladybug, re-introducing the circular dependency. This PR fixes the issue.

### DIFF
--- a/ladybug/common/src/main/resources/springTestToolCommon.xml
+++ b/ladybug/common/src/main/resources/springTestToolCommon.xml
@@ -10,11 +10,11 @@
 		http://www.springframework.org/schema/util    http://www.springframework.org/schema/util/spring-util.xsd
 	">
 
-	<bean name="whiteBoxView" class="org.wearefrank.ladybug.filter.View" autowire="byName">
+	<bean name="whiteBoxView" class="org.wearefrank.ladybug.filter.View">
 		<property name="name" value="White box"/>
 	</bean>
 
-	<bean name="grayBoxView" parent="whiteBoxView" autowire="byName">
+	<bean name="grayBoxView" parent="whiteBoxView">
 		<property name="name" value="Gray box"/>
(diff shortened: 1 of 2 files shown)


mhdirkse:
This is the relevant test in the Ladybug CI/CD: https://github.com/wearefrank/ladybug/actions/runs/29572505967/job/87859338917.

mhdirkse:
The ladybug tests I mentioned in the previous comment do not all succeed. There are less failures than without this fix. The tests that are still failing seem to be caused by a different issue.

mhdirkse:
I looked a bit better at mentioned test and I now think it is about the same issue. I made Ladybug PR https://github.com/wearefrank/ladybug/pull/873. The relevant Ladybug test run is now https://github.com/wearefrank/ladybug/actions/runs/29577565179.

mhdirkse:
Updated the Ladybug PR. Next try of Ladybug test: https://github.com/wearefrank/ladybug/actions/runs/29578731954.

mhdirkse:
Tests succeed. The Ladybug PR introduced here only changes Ladybug's test code. The Ladybug PR and this PR can be merged independently of each other.

jacodg (approved):

nielsm5 (approved):

PR #11315, opened 2026-07-22
Lookup of parameter values is now always case-insensitive but case-preserving

### LINKED ISSUE #11263: Case-insensitive parameter names
In some pipes, the names of parameters are case insensitive while in other pipes, the case used for spelling parameter names does matter.

We should make this consistent, preferably consistently apply case insensitive lookup for parameter names.

NB: We could achieve this by making the map of param. values a TreeMap using `new TreeMap<>(String.CASE_INSENSITIVE_ORDER)`; that facilitaties case-insensitive lookup. We do lose order in which parameters were defined in the map, but for that we still have the list.
(We now create a LinkedHashMap which does preserve order of param creation when iterating over entries).

tnleeuw:
## Changes
Closes #11263

- Lookup of parameter values in the ParameterValueList / Map is now always case-insensitive
- This means that in Configuration files you do not need to use the exact same case of parameter names as in code
- Some parameter lookup methods have been moved from class AbstractPipe to ParameterValueList to make them more generic
- Some static methods have been made into instance methods since the instance is always available now

<details>
<summary>
</summary>
  
### Title

### Issues

### Backports

### Documentation

### Tests

### Breaking changes

</details>

### DIFF
--- a/core/src/main/java/org/frankframework/cache/AbstractCacheAdapter.java
+++ b/core/src/main/java/org/frankframework/cache/AbstractCacheAdapter.java
@@ -213,7 +213,7 @@ public abstract class AbstractCacheAdapter<V> implements ICache<String, V>, Fran
 		if (getParameterList().hasParameter(parameterName)) {
 			try {
 				ParameterValueList parameterValueList = getParameterList().getValues(input, session);
-				ParameterValue parameterValue = parameterValueList.findParameterValue(parameterName);
+				ParameterValue parameterValue = parameterValueList.get(parameterName);
 
 				if (parameterValue != null) {
 					return parameterValue.asMessage();
--- a/core/src/main/java/org/frankframework/parameters/ParameterValueList.java
+++ b/core/src/main/java/org/frankframework/parameters/ParameterValueList.java
@@ -18,9 +18,9 @@ package org.frankframework.parameters;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
 import java.util.stream.Stream;
 
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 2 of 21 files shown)


core/src/main/java/org/frankframework/parameters/ParameterValueList.java
```diff
@@ -44,7 +44,7 @@ public class ParameterValueList implements Iterable<ParameterValue> {
 	public ParameterValueList() {
 		super();
 		list = new ArrayList<>();
-		map  = new LinkedHashMap<>();
+		map  = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```
tnleeuw, line 47: Many changes in the PR but this is the change that matters

core/src/main/java/org/frankframework/cache/AbstractCacheAdapter.java
```diff
@@ -213,7 +213,7 @@ private Message getParameter(String parameterName, Message input, PipeLineSessio
 		if (getParameterList().hasParameter(parameterName)) {
 			try {
 				ParameterValueList parameterValueList = getParameterList().getValues(input, session);
-				ParameterValue parameterValue = parameterValueList.findParameterValue(parameterName);
+				ParameterValue parameterValue = parameterValueList.get(parameterName);
```
tnleeuw, line 216: `ParameterValueList#get` is now case-insensitive so the `find` method is no longer needed.

core/src/main/java/org/frankframework/senders/FrankSender.java
```diff
@@ -545,15 +545,17 @@ protected Adapter findAdapter(String target) throws SenderException {
 	}
 
 	Scope determineActualScope(@Nullable ParameterValueList pvl) {
-		ParameterValue scopeParam = pvl != null ? pvl.findParameterValue(SCOPE_PARAM_NAME) : null;
+		ParameterValue scopeParam;
+		scopeParam = pvl != null ? pvl.get(SCOPE_PARAM_NAME) : null;
```
nielsm5:
```suggestion
		ParameterValue scopeParam = pvl != null ? pvl.get(SCOPE_PARAM_NAME) : null;
```

nielsm5 (approved):

core/src/main/java/org/frankframework/parameters/ParameterValueList.java
evandongen: Always good to point this out!

PR #11316, opened 2026-07-23
Bump ladybug and apply testTool.setApplication()

mhdirkse:
## Changes

Use the newest version of Ladybug. This version of Ladybug adds an extra column to the debug tab named Application. In combination with the Frank!Framework, the value to be shown there is `${instance.name}`. This PR passes this value to Ladybug, allowing Ladybug to show this value.

This Ladybug version also supports filter parameters in the URL used to open it. When you add a filter parameter like `filter-application=something`, then Ladybug presets a filter in the filter drawer for reports having column "Application" containing value "something". The user can clear this filter in the side drawer if he likes.

The present PR does not apply this feature of ladybug. A future PR is needed on the Frank!Framework to call Ladybug with the desired filter parameters.

### DIFF
--- a/ladybug/debugger/src/main/java/org/frankframework/ladybug/LadybugDebugger.java
+++ b/ladybug/debugger/src/main/java/org/frankframework/ladybug/LadybugDebugger.java
@@ -21,6 +21,9 @@ import java.util.List;
 import java.util.Set;
 
 import org.apache.logging.log4j.Logger;
+
+import org.frankframework.util.AppConstants;
+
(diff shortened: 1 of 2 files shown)


mhdirkse:
image

I tested this. The shown picture shows two reports that were produced with an old version of Ladybug. These reports do not have an application name. A new report was created with the newest version of Ladybug. That report does have an application and this value is shown.

nielsm5 (approved):

PR #11369, opened 2026-08-07
Ensure receiver always has a correct name

### LINKED ISSUE #11373: Receiver name uniqueness
### Discussed in https://github.com/frankframework/frankframework/discussions/10553

<div type='discussions-op-text'>

<sup>Originally posted by **nielsm5** March 18, 2026</sup>
Receiver names have never been unique. For some reason nobody seems to have been bothered by this thus far but we're running into a few problems:

- Starting/Stopping receivers in the FF! Console doesn't work, it will always 'manage' the first one.
- Message-stores, such as the `ErrorStorage` cannot be manged properly (moving/deleting/resending messages).
- Monitoring receives triggers (events) on both receivers.
- Logging is unclear, you can't differentiate the data-entrypoint (easily).
- Statistics are grouped, even though there could be multiple receivers.

On top of this, adding a name to a listener is optional. In order to deal with this we've always added 'receiver of [<listener>]' names, or 'listener of [<receiver>]' if either of the two had a listener. And if neither has a listener, the receiver's name will be the classname of the listener. This results in strange names such as 'Receiver of <WebServiceListener>'.

I'm wondering what the added benefit is of the Receiver name and a Listener name.
We now display both in the console in the `Adapter information` section, but it doesn't add much. Pages such as 'Adapter Statistics' only care about 1 name (in this case the Receiver).

Simply adding a mandatory attribute 'name' to either the Listener or Receiver wouldn't work, as this would break configuration backwards compatibility. The Receiver name is most important, however certain listeners such as the JavaListener and FrankListener use their (already mandatory) name attribute for others to have messages routed to them. Removing name there wouldn't work, and having both is very confusing.

Who has ideas on how to deal with this? 😿 </div>

nielsm5:
## Changes

<details>
<summary>
</summary>
  
### Title

### Issues

### Backports

### Documentation

### Tests

### Breaking changes

</details>

### DIFF
--- a/core/src/main/java/org/frankframework/receivers/PullingListenerContainer.java
+++ b/core/src/main/java/org/frankframework/receivers/PullingListenerContainer.java
@@ -379,7 +379,7 @@ public class PullingListenerContainer<M> implements IThreadCountControllable {
 					pollToken.release();
 				}
 				threadsRunning.decrementAndGet();
-				if (listener != null) {
+				if (listener != null && threadContext != null) {
 					try {
 						listener.closeThread(threadContext);
 					} catch (ListenerException e) {
--- a/core/src/main/java/org/frankframework/core/Adapter.java
+++ b/core/src/main/java/org/frankframework/core/Adapter.java
@@ -891,15 +891,17 @@ public class Adapter extends GenericApplicationContext implements ManagableLifec
 	@SuppressWarnings("java:S3457") // Cast arguments to String before invocation so that we do not have a recursive call to logger when trace-level logging is enabled
 	public void addReceiver(Receiver<?> receiver) {
 		if (StringUtils.isBlank(receiver.getName())) {
+			String newName = createReceiverName(receivers.size() + 1);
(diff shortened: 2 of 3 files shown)


evandongen (approved):

core/src/main/java/org/frankframework/receivers/PullingListenerContainer.java
```diff
@@ -379,7 +379,7 @@ public void run() {
 					pollToken.release();
 				}
 				threadsRunning.decrementAndGet();
-				if (listener != null) {
+				if (listener != null && threadContext != null) {
```
tnleeuw, line 382: Kan `listener` hier wel null zijn eigenlijk?

tnleeuw (approved):

core/src/main/java/org/frankframework/receivers/PullingListenerContainer.java
nielsm5: Volgens IntelliJ wel, maar er moet wel echt iets fout zijn gegaan mocht je in dat scenario komen...

PR #11370, opened 2026-08-07
Check whether an empty command is passed

evandongen:
## Changes
Another proposed change to only allow absolute paths would break too much


docker/tomcat-debug.yml
```diff
@@ -7,6 +7,8 @@ services:
     volumes:
       - "../test/src/test/testtool:/opt/frank/testtool-ext"
       - "../core/target/frankframework-core-$VERSION.jar:/usr/local/tomcat/webapps/iaf-test/WEB-INF/lib/frankframework-core-$VERSION.jar:ro"
+      - "../../haalcentraal-connector/src/main/configurations:/opt/frank/configurations:ro"
+      - "../../haalcentraal-connector/src/main/resources:/opt/frank/resources:ro"
```
nielsm5:
```suggestion
```
?

evandongen: Oeps. Dank je

nielsm5 (approved):

PR #11374, opened 2026-08-09
Fix [object Object] in Security Items "Used Certificates" info column

tommy2d:
## Problem

On the **Security Items** page, the *Used Certificates* table renders **`[object Object]`** in the **Info** column.

## Cause

The template binds the whole certificate object instead of its info:

```html
<!-- security-items.component.html -->
<td>{{ cert.certificate.name }}</td>
<td>{{ cert.certificate }}</td>   <!-- Info: whole object -> [object Object] -->
```

The backend already provides the data — `AdapterStatus#addCertificateInfo` returns `{ name, authAlias, url, info }`, where `info` is the certificate-chain details (a list whose entries are either a single line or a per-alias list of lines, or the string `*** ERROR ***`). The frontend `Certificate` type only declared `{ name }`, so nothing flagged the wrong binding.

## Fix

- Extend the `Certificate` type with the fields the backend actually sends (`authAlias`, `url`, `info`).
- Add a small `certificateInfoLines()` helper that flattens the nested `info` into a flat list of display lines.
- Render those lines in the Info cell instead of the raw object.

Before: `[object Object]` · After: the certificate-chain lines (Subject DN / Valid until / Issuer / …), one per line.

Prettier-clean. (Local `ng lint` couldn't run here due to an unrelated `unicorn` plugin version mismatch in my environment; CI lint should cover it.)


tommy2d:
For completeness, a before and after screenshot:
Scherm­afbeelding 2026-08-09 om 13 33 47
Scherm­afbeelding 2026-08-09 om 13 31 29

console/frontend/src/main/frontend/src/app/app.service.ts
```diff
@@ -69,6 +69,8 @@ export type PipeMessage = {
 
 export type Certificate = {
   name: string;
+  // Certificate-chain details per keystore alias (a list of lines), or a single error string.
+  info?: Array<string | string[]>;
 };
```
Matthbo, line 74: Certificates in the framework are hashmaps containing "name", "authAlias" & "info". Would be nice if they were defined here for completeness.

We never use the `Array<T>` type, instead use `T[]`.
In this case that would make info `(string | string[])[]`

Matthbo: The comments also don't really add relevant context so I rather not have them unless something needs clearing up

Matthbo (changes requested):

console/frontend/src/main/frontend/src/app/app.service.ts
tommy2d: Done — added `authAlias` and switched to `T[]`. One nuance: `AdapterStatus#addCertificateInfo` sets `info` to the string `"*** ERROR ***"` when the keystore URL can't be resolved, and to the chain array otherwise, so I typed it `string | (string | string[])[]` and made `certificateInfoLines()` handle the string case. Happy to narrow it to just `(string | string[])[]` if you'd rather not model the error path here.

tommy2d: Removed the comments (both on the type and the helper).

Matthbo: Its not a pretty type by any means and the backend should probably reflect a better format so the error result should fit the type structure more akin to the normal result.
But that's something to consider at a later point. This looks good

Matthbo (approved):

PR #11386, opened 2026-08-10
Some experiments using AssertJ instead of Hamcrest

tnleeuw:
What do you guys think -- is this way of writing test assertions an improvement, or not?

There are also ways to add custom `assertThat` methods based on custom domain model entities, with special test-methods based on custom entities. I did not yet implement any of that for this PR, scope of its utility would be limited so far.

Docs for AssertJ are here:

https://assertj.github.io/doc/#assertj-core-extensions

### DIFF
--- a/core/src/test/java/org/frankframework/http/rest/ApiListenerServletTest.java
+++ b/core/src/test/java/org/frankframework/http/rest/ApiListenerServletTest.java
@@ -1057,7 +1057,6 @@ public class ApiListenerServletTest {
 				.has(header("content-type", "application/json;charset=UTF-8"))
 				.has(header("Cache-Control", "no-store, no-cache, must-revalidate, max-age=0, post-check=0, pre-check=0"))
 				.has(header("pragma"));
-		;
 	}
 
 	@Test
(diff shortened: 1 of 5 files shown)


core/src/test/java/org/frankframework/http/rest/ApiListenerServletTest.java
```diff
@@ -991,12 +994,13 @@ public void apiListenerWithExplicitlyEnabledEtag() throws Exception {
...
-		assertTrue(result.containsHeader("etag"));
-		assertEquals("must-revalidate, max-age=0, post-check=0, pre-check=0", result.getHeader("Cache-Control"));
-		assertFalse(result.containsHeader("pragma"));
-		assertEquals("Fri, 13 Jan 2023 13:02:00 GMT", result.getHeader("Last-Modified"));
+		assertThat(result)
+				.has(header("Allow", "OPTIONS, GET"))
+				.has(header("etag"))
+				.has(header("Cache-Control", "must-revalidate, max-age=0, post-check=0, pre-check=0"))
+				.doesNotHave(header("pragma"))
+				.has(header("Last-Modified", "Fri, 13 Jan 2023 13:02:00 GMT"));
```
nielsm5, line 1003: Oh das leuk!

nielsm5 (approved):

PR #11399, opened 2026-08-12
Move SOAP encryption methods to its own util class

### LINKED ISSUE #11379: Further implement SOAP Signing features
Right now we have an older (legacy) way of singing soap messages using a `DigestValue` based on a password.
We'd like to use certificates instead. Signature algorithm: SHA265. WSA/2005/08

- [x] Include certificate in message
- [x] Remove the WS-Security header
- [x] Place digital signature (which requires you to specify a certificate)
- [x] Encrypt message
- [x] Sign message

nielsm5:
Make configurable to omit cert from message
Make configurable to remove the soap security header

## Changes

<details>
<summary>
</summary>
  
### Title

### Issues

### Backports

### Documentation

### Tests

### Breaking changes

</details>


core/src/main/java/org/frankframework/soap/SoapUtils.java
```diff
@@ -0,0 +1,275 @@
...
+		WSSConfig.init();
+	}
+
+	@NonNull
+	public static Document toSoapDocument(Message soapMessage) throws SOAPException, IOException {
+		if (soapMessage.isRequestOfType(Document.class) && soapMessage.asObject() instanceof Document d) {
+			// If the source is already a Document (not NODE) it we can directly use it.
+			// Yes we check things twice, but it keeps my compiler happy...
+			return d;
+		}
```
tnleeuw, line 67: Why not do only the instance-of check?

core/src/main/java/org/frankframework/soap/SoapUtils.java
```diff
@@ -0,0 +1,275 @@
...
+
+		// Support for different soap versions.
+		String soapVersion = soapMessage.getContext().get(SoapContext.SOAP_VERSION_KEY);
```
tnleeuw, line 70: What if nobody has set these properties yet in the context?

tnleeuw (approved):

core/src/main/java/org/frankframework/soap/SoapUtils.java
nielsm5: I had the requestOfType first, but my compiler complained about an unsafe cast...

core/src/main/java/org/frankframework/soap/SoapUtils.java
nielsm5: Default was/is soap1.1 :)
The fact that the code previously only supported soap1.1 annoyed me. Once a message has been validated these are set.

I hope to use the `SoapContet` in the SoapWrapper which (also) sets these.

core/src/main/java/org/frankframework/soap/SoapUtils.java
tnleeuw: I think you can remove the requestOfType and the compiler won't complain?

core/src/main/java/org/frankframework/soap/SoapUtils.java
tnleeuw: Yes but, what if it's SOAP 1.2 but nobody yet looked at the message to set those SOAP version properties?
Is that possible?

Or is there other code that must have already executed, before we can reach this point, that has set those properties?

(Both when message comes in via SOAP WebService and when it comes in via ESB bus / queue).

core/src/main/java/org/frankframework/soap/SoapUtils.java
nielsm5: Yup, I believe so ;)

core/src/main/java/org/frankframework/soap/SoapUtils.java
nielsm5: If it's soap 1.2 and nobody has set those properties you get an exception.

PR #11417, opened 2026-08-13
We need the annotations dependency to avoid using transitive versions

evandongen:


tnleeuw:
This does not seem to solve the WildFly dependency conflict problem; upgrading to WildFly 39 or later does seem to fix it.

I'm therefore closing this PR (for now at least).
image

PR #11426, opened 2026-08-14
Allow OAUTH2 authenticator to also accept bearer tokens (browser OIDC + client-credentials on one endpoint)

tommy2d:
## Why

We expose an API on top of framework features (adapter restart, Ladybug) to external systems, while the same endpoints must stay reachable by humans in a browser. Today an endpoint binds exactly one authenticator, so you must pick either `OAUTH2` (interactive login, browser) **or** `BEARER_ONLY` (JWT, machine) — not both.

## What

Add **`allowBearerAuthentication`** to `OAuth2Authenticator`. When enabled, a single endpoint accepts **both**:
- a **human in a browser** → interactive OIDC login (`oauth2Login`, unchanged), and
- an **external system** presenting `Authorization: Bearer <jwt>` → validated statelessly as an OAuth2 resource server. The machine obtains its token via the `client_credentials` grant at the same IdP.

On one chain: a request with a bearer token hits the resource-server filter; a browser request without a token falls through to the login redirect.

## How

- Extract the JWT validation that already lived in `BearerOnlyAuthenticator` — decoder (`issuerUri`/`jwkSetUri`), `audience` check, `userInfoUri` enrichment, and the claim→authority mapping — into a shared **`AbstractJwtAuthenticator`** that both `BearerOnlyAuthenticator` and `OAuth2Authenticator` extend. `BearerOnlyAuthenticator` is unchanged in behaviour (its tests pass untouched).
- `OAuth2Authenticator.configure()` calls `configureBearerTokenResourceServer(http)` after `oauth2Login` when the flag is set, and installs a `DelegatingAuthenticationEntryPoint`: API/bearer requests (bearer header, or `Accept` not `text/html`) get **401 `WWW-Authenticate: Bearer`**; browser requests get the **login redirect**.
- Requires `issuerUri` or `jwkSetUri`. `allowBearerAuthentication` defaults to **false**, so existing `OAUTH2` deployments are unaffected.

Config example:
```
application.security.console.authentication.type=OAUTH2
application.security.console.authentication.provider=...      # browser login
application.security.console.authentication.clientId=...
application.security.console.authentication.clientSecret=...
application.security.console.authentication.allowBearerAuthentication=true
application.security.console.authentication.issuerUri=https://keycloak/.../realms/<realm>
```

## Tests

`OAuth2AuthenticatorTest`: bearer enabled + jwkSetUri configures cleanly; bearer enabled without issuerUri/jwkSetUri throws; disabled-by-default keeps current behaviour. `BearerOnlyAuthenticatorTest` unchanged (28 tests, green). 0 checkstyle violations.

Open to feedback on the shape — happy to instead make this a separate authenticator type, or adjust the entry-point heuristic (currently bearer-header/Accept based).

### DIFF
--- a/security/src/main/java/org/frankframework/lifecycle/servlets/AbstractJwtAuthenticator.java
+++ b/security/src/main/java/org/frankframework/lifecycle/servlets/AbstractJwtAuthenticator.java
@@ -1,0 +1,199 @@
+/*
+   Copyright 2026 WeAreFrank!
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+package org.frankframework.lifecycle.servlets;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import org.apache.commons.lang3.StringUtils;
+import org.springframework.core.ParameterizedTypeReference;
+import org.springframework.http.HttpHeaders;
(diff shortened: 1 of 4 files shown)


evandongen:
We liked the idea of moving the shared attributes to a base class since that means less duplication! We've implemented that in https://github.com/frankframework/frankframework/pull/11509 and will take a look at this PR after that has gone through.

PR #11430, opened 2026-08-14
Add HttpHeaderUtils to validate Content-Disposition header filenames

### LINKED ISSUE #10729: Tomcat silently swallows Content-Disposition header when default filename attribute contains non ISO-8859-1 characters
### Current Behavior

According to the relevant specifications (see sources below), the default filename parameter of the Content-Disposition header must conform to ISO‑8859‑1, which does not support emojis.

During our migration to Tomcat on Kubernetes, we encountered an inconsistency in behavior when a filename contains an emoji. WebSphere throws an exception in this scenario (which is also undesirable), whereas Tomcat silently drops the entire Content-Disposition header.

The specifications also define an alternative filename* parameter, which allows the filename to be encoded in UTF‑8. By applying the configuration changes described below, the response behaves as expected: the Content-Disposition header is preserved and modern browsers correctly select the UTF‑8 filename for the download.

Image

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition
https://datatracker.ietf.org/doc/html/rfc5987#section-3.2

### Expected Behavior

By setting the Content-Disposition header, the server explicitly instructs the browser to download the response payload and to use the provided filename for the downloaded file.

The framework should assist Frank developers in complying with the Content-Disposition specifications by offering a generic and consistent way to indicate that a response is intended for download. This should be supported by the API listener and, where applicable, by other HTTP‑based listeners as well.

### Environment Information

7.9.7-ee on k8s, tomcat 9, java 11

### Steps To Reproduce

_No response_

### Configuration

```xml

```

### Input

_No response_

_No response_

_No response_

### Relevant Log Output

```console

```

### Anything else?

_No response_

evandongen:
## Changes
Introduced a validator method in a new `HttpHeaderUtils` class which is able to validate the `filename=file.txt` part in different formats from the Content disposition header. Set up to throw a RuntimeException if anything is passed in which contains characters not in ISO-8859-1 according to the [RFC](https://www.rfc-editor.org/info/rfc6266/#section-4.3)

<details>
<summary>
</summary>
  
### Title

### Issues

### Backports

### Documentation

- [x] FF! Reference updated (user-facing behaviour/config)
- [n/a ] FF! Docs updated (if applicable)

### Tests

### Breaking changes

</details>

### DIFF
--- a/core/src/main/java/org/frankframework/http/HttpHeaderUtils.java
+++ b/core/src/main/java/org/frankframework/http/HttpHeaderUtils.java
@@ -1,0 +1,74 @@
+/*
+  Copyright 2026 WeAreFrank!
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+package org.frankframework.http;
+
+import java.net.URLDecoder;
+import java.nio.charset.CharsetEncoder;
+import java.nio.charset.StandardCharsets;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.Nullable;
+
+public class HttpHeaderUtils {
+	private static final CharsetEncoder ISO_8859_1_ENCODER = StandardCharsets.ISO_8859_1.newEncoder();
+
+	/**
+	 * Matches the plain {@code filename=} parameter in a Content-Disposition header (case-insensitive), but NOT the extended {@code filename*=}
+	 * variant (RFC 5987), whose value is always percent-encoded and therefore safe.
+	 * <br />
+	 * Captures the filename value in group 1 (quoted) or group 2 (unquoted).
+	 */
+	private static final Pattern FILENAME_PARAM_PATTERN = Pattern.compile(
+			"(?:^|;)\\s*filename\\s*=\\s*(?:\"([^\"]*)\"|([^;\\s]*))",
+			Pattern.CASE_INSENSITIVE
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/frankframework/http/HttpHeaderUtils.java
```diff
@@ -0,0 +1,74 @@
...
+		Matcher matcher = FILENAME_PARAM_PATTERN.matcher(contentDispositionHeader);
+		while (matcher.find()) {
+			String filename = matcher.group(1) != null ? matcher.group(1) : matcher.group(2);
+
+			if (filename == null) {
+				continue;
+			}
+
+			if (!ISO_8859_1_ENCODER.canEncode(getStringToCheck(filename))) {
+				throw new IllegalArgumentException("Content-Disposition 'filename' parameter contains characters that cannot be encoded in ISO-8859-1: " + filename);
```
evandongen, line 56: @nielsm5, I wasn't sure this was the way you had this in mind, please let me know. The current modification covers all three entry points in the code using content disposition, this should make a change easy.

evandongen:
The issue mentions this was encountered in 7.9.7, but that isn't supported anymore. In what version(s) does this need to be backported?

core/src/main/java/org/frankframework/http/HttpHeaderUtils.java
tnleeuw: Would it be preferable to throw an exception, or to sanitize the name and drop characters that cannot be ISO-8859-1 encoded?

nielsm5: On other application servers the default on one is to throw an exception and others drop the header...

nielsm5:
> The issue mentions this was encountered in 7.9.7, but that isn't supported anymore. In what version(s) does this need to be backported?

10.0/10.2

core/src/main/java/org/frankframework/http/HttpHeaderUtils.java
tnleeuw: So sanitizing the header-value is not preferred, but rather throwing the exception (as in the PR) is.

👍

nielsm5: I think for now, yes.

nielsm5 (approved):

PR #11472, opened 2026-08-21
Multiple configurations support with autoload functionality

### LINKED ISSUE #11471: Support multiple configurations in Jar file / auto discovery
### Current situation

Currently we support the automatic discovery of configurations in `/opt/frank/configurations` by placing a JAR file there. This jar file has to have the following parent:
```
<parent>
	<groupId>org.frankframework</groupId>
	<artifactId>configuration-parent</artifactId>
	<version>10.2.0</version>
</parent>
```

and have a name:
```
<name>configurationName</name>
```

The contents of the jar file would like something like this:
``` 
* META-INF
  * MANIFEST.MF
* configurationName
  * Configuration.xml
```

As you can see, this enforces a one on one relation between the name and the number of supported configurations.

### Desired situation

We'd like to add a new supported property to the `configuration-parent`, which translates to a property in MANIFEST.MF -- to define the configurations contained in the JAR file. That'd look like this in the pom file: `<framework.configuration.names>configurationName1, configurationName2, configurationName3</framework.configuration.names>`

Which in turn would translate to: `FrankFramework-configurations: configurationName1, configurationName2, configurationName3` in the MANIFEST. This should support a jar file with the following structure:
```
* META-INF
  * MANIFEST.MF
* configurationName1
  * Configuration.xml
* configurationName2
  * Configuration.xml
* configurationName3
  * Configuration.xml
```

evandongen:
## Changes

<details>
<summary>
</summary>
  
### Title

### Issues

### Backports

### Documentation

### Tests

### Breaking changes

</details>


build-templates/pom.xml
```diff
@@ -52,9 +52,9 @@
 		<maven.build.timestamp.format>yyyy-MM-dd HH:mm</maven.build.timestamp.format>
 
 		<revision>10.3.0-SNAPSHOT</revision>
-		<framework.version>${project.version}</framework.version>
+		<framework.version>${revision}</framework.version>
 		<framework.version.range>[${framework.version},)</framework.version.range>
-
+		<framework.configuration.names />
```
nielsm5:
```suggestion
		<configuration.names>${project.artifactId}</configuration.names>
```
Pehaps a bit shorter and is a default to the artifact name handy?

core/src/main/java/org/frankframework/configuration/classloaders/JarFileClassLoader.java
```diff
@@ -68,7 +69,7 @@ private Path locateJarFile() throws ClassLoaderException {
 			Path configDir = ConfigurationUtils.getConfigurationDirectory();
 			try (Stream<Path> input = Files.list(configDir)) {
 				return input.filter(JarFileClassLoader::isJarFile)
-						.filter(e -> getConfigurationName().equals(findConfigurationName(e)))
+						.filter(e -> findConfigurationNames(e).contains(getConfigurationName()))
```
nielsm5, line 72: I think the other changes are fine. Even though multiple resources are packaged in 1 file, they will get their own classloader. Each classloader should only find resources in their own scoped path.

nielsm5 (changes requested):

core/src/main/java/org/frankframework/configuration/classloaders/JarFileClassLoader.java
evandongen: Thanks, that part was a bit hard to get through

MatthijsSmets:
I am missing some kind of warning log when more configurations are found in the jar, that were not configured under `configuration.names`.

core/src/main/java/org/frankframework/configuration/classloaders/JarFileClassLoader.java
evandongen: Niels will take a look if there are any complications in `org.frankframework.configuration.classloaders.AbstractJarBytesClassLoader#readResources`

nielsm5 (approved):

core/src/main/java/org/frankframework/configuration/classloaders/AbstractJarBytesClassLoader.java
```diff
@@ -41,26 +41,41 @@ protected final Map<String, byte[]> readResources(InputStream stream) throws Cla
...
 			Map<String, byte[]> resources = new HashMap<>();
 			JarEntry jarEntry;
+			int skippedFiles = 0;
+
 			while ((jarEntry = jarInputStream.getNextJarEntry()) != null) {
 				String fileName = jarEntry.getName();
 				if(getBasePath() != null) {
-					boolean isFolder = fileName.endsWith("/"); // if the name ends with a slash, assume it's a folder
+					boolean isFolder = fileName.endsWith("/");
+					// If the name ends with a slash, assume it's a folder.
```
tnleeuw: Either join the lines again, or put this comment above the code that does the check.

core/src/test/java/org/frankframework/configuration/classloaders/JarFileClassLoaderTest.java
```diff
@@ -93,6 +93,15 @@ public void checkFilePathWithBasePath() {
...
 
+	@Test
+	public void testJarWithMultipleConfigurations() throws  Exception {
+		JarFileClassLoader classLoader = createClassLoader(null, "/ClassLoader/MultipleConfigsInJar/Weer-1.0.0-SNAPSHOT-configurations.jar");
+
+		appConstants.setProperty("configurations.myConfig.classLoaderType", classLoader.getClass().getSimpleName());
+
+		classLoader.configure(ibisContext, "Weer");
+	}
+
```
tnleeuw: What is being tested / asserted in this test?

Which are the multiple configurations in this JAR file? Test-name implies that there should be more than 1.

core/src/test/resources/ClassLoader/MultipleConfigsInJar/Weer-1.0.0-SNAPSHOT-configurations.jar
tnleeuw, line 1: I don't think the same binary needs to be added twice, I think we can use it in both tests from this location?

And if they're not identical, then the names should be changed to distinguish them better.

evandongen: It's not the same binary. The MANIFEST differs

core/src/test/java/org/frankframework/configuration/classloaders/JarFileClassLoaderTest.java
evandongen: You're right, this was the initial entry point for me to test the code. The other tests are more relevant.

tnleeuw: It could still be a valid test if you surround some statements with `assertDoesNotThrow`.

core/src/test/resources/ClassLoader/MultipleConfigsInJar/Weer-1.0.0-SNAPSHOT-configurations.jar
tnleeuw: In that case I prefer that filenames are different to reduce confusion.

PR #11493, opened 2026-08-24
Don't use URL type as Map key

tnleeuw:
## Changes

According to IntelliJ we should not use URL as Map keys:
Reports hashCode() and equals() calls on java.net.URL objects and calls that add URL objects to maps and sets.
URL's equals() and hashCode() methods can perform a DNS lookup to resolve the host name. This may cause significant delays, depending on the availability and speed of the network and the DNS server. Using java.net.URI instead of java.net.URL will avoid the DNS lookup.

Using String as the Map key results in slightly less ugly code as using the URI however, so I chose that instead.

<details>
<summary>
</summary>
  
### Title

### Issues

### Backports

### Documentation

### Tests

### Breaking changes

</details>


nielsm5:
Transient field?

nielsm5 (approved):

PR #11499, opened 2026-08-25
Add data-cy attribute for accessing cluster member dropdown in Cypress tests

mhdirkse:
Add a data-cy attribute at the drop-down menu of the cluster member. This is done to support testing Ladybug.


Matthbo:
The FF Console is not a part of ladybug and since this attribute is not relevant for Console tests at this point in time I'm not going to approve this

PR #11543, opened 2026-09-01
Clarify getInputFromSessionKey missing-key exception #11377

### LINKED ISSUE #11377: then pipe that is not a org.frankframework.pipes.FixedForwardPipe" phrasing unclear
### Context Information

**Current Page:** `/Components/Pipes/CounterSwitchPipe`
**Frank!Doc Version:** `10.3.0-20260807.042335`
**Selected Text:** `None`

### What would you like to change?

```
<html>
<body>
<!--StartFragment-->
getInputFromSessionKey 
If set, input is taken from this session key, instead of regular input. If the session-key does not exist or is empty, then the input will instead be taken from the defaultValue. If there is no defaultValue and the session-key is missing, then pipe that is not a org.frankframework.pipes.FixedForwardPipe will throw an exception that the session-key is missing.
</body>
</html>
```

" then pipe that is not a org.frankframework.pipes.FixedForwardPipe" Makes it sound like this only applies to a pipe named "FixedForwardPipe" 
Did some digging through the code and found it's a base pipe other pipe inherit properties from, would be better to change the phrasing to "Child of org.frankframework.pipes.FixedForwardPipe" 
Also couldn't find any indicator as to what pipes are and what pipes aren't in the documentation pages, maybe i missed something? Else it would be nice to have some kind of indicator

AzazelSensei:
The Frank!Doc for `getInputFromSessionKey` said a missing session key only threw when the pipe was not a FixedForwardPipe. That reads like a class-name special case.

`InputOutputPipeProcessor.getInputFrom` throws for any pipe when the key is missing and there is no `defaultValue`. I rephrased the Javadoc to match that.

Fixes #11377

### DIFF
--- a/core/src/main/java/org/frankframework/core/IPipe.java
+++ b/core/src/main/java/org/frankframework/core/IPipe.java
@@ -113,8 +113,7 @@ public interface IPipe extends IConfigurable, IForwardTarget, FrankElement, Name
 	/**
 	 * If set, input is taken from this session key, instead of regular input. If the session-key does not exist or is empty, then the input will
 	 * instead be taken from the {@code defaultValue}.
-	 * If there is no {@code defaultValue} and the session-key is missing, then pipe that is not a {@link org.frankframework.pipes.FixedForwardPipe} will
-	 * throw an exception that the session-key is missing.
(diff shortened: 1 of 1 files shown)


AzazelSensei:
Closing this after 15 days without a comment.

PR #11596, opened 2026-09-04
Remove CMIS from full bundle

### LINKED ISSUE #11384: Disable CMIS module by default
Remove module from full-bundle, to reduce the war file and container image size.

We should make this plugins for the framework.
This applies to the following modules:
* Tibco, sap, IDIN
* Perhaps Aspose
* CMIS

nielsm5:
Unsure how full this full bundle becomes, and if we should create an inbetween bundle (I'm unsure what to call it ...).

At the moment we have Aspose and CMIS which are not part of any bundle.
The 'now not so full bundle' contains:
- akamai
- aws
- batch
- nn-specials
- messaging
- kubernetes

If we were to create a full-full bundle, we should add SAP/Tibco/IDiN to it as well.

### DIFF
--- a/bundle-full/pom.xml
+++ b/bundle-full/pom.xml
@@ -32,10 +32,6 @@
 		</dependency>
 		<dependency>
 			<groupId>org.frankframework</groupId>
-			<artifactId>frankframework-cmis</artifactId>
-		</dependency>
-		<dependency>
-			<groupId>org.frankframework</groupId>
 			<artifactId>frankframework-aws</artifactId>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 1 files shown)


tnleeuw:
> If we were to create a full-full bundle, we should add SAP/Tibco/IDiN to it as well.

Isn't that full-full-bundle really the NN-bundle then? (With a bit of ING perhaps?)

bundle-full/pom.xml
```diff
@@ -57,6 +53,24 @@
 	</dependencies>
 
 	<profiles>
+		<profile>
+			<id>all-dependencies</id>
```
tnleeuw, line 57: "all" still isn't really "all" ... missing IDIN, Tibco, etc. Right?

evandongen: All implicates something else, agreed

nielsm5: Yes thanks, so should we create a third bundle with all these modules enabled bydefault, and no more profiles?

tnleeuw: That sounds better to me.

tnleeuw: The 'enterprise' bundle, published only on private Nexus for enterprise customers (aka NN)? 😉

nielsm5:
- Add `Aspose` to the docker image.
- Remove `CMIS` from our WAR archive.

tnleeuw (approved):

evandongen (approved):

PR #11604, opened 2026-09-07
Explain how contextKey attributes on Param element works

### LINKED ISSUE #11352: Unclear that 'contextKey' on Param requires the specific message to be the input of the Param
### Current Behavior

Unless you know about the internal workings of FF, it is not immediately clear how the messageContext is attached to messages or how output from pipes is actually stored in sessionkeys.

It is very confusing and unclear from documentation that the 'contextKey' attribute on a Param is dependent on the input of the Param. Even more confusing that Ladybug shows the messageContext to be available on the input of a pipe, but because of a `inputFromFixedValue="{}"` on the pipe, the Param will not have the messageContext of the original input message available.

### Expected Behavior

- Documentation with examples of usage.
- Made clear how the message context relates to a message.
- Made clear that not just a raw value is stored in sessionkeys, but a message object including the message context.

### Environment Information

**FF! 10.2.1-20260728.085909**: ZaakBrug 1.26.16
Running on **39c91c9c5a32** using **Apache Tomcat/11.0.24**
Java Version: **OpenJDK Runtime Environment (25.0.3+9-LTS)**
Heap size: **736.0 MiB**, total JVM memory: **1.1 GiB**
Free memory: **376.0 MiB**, max memory: **11.5 GiB**
Free disk space: **930.0 GiB**, total disk space: **1006.9 GiB**
Up since: **2026-08-05 15:36:59**, timezone: **Europe/Amsterdam**

### Steps To Reproduce

_No response_

### Configuration

```xml

```

### Input

_No response_

_No response_

_No response_

### Relevant Log Output

```console

```

### Anything else?

_No response_

evandongen:
Updated javadoc

### DIFF
--- a/core/src/main/java/org/frankframework/parameters/AbstractParameter.java
+++ b/core/src/main/java/org/frankframework/parameters/AbstractParameter.java
@@ -897,6 +897,10 @@ public abstract class AbstractParameter<T> implements IConfigurable, IWithParame
 	/**
 	 * Key of {@link org.frankframework.stream.MessageContext} variable to use as source, instead of the {@link Message} found from input message or sessionKey itself. Use a {@literal *}
 	 * to get an XML or JSON document containing all values from the {@link org.frankframework.stream.MessageContext}.
+	 * This assumes that the input of this parameter is a {@link Message}, in which the contextKey will be looked up.
+	 * <p>
+	 * For instance: {@code <Param name="context.value" sessionKey="originalMessage" contextKey="Header.content-type" />} - this will look up the value of the
+	 * context variable "Header.content-type" from the message found in the session variable "originalMessage".
 	 */
 	public void setContextKey(@Nullable String string) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/frankframework/parameters/AbstractParameter.java
```diff
@@ -897,6 +897,10 @@ public void setSessionKey(@Nullable String string) {
 	/**
 	 * Key of {@link org.frankframework.stream.MessageContext} variable to use as source, instead of the {@link Message} found from input message or sessionKey itself. Use a {@literal *}
```
nielsm5:
```suggestion
	 * Key of a {@link org.frankframework.stream.MessageContext} variable to use, instead of the Parameter value found from the input message or sessionKey. Use a {@literal *}
```

nielsm5 (approved):

PR #11648, opened 2026-09-12
Keep the OAuth2 login session and return users to the page they requested

tommy2d:
## Changes

Users of the OAuth2 (authorization-code) authenticator land on the page they originally requested after logging in, and stay logged in instead of being sent back to the identity provider on every request.

`AbstractServletAuthenticator` applies `SessionCreationPolicy.STATELESS` to every chain, which is correct for the bearer-token and basic-auth authenticators. For the interactive login of `OAuth2Authenticator` it has two side effects, because Spring Security derives a `NullRequestCache` and a request-scoped `SecurityContextRepository` from that policy:

- the originally requested URL is never saved, so every login redirects to the default success URL `/` instead of the deep link the user opened;
- the authentication never outlives the request, so every subsequent request repeats the authorization round-trip to the identity provider.

The fix is scoped to the `OAuth2Authenticator` chain only:

- `SessionCreationPolicy.IF_REQUIRED` for that chain;
- an `HttpSessionRequestCache`, so the saved request is replayed after the identity provider redirects back;
- an `HttpSessionSecurityContextRepository`, registered both on the `securityContext` configurer and as shared object — the `oauth2Login` filter takes its repository from the shared objects — so the authentication is persisted in the session.

All other authenticators keep `STATELESS`; bearer-authenticated requests still create no session.

`OAuth2AuthenticatorTest.testInteractiveLoginChainIsStateful` asserts that the configured chain carries the session-backed request cache and security-context repository.

<details>
<summary>
</summary>

### Title

### Issues
- [ ] Relevant issue(s) linked — no existing issue

### Backports

### Documentation
- [x] FF! Reference updated (user-facing behaviour/config) — not applicable, no configuration change
- [x] Javadoc updated/generated (developer-facing APIs) — not applicable, no API change
- [x] FF! Docs updated (if applicable) — not applicable

### Tests
- [x] E2E/Integration tests added/updated (if applicable) — not applicable

### Breaking changes
- [x] Breaking change recorded in markdown file — none
- [x] Migration notes included (if needed) — none
</details>


tommy2d:
Context on why I need this: we serve application pages under `/webcontent/*` and hand out those URIs directly to users (links in mail, bookmarks, other systems). Without this, every such link drops the user on `/` after the IdP round-trip, so the deep link is effectively lost — and without the persisted security context each follow-up request bounces back to the IdP again.

nielsm5:
This creates session leaks over multiple servlets. If a user is authenticated on 1 servlet, it will automatically be authenticated on all servlets.

tommy2d:
@nielsm5 Fair point, and in my defence, the only thing warning about this was a comment saying exactly *"STATELESS prevents session from leaking over multiple servlets."* Right above the line I overrode. 😂

Fixed: the session attributes (`SPRING_SECURITY_CONTEXT` and the saved request) are now scoped per security chain, keyed on that chain's URL mappings. A login on `/webcontent/*` is no longer picked up by the console or ladybug chains, while the deep-link replay and the persisted login still work within the chain itself. API calls no longer create a session either. Added a test that saves an authentication on one chain and asserts another chain can't see it.

PR #11655, opened 2026-09-13
Store the actual text of a small error result in the error-store comment

tommy2d:
On an error exit the Receiver records the pipeline result in the error-store comment as `exitState [..], result [..]`. Only a result **larger** than the comment cap was rendered as its actual text; every smaller result fell to the `else` branch and was appended via `Message.toString()`, so the comment read `result [Message[<id>:String] size [40]]` — the descriptor, not the content.

That hides exactly the short, human-readable error messages a configuration produces on an error exit (the reason a message was rejected) — the thing an operator most wants when triaging the error store.

This stores a small result of a **known** size as its actual text. Unknown-size (streaming, `size() < 0`) and empty results still keep the descriptor, so an unbounded stream is never materialised into the comment; a large result is still truncated to the cap as before.

```java
long resultSize = Message.isEmpty(result) ? 0 : result.size();
if (resultSize > MAXCOMMENTLEN - statusMessage.length()) {   // large: truncate (unchanged)
    ...
} else if (resultSize > 0) {                                 // small, known: actual text
    statusMessage += result.asString();
} else {                                                      // empty / unknown size: descriptor
    statusMessage += result;
}
```


tommy2d:
Withdrawing — handling this on our side instead, not asking the framework to change.

PR #11658, opened 2026-09-14
Extend CONTRIBUTING.md with a section about AI usage

evandongen:


CONTRIBUTING.md
```diff
@@ -66,6 +66,16 @@ If a comment you answered to is very old and if no additional
 answer has been added to the conversation for a long time, you
 can also set it to resolved to ease the review process.
 
+## AI usage
+We don't forbid using AI tools, but would like to make sure that you are aware of the risks. Since everything in the F!F is very tightly-knit together, 
+it's easy to make a change which breaks something else.
+
+We assume the following guidelines:
+* Do not use AI to create the feature for you. You are responsible for the code you submit.
```
tnleeuw:
```suggestion
* Do not use AI to create the feature for you. You are responsible for the code you submit, and should be able to explain your changes in detail.
```

CONTRIBUTING.md
```diff
@@ -66,6 +66,16 @@ If a comment you answered to is very old and if no additional
...
+* You can use AI to help you with the boilerplate code and derived tasks, like creating unit tests, or documentation. But also in these cases, review the generated changes.
```
tnleeuw: Not sure about letting AI create the unit tests.

The unit tests verify the contract your code should abide by, the unit tests should be thorough and well-thought out.

They shouldn't be "whatever tests green against the current implementation" and I'm afraid that using AI just to generate unit tests will lead to that kind of tests, that just verify that "code works as implemented".

I wonder how others see that?

evandongen: What you interpreted is not what we meant to say. AI can help you with creating unit tests, but you are still responsible for writing correct tests. We meant to say that when you have a working piece of code, AI is very well equipped to be able to explain that code (documentation) or test it (unit test).

I'll try to rephrase

tnleeuw (approved):

nielsm5 (approved):

